PR #237, opened 2023-08-16
fix(console): auto reload via parcel was failing due to a node_module

EdwardIrby:
This PR removes packages not necessary for further development of v0 console at this time.  I was not able to identify exactly what package was causing the issue but I removed the packaged and code related to them.  This includes vite, plop, and storybook files and deps.


console/client/src/components/Navigation.tsx
```diff
@@ -25,10 +25,7 @@ export default function Navigation() {
               <div className='flex h-16 items-center justify-between'>
                 <div className='flex items-center'>
                   <div className='flex-shrink-0 flex items-baseline space-x-4'>
-                    <NavLink to='/'>
-                      <span className='text-indigo-200 text-xl font-medium'>FTL</span>
-                      <span className='px-2 text-rose-400 text-2xl font-medium'>∞</span>
-                    </NavLink>
```
wesbillman: We don't want this anymore?

wesbillman (approved):

PR #238, opened 2023-08-16
fix: Use existing deployment key if found during creation

### LINKED ISSUE #236: After killing a deploy. I'm not able to redeploy the same module
I started with a fresh `db` and issued the following commands.

```bash
ftl go deploy -S $PWD -d ./examples/time
ftl kill D01H7ZQA8VCQ770W4YQQD6VRN69
ftl go deploy -S $PWD -d ./examples/time
```

Screenshot 2023-08-16 at 10 47 19 AM

wesbillman:
Fixes #236 

The previous change was made to create the key earlier which would allow us to log entries about the deploy during the creation process, before it was in the db. This PR retains that and fixes the reported bug.

This could be simplified but we'll lose the ability to log is some spots before the db call.

Still need to figure out how to log a message after the deployment has been replaced. Currently getting:
```bash
controller | info: Created deployment wes D01H809K9G09THVB5Z0HAMST7RZ
controller | info: Replace deployment for: D01H809K9G09THVB5Z0HAMST7RZ
controller | failed to insert log entry: {2023-08-16 16:14:02.341002 -0700 PDT m=+17.468284876 info map[deployment:D01H80A18Z366G9P1F0Y9PNK77C] Created deployment wes D01H809K9G09THVB5Z0HAMST7RZ <nil>} :: error: ERROR: null value in column "deployment_id" of relation "events" violates not-null constraint (SQLSTATE 23502)
```

@alecthomas maybe we should keep old deployment records around after they are replaced so that the IDs will still be in the db, but new deployments will create new IDs?

### DIFF
--- a/backend/controller/controller.go
+++ b/backend/controller/controller.go
@@ -563,8 +563,7 @@ func (s *Service) UploadArtefact(ctx context.Context, req *connect.Request[ftlv1
 }
 
 func (s *Service) CreateDeployment(ctx context.Context, req *connect.Request[ftlv1.CreateDeploymentRequest]) (*connect.Response[ftlv1.CreateDeploymentResponse], error) {
-	deploymentKey := model.NewDeploymentKey()
-	logger := s.getDeploymentLogger(ctx, deploymentKey)
+	logger := log.FromContext(ctx)
 
 	artefacts := make([]dal.DeploymentArtefact, len(req.Msg.Artefacts))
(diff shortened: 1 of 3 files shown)


wesbillman:
Decided to go back to the simplified and more correct version where the key is created right before the database call. This ensures the key is in the db for future calls and logs.

alecthomas (approved):

PR #288, opened 2023-08-24
eirby/tab search params

EdwardIrby:
Creates urls for tabs and tab related activities specifically clicking on the details


console/client/src/providers/tabs-provider.tsx
```diff
@@ -1,33 +1,34 @@
...
   label: string;
-  type: TabType;
-  module?: Module | null;
-  verb?: Verb | null;
+  type: typeof TabType[keyof typeof TabType];
 }
 
+export const TabSearchParams = {
+  active: 'active-tab',
+} as const
```
wesbillman, line 16: I think this should have a structure more like this to make it more generic:

```ts
export const TabSearchParams = {
  type: 'tab-type',
  id: 'tab-id',
} as const
```

Then we could pass generic tab info in the query params instead of `active-tab=echo.echo` for example. If we update this structure we could support opening different tab types as well. Some examples might be:

```
tab-type=verb&tab-id=echo.echo
tab-type=request&tab-id=<RequestId>
etc.
```

console/client/src/components/Navigation.tsx
```diff
@@ -3,39 +3,36 @@ import { classNames } from '../utils/react.utils'
 import { navColor } from '../utils/style.utils'
 import DarkModeSwitch from './DarkModeSwitch'
 
-export default function Navigation() {
+export function Navigation() {
```
wesbillman, line 6: 👍 I've been trying to clean these up as well. Thanks!

console/client/src/layout/IDELayout.tsx
```diff
@@ -1,30 +1,68 @@
...
 export function IDELayout() {
-  const { selectedModule } = useContext(SelectedModuleContext)
-  const { tabs,activeTab, setActiveTab, setTabs } = useContext(TabsContext)
-
+  const { tabs,activeTab, setActiveTab, setTabs } = React.useContext(TabsContext)
+  const navigate = useNavigate()
+  const location = useLocation()
+  const [ searchParams ] = useSearchParams()
+  
+
```
wesbillman, line 23: Can the linter catch stuff like this? I think prettier might have before, but it seems like the `lint:fix` would remove extra whitespace like this.

console/client/src/layout/IDELayout.tsx
```diff
@@ -1,30 +1,68 @@
...
+  // Handle opening the correct tab on load
+  React.useEffect(() => {
+    const id = searchParams.get(TabSearchParams.active)
+    if(!id) {
+      setActiveTab(0)
+      return
+    }
+    const index = tabs.findIndex(tab => tab.id === id)
+    if(index >= 0) {
+      setActiveTab(activeTab  ?? index)
```
wesbillman, line 52: Seems like there are double spaces here `setActiveTab(activeTab  ?? index)` should be `setActiveTab(activeTab ?? index)`

Should the linter find and fix these?

console/client/src/layout/IDELayout.tsx
```diff
@@ -50,35 +88,53 @@ export function IDELayout() {
...
-          <div className={`flex-1 overflow-y-scroll ${panelColor}`}>
-            {activeTab?.type === TabType.Timeline && <Timeline />}
-            {activeTab?.type === TabType.Verb && <VerbTab module={selectedModule} verb={activeTab.verb} />}
-          </div>
+                    <span
+                      className={`px-4 py-2 rounded-t ${id !== 'timeline' ? 'pr-8' : ''} ${activeTab === i ? `${selectedTabStyle}` : `${unselectedTabStyle}`}`}
+                    >
+                      {label}
+                    </span>
+                    {i !== 0 && (<button
```
wesbillman, line 108: Should this check the `tab.id` vs. `i !== 0`?

console/client/src/providers/tabs-provider.tsx
```diff
@@ -1,33 +1,34 @@
...
+
 export const timelineTab = { id: 'timeline', label: 'Timeline', type: TabType.Timeline }
 
 type TabsContextType = {
   tabs: Tab[];
-  activeTab?: Tab | null;
+  activeTab?: number;
   setTabs: React.Dispatch<React.SetStateAction<Tab[]>>;
-  setActiveTab: React.Dispatch<React.SetStateAction<Tab>>;
+  setActiveTab: React.Dispatch<React.SetStateAction<number>>;
```
wesbillman, line 24: I wonder if it would be useful to add a couple more methods to this interface.
```ts
addTab(Tab)
closeTab(tab.id)
```
Then these could handle the logic for selecting the correct tab when `addTab` or `closeTab` is called

console/client/src/features/verbs/VerbCalls.tsx
```diff
@@ -22,6 +22,15 @@ export const VerbCalls: React.FC<Props> = ({ module, verb }) => {
     fetchCalls()
   }, [ client, module, verb ])
 
+  const navigate = useNavigate()
+  const location = useLocation()
+  const [ searchParams ] = useSearchParams()
+  const handleClick: React.MouseEventHandler<HTMLButtonElement> = evt =>{
```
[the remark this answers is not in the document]
EdwardIrby: Nope it's correct when you use React.MouseEventHandler<HTMLButtonElement> on the left side of the `=` it negates the need for a typing the `evt`

console/client/src/features/requests/RequestsModal.tsx
```diff
@@ -0,0 +1,203 @@
...
+import { RequestGraph } from './RequestGraph.tsx'
+
+export function RequestModal() {
+  const navigate = useNavigate()
+  const location = useLocation()
+  const [ searchParams ] = useSearchParams()
+  const client = useClient(ConsoleService)
+  const [ calls, setCalls ] = React.useState<Call[]>([])
+  const key = searchParams.get('requests')?? undefined
+  const moduleName = searchParams.get('details')
```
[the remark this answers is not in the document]
EdwardIrby: Honestly I consider most of this stuff code we'll discard, so not big. 

I do have a generic version I made but realized but headless is also a generic library so my efforts were pointless and also made things more complicated. 

I'll create some mixins object and function for applying css and other properties. We'd also benefit from 2 declarative components for background. Other than that I'd hold off on a pure Modal component. It would lead to prop creep as you continue to try and handle differing requirements. From my experience making design systems it's better to just have nice css & js mixins and very few declarative helpers components like what I described above.

console/client/src/features/verbs/VerbCalls.tsx
```diff
@@ -22,6 +22,15 @@ export const VerbCalls: React.FC<Props> = ({ module, verb }) => {
...
+    const value = evt.currentTarget.value
+    searchParams.set('requests', value)
```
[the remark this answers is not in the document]
EdwardIrby: Sure

console/client/src/features/verbs/VerbTab.tsx
```diff
@@ -1,15 +1,18 @@
...
-  verb?: Verb | null
+  id: string
 }
 
-export const VerbTab: React.FC<Props> = ({ module, verb }) => {
+export const VerbTab: React.FC<Props> = ({ id }) => {
+  const [ moduleId, verbName ] = id.split('.')
+  const modules = React.useContext(modulesContext)
+  const module = modules.modules.find(module => module?.name === moduleId)
+  const verb = module?.verbs.find(v => v.verb?.name === verbName?.toLocaleLowerCase())
```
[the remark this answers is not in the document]
EdwardIrby: Cool that was in your original code so I just kept it

console/client/src/layout/IDELayout.tsx
EdwardIrby: No rule for that but I also don't particularly concern myself with line space except for at the end of the file. I'll look for a rule that does that if it matters to you.

console/client/src/providers/tabs-provider.tsx
EdwardIrby: I'm using headless and headless basses it on index/number. Honestly I prefer that to id's. It means we only have to concern ourself with keeping tab and panel order based on object position.

console/client/src/features/verbs/VerbCalls.tsx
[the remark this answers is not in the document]
wesbillman: I'm not sure I follow. I'm referring specifically to this bit of code `=>{` shouldn't this be `=> {` and shouldn't the linter add that space for us?

console/client/src/layout/IDELayout.tsx
wesbillman: I mostly just want the linter to keep all our code consistent. For example if a developer adds like 8 lines of whitespace that probably isn't what everybody wants so the linter should help us out automatically.

The autoformatting seems fine with stuff like this, which seems off to me:
Screenshot 2023-08-24 at 10 42 35 AM

console/client/src/features/verbs/VerbTab.tsx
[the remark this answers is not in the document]
wesbillman: We should fix it though now that we have some examples that show the issue.

console/client/src/providers/tabs-provider.tsx
wesbillman: I think this comment is unrelated to how headless works. This would just be for maintaining the state of what tabs we have and which are open. If `closeTab(index)` is better for you that's fine, but it would be nice for users of tabs to not have to worry about selecting the correct tab when they add or remove a tab since this `tabs-provider` can handle it for them.

console/client/src/features/verbs/VerbCalls.tsx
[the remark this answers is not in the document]
wesbillman: https://github.com/TBD54566975/ftl/pull/291

console/client/src/layout/IDELayout.tsx
wesbillman: https://github.com/TBD54566975/ftl/pull/291

PR #320, opened 2023-08-25
eirby/console tab fixes

### LINKED ISSUE #245: v0 Console UI mods

### LINKED ISSUE #253: Improve console tab behavior
The console tabs were quickly put together and have a few issues we should try to fix as well as some improvements.

I'm open to using a tab component from headless if it makes things easier and retains the general look and feel we have now. Or we can just update the current code we have today.

- [ ] Currently, the tab is just recreated based on the selected tab ID so a new tab is created each time. Perhaps we should keep an array of the tab components so that the data/context is preserved when we switch between tabs.
- [ ] Support creating initial tabs/context based on URL query parameters. So something like `modules/echo/verbs/echo` would open the `echo` VerbPage next to the default Timeline tab. I think this can use the code you have in your branch for query params, @EdwardIrby 

Dhanji mentioned a preference for `paths` over `query params` for the URL as well so maybe we can take that path for now vs. the query params. An example for opening a verb page might be `http://localhost:8892/modules/echo/verbs/echo`

If a user sends a URL for `http://localhost:8892/modules/echo/verbs/echo`, FTL should open with the `Timeline` tab (because it's default) + 1 new tab which would be `VerbTab` with `echo:echo`

EdwardIrby:
resolves #253 
resolves #245

and fixes requests from #288


EdwardIrby:
https://github.com/TBD54566975/ftl/assets/1058725/157d169b-510a-4140-a3b8-7a893ffa71d4

alecthomas:
Very nice, looking awesome.

EdwardIrby:
I still need to add the error message component to replace commented out throws in https://github.com/TBD54566975/ftl/pull/320/files#diff-34604a63c516e886f158e2569fb4f0bb03afa023ad49bb1d287422c43736a091. I think I'm going to move `isValidTab` out of the component.

console/client/src/features/modules/ModuleDetails.tsx
```diff
@@ -1,15 +1,30 @@
...
-  const { tabs, setTabs, setActiveTab } = useContext(TabsContext)
-  const navigate = useNavigate()
-  const location = useLocation()
-  const [ searchParams ] = useSearchParams()
+  const modules = React.useContext(modulesContext)
+  const { selectedModule, setSelectedModule } = React.useContext(SelectedModuleContext)
+  const { tabs, setTabs, setActiveTab } = React.useContext(TabsContext)
+  const [ searchParams, setSearchParams ] = useSearchParams()
+  const moduleId = searchParams.get('module')
+  //P1 when mounting with a valid module in query params set selected module
```
wesbillman, line 14: What are the `P1` `P2` type comments. Are they param 1,2? Maybe we can add more context here to help?

console/client/src/features/modules/ModuleDetails.tsx
```diff
@@ -1,15 +1,30 @@
...
+      module && setSelectedModule(module)
+    }
+  })
+  //P2 when query param module changes update selected module if valid
+  React.useEffect(() => {
+    if(moduleId) {
+      const module = modules.modules.find(module => module?.name === moduleId)
+      module && setSelectedModule(module)
+    }
+  }, [ moduleId ])
```
wesbillman, line 27: Are these the same? They have different dependencies but seem similar in code.

console/client/src/layout/IDELayout.tsx
```diff
@@ -1,66 +1,74 @@
...
-      setActiveTab(activeTab  ?? index)
-      return
-    }
-    const [ _, label ] = id.split('.')
-    const newTab = {
-      id,
-      label,
-      type: TabType.Verb,
+    if(isValidTab({ id, type })) { // P1 is a valid tab ID and Type
+      // P1.1 id and type are not in tab list
```
wesbillman, line 55: Similar questions about `P1` and `P1.1` here.

console/client/src/providers/tabs-provider.tsx
```diff
@@ -12,23 +12,65 @@ export type Tab = {
...
+  activeTab: undefined,
+  setTabs: () => { },
+  setActiveTab:  () => {},
+})
+
+export const isValidTab = ({ id, type }: {id?: string; type?: string}): boolean => {
+  //P1 no ID or type
+  if(!id || !type) {
+    // throw new Error(`required tab field undefined: ${JSON.stringify({ type, id })}`)
+    return false
```
wesbillman, line 44: For these, should we throw or `return false`? Or are these comments here for future work?

console/client/src/utils/search-params.utils.ts
```diff
@@ -0,0 +1,6 @@
+export const urlSearchParamsToObject = (params: URLSearchParams) =>  {
+  return [ ...params.entries() ].reduce((acc, [ key, value ]) => {
+    acc[key] = value
+    return acc
+  }, {})
+}
```
wesbillman, line 6: You might be able to use here:

```ts
export const urlSearchParamsToObject = (params: URLSearchParams) => Object.fromEntries(params)
```

If this is similar we can maybe remove this helper/utils.

console/client/src/providers/tabs-provider.tsx
```diff
@@ -12,23 +12,65 @@ export type Tab = {
...
+      return false
+    }
+  }
+  return true
+}
 
 export const TabsProvider = (props: React.PropsWithChildren) => {
   const [ tabs, setTabs ] = React.useState<Tab[]>([ timelineTab ])
-  const [ activeTab, setActiveTab ] = React.useState<number>(0)
+  const [ activeTab, setActiveTab ] = React.useState<{id: string, type: string}| undefined>()
```
wesbillman, line 73: Minor style thing, but should we have a space before the pipe here?
```
React.useState<{id: string, type: string} | undefined>()
```

console/client/src/features/verbs/VerbTab.tsx
```diff
@@ -23,7 +24,7 @@ export const VerbTab: React.FC<Props> = ({ id }) => {
   }
 
   return (
-    <div className='min-w-0 flex-auto p-4'>
+    <div className='min-w-0 flex-auto'>
```
wesbillman, line 27: Do we want this change here? It looks like it makes the verb tab have no margin

Screenshot 2023-08-24 at 8 38 15 PM

console/client/src/features/modules/ModuleDetails.tsx
```diff
@@ -1,15 +1,30 @@
...
+  const moduleId = searchParams.get('module')
+  //P1 when mounting with a valid module in query params set selected module
+  React.useEffect(() => {
+    if(moduleId) {
+      const module = modules.modules.find(module => module?.name === moduleId)
+      module && setSelectedModule(module)
+    }
+  })
+  //P2 when query param module changes update selected module if valid
+  React.useEffect(() => {
```
wesbillman, line 22: Do you prefer the `React.useEffect` over just `useEffect` here? Is this something we can lint on? I wonder if there's a standard approach.

console/client/src/features/modules/ModuleDetails.tsx
EdwardIrby: Honestly I just like the letter P no reason other than maybe Priority. I use it to make notes about exception and edge case handling. I use P to differentiate from a basic comment notes describing functionality. Guess I could use Condition | Priority | Case | Exception specifically.

console/client/src/providers/tabs-provider.tsx
EdwardIrby: Not anymore as I stated in my comment above I need to add a error message componen we talked about this in stand up

console/client/src/features/verbs/VerbTab.tsx
EdwardIrby: Looks like padding it got dropped when I pulled in commits from main: https://github.com/TBD54566975/ftl/pull/320/commits/1a397f9aefaf1f1ba1d1dc357d57027da3911aa4#diff-7a08d7a5e05b5233e6546527ec691b6514a387909f944a33ab3e92df31dc680eL130

 Also yeah I mean to rip that off. We should apply consistent padding to all panels not just the verb panel.

console/client/src/features/modules/ModuleDetails.tsx
EdwardIrby: React isn't a ESM package destructuring doesn't do anything. Wether you import useEffect or React.useEffect you are getting the entire react library.  I use to destructure but stoped at Pluralsight when a an architect called this out and we moved to a no n destructuring pattern for react which is either UMD | CJS but never ESM.

console/client/src/utils/search-params.utils.ts
EdwardIrby: Yep that works

EdwardIrby:
https://github.com/TBD54566975/ftl/assets/1058725/cad5aeb1-a13d-468d-a27b-234726478bee

console/client/src/features/verbs/VerbTab.tsx
EdwardIrby: Added the padding back the was missing and adjusted padding on verb form selector

EdwardIrby:
https://github.com/TBD54566975/ftl/assets/1058725/15ac49bb-1e80-4c14-9a2d-d1ccbb6e2a34

console/client/src/layout/IDELayout.tsx
```diff
@@ -121,7 +133,7 @@ export function IDELayout() {
               </Tab.List>
               <div className='flex-grow'></div>
             </div>
-            <div className={`flex-1 overflow-y-scroll ${panelColor}`}>
+            <div className={`flex-1 overflow-y-scroll ${panelColor} p-4`}>
```
wesbillman, line 136: This `p-4` adds extra padding to the timeline view which causes items to appear above the bar
Screenshot 2023-08-25 at 9 22 12 AM

Main currently looks like this:
Screenshot 2023-08-25 at 9 25 53 AM

console/client/src/layout/IDELayout.tsx
```diff
@@ -133,8 +145,13 @@ export function IDELayout() {
...
         </div>
         <SidePanel />
-        <RequestModal />
       </div>
+      {invalidTabMessage && <Notification
+        color='text-red-400'
+        icon={<InformationCircleIcon className='flex-shrink-0 inline w-4 h-4 mr-3' />}
+        title='Info Alert!'
+        message={invalidTabMessage}
+      />}
```
wesbillman, line 154: Is there a good way for us to test this out? I tried typing in an invalid tab name, but I didn't see an error just an empty tab:
Screenshot 2023-08-25 at 9 27 58 AM

console/client/src/providers/tabs-provider.tsx
wesbillman: Did you see this one @EdwardIrby?

console/client/src/layout/IDELayout.tsx
wesbillman: This is what I get with `active-tab=payment` looks like the tab is kind of floating here:
Screenshot 2023-08-25 at 9 30 55 AM

console/client/src/features/modules/ModuleDetails.tsx
```diff
@@ -32,9 +41,12 @@ export function ModuleDetails() {
       }
       setTabs([ ...tabs, newTab ])
     }
-    setActiveTab(existingTab ? index : tabs.length)
-    searchParams.set(TabSearchParams.active, newTab?.id ?? tabs[index].id)
-    navigate({ ...location, search: searchParams.toString() })
+    setActiveTab({ id: tabId, type: TabType.Verb })
+    setSearchParams({
+      ...Object. fromEntries(searchParams),
```
wesbillman, line 46: Is there an extra space here? `...Object. fromEntries(searchParams),` seems like this should be `...Object.fromEntries(searchParams),`

Maybe the linter is missing a rule for this?

console/client/src/features/modules/ModulesList.tsx
```diff
@@ -22,8 +15,10 @@ export function ModulesList() {
         <li key={module.deploymentName}
           onClick={() => {
             setSelectedModule(prevModule => prevModule === module ? null : module)
-            searchParams.set('details', module.name)
-            navigate({ ...location, search: searchParams.toString() })
+            setSearchParams({
+              ...Object. fromEntries(searchParams),
```
wesbillman, line 19: Similar here: `...Object.fromEntries(searchParams),`

console/client/src/features/timeline/TimelineFilterBar.tsx
```diff
@@ -24,7 +24,7 @@ export const TimelineFilterBar = ({ selectedFilters, onFilterChange }) => {
   return (
     <>
       <div className={`sticky top-0 z-10 ${panelColor} shadow`}>
-        <div className='flex items-center justify-between p-4'>
+        <div className='flex items-center justify-between px-4'>
```
wesbillman, line 27: I think we want to keep this `p-4` so there's padding on the top and bottom of the filter bar as well

console/client/src/features/verbs/VerbCalls.tsx
```diff
@@ -14,10 +14,10 @@ type Props = {
 
 export const VerbCalls: React.FC<Props> = ({ module, verb }) => {
   const client = useClient(ConsoleService)
-  const [ calls, setCalls ] = useState<Call[]>([])
-  const { openPanel } = useContext(SidePanelContext)
+  const [ calls, setCalls ] = React.useState<Call[]>([])
+  const { openPanel } = React.useContext(SidePanelContext)
 
-  useEffect(() => {
+  React.useEffect(() => {
```
wesbillman, line 20: What's the reason for this type of change? Do we want to prefer `React.<thing>` style vs just importing what we want? I probably have a preference to just use `useState` here vs. `React.useState` whatcha think?

console/client/src/features/verbs/VerbTab.tsx
wesbillman: Yup! I'm very open to consistent padding here. If you have ideas I'd love to see some options. For this example, we don't pad all the panels the same because some of the have edge to edge components, like the filter bar on the timeline. So we don't pad the panel and let the embedded component control it's padding.

console/client/src/layout/IDELayout.tsx
EdwardIrby: > ? I tried typing in an invalid tab name, but I didn't see an error just an empty tab:

I showed how in the vide

EdwardIrby: As for payment I can't actually get that module to work locally

EdwardIrby: Okay so you image indicates you were not on commit https://github.com/TBD54566975/ftl/pull/320/commits/f2494dfac55b388a0dc64b3b531b60dc10b913a9

Which was pushed well before you comment. I'd recommend pull down the most recent version and trying that [url](http://localhost:8892/?module=payment&active-tab=payment) again https://github.com/TBD54566975/ftl/pull/320/commits/f2494dfac55b388a0dc64b3b531b60dc10b913a9 which will do nothing nor the weird render thing I can't recreate. 

A link that will error correctly is this http://localhost:8892/?module=payment&tab-type=verb which is missing tab-id

wesbillman: Yup, that makes sense since this commit didn't exist yet when I tested this. My comment was made before https://github.com/TBD54566975/ftl/commit/f2494dfac55b388a0dc64b3b531b60dc10b913a9 was done. 

I think there might be one remaining check on whether the URL is valid that should be added. Here's the URL I tested with `http://localhost:8892/?module=currency&tab-id=bogus.convert&tab-type=verb` and `http://localhost:8892/?module=currency&tab-id=currency.bogus&tab-type=verb` and the result is a blank tab with no error message.

Screenshot 2023-08-27 at 1 34 02 PM

examples/online-boutique/services/echo/generated_ftl_module.go
```diff
@@ -0,0 +1,21 @@
...
+type EchoResponse struct {
+  Message string `json:"message"`
+}
+
+// Echo returns a greeting with the current time.
+//
+//ftl:verb
+func Echo(context.Context, EchoRequest) (EchoResponse, error) {
+  panic("Verb stubs should not be called directly, instead use github.com/TBD54566975/ftl/runtime-go/sdk.Call()")
+}
```
wesbillman, line 21: I noticed this file and the file below were added in this PR, do we want them here?

console/client/src/layout/IDELayout.tsx
```diff
@@ -165,8 +202,17 @@ export function IDELayout() {
...
         <SidePanel />
-        <RequestModal />
       </div>
+      {invalidTabMessage && (
+        <Notification
+          color='text-red-400'
+          icon={
+            <InformationCircleIcon className='flex-shrink-0 inline w-4 h-4 mr-3' />
+          }
+          title='Info Alert!'
```
wesbillman, line 212: Should this just be `Alert` vs. `Info Alert`? Or is the intended message format?

Screenshot 2023-08-28 at 1 01 51 PM

examples/online-boutique/services/echo/generated_ftl_module.go
EdwardIrby: Nope rha should not be there

console/client/src/layout/IDELayout.tsx
EdwardIrby: Let's go with alert

console/client/.eslintignore
```diff
@@ -1,4 +1,4 @@
 src/protos/**/*
 dist/**/*
 .playwright/**/*
-playwright.config.ts
\ No newline at end of file
+playwright.config.ts
```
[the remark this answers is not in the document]
EdwardIrby: I don't think I did that actually pretty sure that was the formatter can try and add it back and see what happens.

console/client/src/layout/IDELayout.tsx
```diff
@@ -1,66 +1,109 @@
...
+      )
+      // Set tab if they both exists
+      if (moduleExist && verbExist) {
+        const newTab = {
+          id: moduleId,
+          label: verbId,
+          type: TabType.Verb,
+        }
+        const nextTabs = [...tabs, newTab]
+        return setTabs(nextTabs)
```
wesbillman, line 95: Should we be setting the active tab before returning here? This seems to leave the UI in an odd state.

Screenshot 2023-08-28 at 12 55 47 PM

EdwardIrby: yep it's missing added it back

console/client/.eslintignore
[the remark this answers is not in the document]
EdwardIrby: my local has a new line not sure why it doesn't here

console/client/src/layout/IDELayout.tsx
wesbillman: Hmm, weird. I still don't get the active tab set when I test with the latest code:

Is it because you're setting the tab before adding it to the tab list? Does it work ok for you? Maybe I'm missing something.

Screenshot 2023-08-28 at 12 55 47 PM

wesbillman (approved):

PR #324, opened 2023-08-25
eirby/formatting config

EdwardIrby:


console/client/src/features/modules/ModuleDetails.tsx
```diff
@@ -20,21 +25,21 @@ export function ModuleDetails() {
   }
 
   const handleVerbClicked = verb => {
-    const tabId = [ selectedModule.name, verb.verb?.name ].join('.')
+    const tabId = [selectedModule.name, verb.verb?.name].join('.')
```
wesbillman, line 28: Is there a way we can get the same level of type checking as the other PR. I would expect the linter catch things like this
```ts
const handleVerbClicked = verb => {
```
and want a type defined for verb
```ts
const handleVerbClicked = (verb: Verb) => {
```

Here's an example from the other PR: https://github.com/TBD54566975/ftl/pull/323/files#diff-9b87d665888695a1f97a6c98c8ad550eb1b24dcf08d9a56da5d22742c30a9274R23

This actually pointed out quite a few spots where we weren't defining types and instead defaulting to `any`
Screenshot 2023-08-27 at 1 54 31 PM

console/client/src/components/CodeBlock.tsx
```diff
@@ -1,20 +1,24 @@
...
+import {useDarkMode} from '../providers/dark-mode-provider'
 
 type Props = {
   code: string
   language: string
 }
 
-export const CodeBlock: React.FC<Props> = ({ code, language }) => {
-  const { isDarkMode } = useDarkMode()
+export const CodeBlock: React.FC<Props> = ({code, language}) => {
```
wesbillman, line 14: Another example where the linter could catch that we're not specifying type here.

The other PR changed this code to this, which feels a bit more explicit.
```ts
interface Props {
  code: string
  language: string
}

export const CodeBlock = ({ code, language }: Props) => {
```

wesbillman (approved):

console/client/src/features/modules/ModuleDetails.tsx
EdwardIrby: NP

PR #379, opened 2023-09-12
New navigation

wesbillman:
Screenshot 2023-09-12 at 12 06 23 PM
Screenshot 2023-09-12 at 12 06 28 PM
Screenshot 2023-09-12 at 12 17 26 PM


wesbillman:
@EdwardIrby here's the change I mentioned before. I still need to do all the work for filtering etc., but this get's the basic structure in there. I also left the Modules page mostly empty until we get around to styling it.

PR #382, opened 2023-09-14
fix: swap out parcel for vite to improve performance

wesbillman:


wesbillman:
@EdwardIrby got rid of parcel and added vite here

PR #383, opened 2023-09-14
fix: Revert material and mui

wesbillman:


wesbillman:
@EdwardIrby this is the material ui revert

PR #388, opened 2023-09-15
feat: refactor console and make events filters

wesbillman:


wesbillman:
@EdwardIrby this PR adds a `console.service.ts` file which wraps up the API to the console service, making it easier for components to use.

PR #390, opened 2023-09-15
feat: start hooking up time controls

wesbillman:
This isn't fully functional yet, but it had a few bug fixes and got some of it hooked up.

You can select a time range and use the forward and backward buttons to page through time ranges.
But, tailing doesn't work (it just gets the 100 latest items) and there are definitely some other weird ux things I need to update.

Screenshot 2023-09-15 at 4 14 42 PM
Screenshot 2023-09-15 at 4 14 46 PM

Screenshot 2023-09-15 at 4 14 38 PM


wesbillman:
cc: @EdwardIrby

PR #396, opened 2023-09-18
feat: hook up filters to steaming endpoint

wesbillman:
Timeline controls should be fully functional now


wesbillman:
@EdwardIrby FYI

PR #408, opened 2023-09-19
fix: timeline ordering when streaming events

### LINKED ISSUE #404: Timeline ordering is off with the new filters implemenation
New items are added in the wrong order.

wesbillman:
Fixes #404 

Also shorts the time range indicator and adds tooltips

Screenshot 2023-09-19 at 7 14 06 AM
Screenshot 2023-09-19 at 7 14 13 AM


wesbillman:
cc: @EdwardIrby timeline fix here.

Issue #361, opened 2023-09-06
Kotlin code generator has compilation errors when the data class name matches the verb class name


wesbillman:
Here's an example generated when the online-boutique is deployed.

```kotlin
// Code generated by FTL-Plugin, do not edit.
//
package ftl.ad

import kotlin.String
import kotlin.collections.ArrayList
import xyz.block.ftl.Context
import xyz.block.ftl.Ignore
import xyz.block.ftl.Verb
import xyz.block.ftl.v1.schema.DataRef

public data class AdRequest(
  public val contextKeys: ArrayList<String>,
)

public data class Ad(
  public val redirectURL: String,
  public val text: String,
)

public data class AdResponse(
  public val ads: ArrayList<DataRef>,
)

@Ignore
public class Ad() {
  @Verb
  public fun `get`(context: Context, req: AdRequest): AdResponse = throw
      NotImplementedError("Verb stubs should not be called directly, instead use context.call(Ad::get, ...)")
}
```

Here's the error:

```bash
~/ftl/examples/echo-kotlin/build/generated/source/ftl/ad/Ad.kt:16:19 Redeclaration: Ad
```

wesbillman:
Fixed with #427

PR #427, opened 2023-09-29
fix: optionally specify module class suffix for generated kotlin clients

worstell:
ftl #361


wesbillman:
Nice @worstell! this looks great! 

Are you able to test locally to see if changing the `suffix` here generates the new types correctly as well https://github.com/TBD54566975/ftl/blob/main/examples/echo-kotlin/pom.xml#L106-L109

wesbillman (approved):

worstell:
> Nice @worstell! this looks great!
> 
> Are you able to test locally to see if changing the `suffix` here generates the new types correctly as well https://github.com/TBD54566975/ftl/blob/main/examples/echo-kotlin/pom.xml#L106-L109

validated locally!

PR #432, opened 2023-10-03
feat(client): refactors layout to resolve scroll bugs

EdwardIrby:
- Refactors layout to fix scroll bugs
- Creates a Page component and subcomponents to resolve issues and create consistent way of creating new pages without introducing scroll related styling bugs
- @wesbillman notification style I dropped under the transition to prevent it from making the screen black as the notification component as a whole covers the whole screen

### DIFF
--- a/console/client/.eslintrc.cjs
+++ b/console/client/.eslintrc.cjs
@@ -24,6 +24,7 @@ module.exports = {
   rules: {
     'react/react-in-jsx-scope': 'off',
     'react/jsx-uses-react': 'off',
+    'react/prop-types': 'off',
     'func-style': ['error', 'expression'],
     '@typescript-eslint/consistent-type-definitions': ['error', 'interface'],
     '@typescript-eslint/no-unused-vars': ['warn', { argsIgnorePattern: '^_', varsIgnorePattern: '^_' }],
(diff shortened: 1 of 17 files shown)


EdwardIrby:
https://github.com/TBD54566975/ftl/assets/1058725/ff886851-e337-4790-9967-677d53733664

EdwardIrby:
https://github.com/TBD54566975/ftl/assets/1058725/ffe63c1f-c23d-4df8-bcda-d42d126bfd66

console/client/.eslintrc.cjs
```diff
@@ -24,6 +24,7 @@ module.exports = {
   rules: {
     'react/react-in-jsx-scope': 'off',
     'react/jsx-uses-react': 'off',
+    'react/prop-types': 'off',
```
wesbillman, line 27: Do we want to disable this rule? How does type checking work with this off?

wesbillman (approved):

console/client/.eslintrc.cjs
EdwardIrby: Yeah we do if we're using typescript as it's actually not necessary to lint for this.  Further it also allows to use the chained component sub-component pattern.

PR #439, opened 2023-10-04
 feat(client): modules page with basic features set

EdwardIrby:
- adds basic zoom to logic for verbs in the viz
- adds update styling
- adds timeline
- adds verb schemas and calls tabbed panel

**note:** there still some interactions and sizing on the svg I need to tighten up


EdwardIrby:
https://github.com/TBD54566975/ftl/assets/1058725/37dd1859-3f6e-4aab-ae06-6ce7d2eee7c3

console/client/src/features/modules/ModulesRequests.tsx
```diff
@@ -0,0 +1,124 @@
...
+import { formatTimestampShort } from '../../utils/date.utils.ts'
+import { panelColor } from '../../utils/style.utils.ts'
+import { Timestamp } from '@bufbuild/protobuf'
+
+const maxTimelineEntries = 1000
+
+export const ModulesRequests: React.FC<{
+  className?: string
+  modules: Module[]
+}> = ({ className, modules }) => {
```
wesbillman, line 26: This is good for now, but makes me think we might want a timeline component that we can pass in events to in the future so we don't have to duplicate all this code.

console/client/src/features/modules/ModulesSelectedVerbs.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+export type Schema = JSONSchema4 | JSONSchema6 | JSONSchema7
+
+interface Props {
+  module?: Module
+  verb?: Verb
+}
+
+const VerbForm = ({ module, verb }: Props) => {
+  const client = useClient(VerbService)
```
wesbillman, line 25: There's a `VerbForm` component in the project today, would that work for this?

console/client/src/features/modules/ModulesSidebar.tsx
```diff
@@ -0,0 +1,145 @@
...
+import { getNames } from './modules.utils'
+import { classNames } from '../../utils'
+import { backgrounds, borders, colors } from './components'
+
+interface MapValue {
+  deploymentName: string
+  id: string
+  verbs: Set<VerbId>
+  queriedVerbs: Set<VerbId>
+}
```
wesbillman, line 15: `MapValue` seems a bit generic, is there a better name we could use here? Maybe like `DeploymentVerbs` or something? I'm cool if this makes sense as well.

console/client/src/features/modules/ModulesRequests.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const [searchParams, setSearchParams] = useSearchParams()
+  const [entries, setEntries] = React.useState<Event[]>([])
+  const { openPanel, closePanel } = React.useContext(SidePanelContext)
+  const [selectedEntry, setSelectedEntry] = React.useState<Event | null>(null)
+  const deployments = modules.map(({ deploymentName }) => deploymentName)
+
+  const filters: EventsQuery_Filter[] = []
+  if (deployments.length) {
+    filters.push(modulesFilter(deployments))
+  }
```
wesbillman, line 36: This component is named `ModuleRequests`, so I just want to point out this is getting all events for a module. This would likely include `Logs`, `Calls`, `Deployments` etc. Is that what you're looking for here?

EdwardIrby: For the moment yes. I need to site down and and figure out how to move to a more request trace based log like in the wireframes. Talk about it tomorrow?

console/client/src/features/modules/ModulesSidebar.tsx
EdwardIrby: Honestly I started generic and never swapped in something better to describe it. Let's rename.

PR #472, opened 2023-10-10
feat: another option for deployment name colors

wesbillman:
Here's another option for styling deployment names with tailwind colors.

We can adjust this table as needed if we want/need more options. Also, added some reusable card stuff to make sure the colors stay somewhat consistent across the console.

Screenshot 2023-10-10 at 9 42 27 AM
Screenshot 2023-10-10 at 9 42 32 AM
Screenshot 2023-10-10 at 9 42 39 AM
Screenshot 2023-10-10 at 9 42 42 AM


alecthomas (approved):
Nice, that's better!

PR #519, opened 2023-10-25
fix: since we don't support optionals return empty array

wesbillman:
This fixes and issue with the online_boutique sample app crashing on a `null` array


alecthomas (approved):

wesbillman:
> What was the crash?

https://github.com/TBD54566975/ftl-flutter-online-boutique/blob/main/lib/api/cart.dart#L114

Since we don't yet support optionals and this was `null`

PR #593, opened 2023-11-14
fix: code gen actual dataref types

wesbillman:


kotlin-runtime/ftl-generator/src/main/kotlin/xyz/block/ftl/generator/ModuleGenerator.kt
```diff
@@ -162,7 +153,7 @@ class ModuleGenerator() {
       }
 
       type.verbRef != null -> ClassName("xyz.block.ftl.v1.schema", "VerbRef")
```
worstell, line 155: do we need to make this change for VerbRef too?

wesbillman: I'm actually not sure what we should do with `VerbRef`. Do we generate `VerbRef` things?

worstell (approved):

PR #752, opened 2024-01-09
feat: support for raw HTTP ingress

wesbillman:
This extends the current `ingress` system with support for multiple types,
the first of which is the original `ftl` HTTP transcoder, the second of
which is this raw HTTP ingress.

Raw HTTP ingress verbs must have the signature:

    <name>(builtin.HttpRequest) builtin.HttpResponse


kotlin-runtime/ftl-runtime/src/main/kotlin/xyz/block/ftl/schemaextractor/ExtractSchemaRule.kt
```diff
@@ -205,30 +205,32 @@ class SchemaExtractor(
...
+      val typeArg = requireNotNull(annotationEntry.valueArguments[0].getArgumentExpression()?.text) {
+        "$sourcePos Could not extract type from ${verb.name} @Ingress annotation"
       }
-      val methodArg = requireNotNull(argumentLists.first.single().getArgumentExpression()?.text) {
+      val methodArg = requireNotNull(annotationEntry.valueArguments[1].getArgumentExpression()?.text) {
         "$sourcePos Could not extract method from ${verb.name} @Ingress annotation"
       }
-      // NB: trim leading/trailing double quotes because KtStringTemplateExpression.text includes them
-      val pathArg = requireNotNull(argumentLists.second.single().getArgumentExpression()?.text?.trim { it == '\"' }) {
+      val pathArg = requireNotNull(annotationEntry.valueArguments[2].getArgumentExpression()?.text) {
```
wesbillman, line 221: @worstell you might want to check my work here. I def just made some assumptions, so want to make sure it looks ok to you :)

PR #813, opened 2024-01-18
fix: AbstractRef typenames

wesbillman:
Types like `AbstractRef[github.com/TBD54566975/ftl/protos/xyz/block/ftl/v1/schema.DataRef]` would return to the generate with `AbstractRef`

What we really want in the generator/templates is `DataRef` for this.


alecthomas (approved):
Thanks @wesbillman, my bad for not thinking of this when I introduced these type aliases.

PR #833, opened 2024-01-25
fix: add metadata to integration test

wesbillman:
I tried adding just `"m"` but it seemed to fail still. Not sure if that's expected or not @worstell


worstell:
> I tried adding just `"m"` but it seemed to fail still. Not sure if that's expected or not @worstell

this is expected because the integration test call is going through FTL HTTP and the alias is only transformed for ingress requests

worstell:
oof sorry also now realizing i accidentally checked in the metadata field on echo.kt from local testing 😅 gonna revert that

PR #867, opened 2024-02-01
feat: remove FTL ingress in favor of a single HTTP ingress type

worstell:
We will only support a single HTTP ingress type. If no ingress type is specified in a schema declaration, the default type will be HTTP.


kotlin-runtime/ftl-runtime/src/test/kotlin/xyz/block/ftl/schemaextractor/testdata/dependencies/BuiltinModuleClient.kt
```diff
@@ -0,0 +1,33 @@
+// Code generated by FTL-Generator, do not edit.
```
worstell, line 1: added this `testdata/dependencies` directory which gets loaded up and added to the classpath for the detekt tests with type resolution. we can use it to add generated sources like`ftl.builtin` and `ftl.time`, ensuring these types get resolved in tests

kotlin-runtime/ftl-runtime/src/test/kotlin/xyz/block/ftl/schemaextractor/ExtractSchemaRuleTest.kt
```diff
@@ -16,12 +18,20 @@ import kotlin.test.AfterTest
 
 @KotlinCoreEnvironmentTest
 internal class ExtractSchemaRuleTest(private val env: KotlinCoreEnvironment) {
+  @BeforeEach
```
worstell, line 21: here's where we add those `testdata/dependencies` files to the classpath

alecthomas (approved):

kotlin-runtime/ftl-runtime/src/test/kotlin/xyz/block/ftl/schemaextractor/ExtractSchemaRuleTest.kt
wesbillman: Nice!

PR #904, opened 2024-02-08
feat: use builtin.Empty instead of Unit typealias in Kotlin

### LINKED ISSUE #878: Figure out how to create an empty response in Kotlin
If we want to have an http response that returns and empty json payload like `{}` how can we define that structure in Kotlin.

Assuming an ingress verb with this signature:

```kotlin
@Verb
@Ingress(Method.PUT, "/users/{userId}", HTTP)
fun put(context: Context, req: HttpRequest<PutRequest>): HttpResponse<PutResponse> {
  return HttpResponse<PutResponse>(
    status = 200,
    headers = mapOf("Put" to arrayListOf("Header from FTL")),
    body = PutResponse()
  )
}
```

### Using Unit:

```kotlin
typealias PutResponse = Unit
```

results in a `response.body` of  empty bytes (like a void response)

### Using empty data class

```kotlin
// Ideally we could create an empty data class
data class PutResponse()

// but since that's not possible we end up with
data class PutResponse(val _empty: Unit = Unit)
```

When `data class PutResponse(val _empty: Unit = Unit)` is serialized it results in:
```json
{
  "_empty": {},
}
```
Which is also not correct.

worstell:
fixes #878

### DIFF
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -96,12 +96,12 @@ func TestHttpIngress(t *testing.T) {
 				httpCall(rd, http.MethodPut, "/users/123", jsonData(t, obj{"postId": "346"}), func(t testing.TB, resp *httpResponse) {
 					assert.Equal(t, 200, resp.status)
 					assert.Equal(t, []string{"Header from FTL"}, resp.headers["Put"])
-					assert.Equal(t, nil, resp.body)
+					assert.Equal(t, map[string]any{}, resp.body)
 				}),
 				httpCall(rd, http.MethodDelete, "/users/123", jsonData(t, obj{}), func(t testing.TB, resp *httpResponse) {
 					assert.Equal(t, 200, resp.status)
 					assert.Equal(t, []string{"Header from FTL"}, resp.headers["Delete"])
-					assert.Equal(t, nil, resp.body)
+					assert.Equal(t, map[string]any{}, resp.body)
 				}),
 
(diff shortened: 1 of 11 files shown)


integration/integration_test.go
```diff
@@ -96,12 +96,12 @@ func TestHttpIngress(t *testing.T) {
...
 					assert.Equal(t, []string{"Header from FTL"}, resp.headers["Put"])
-					assert.Equal(t, nil, resp.body)
+					assert.Equal(t, map[string]any{}, resp.body)
 				}),
 				httpCall(rd, http.MethodDelete, "/users/123", jsonData(t, obj{}), func(t testing.TB, resp *httpResponse) {
 					assert.Equal(t, 200, resp.status)
 					assert.Equal(t, []string{"Header from FTL"}, resp.headers["Delete"])
-					assert.Equal(t, nil, resp.body)
+					assert.Equal(t, map[string]any{}, resp.body)
 				}),
```
wesbillman, line 105: Maybe we should leave one of these `Unit` so we can ensure that `nil` works as well?

wesbillman (approved):

integration/integration_test.go
worstell: the `/empty` path still tests Unit!

```
httpCall(rd, http.MethodGet, "/empty", nil, func(t testing.TB, resp *httpResponse) {
	assert.Equal(t, 200, resp.status)
	assert.Equal(t, nil, resp.headers["Content-Type"])
	assert.Equal(t, nil, resp.bodyBytes)
}),
```

PR #939, opened 2024-02-14
fix: make sure to log spawn errors at log.Error

wesbillman:
Example of an error on `var db = ftl.PostgresDatabase("not_a_db")`

Screenshot 2024-02-14 at 3 25 52 PM


alecthomas (approved):
I do remember why I did this now: it's because random libraries in Go will output log messages onto stderr. I think from memory that opentelemetry might have been an offender.

I also remember now why I was suggesting using a new FD: because if a random lib outputs valid JSON onto stderr, the logger will try to parse it. But meh, problem for later.

PR #977, opened 2024-02-23
fix: make sure `ftl serve --background` waits for controller to start

wesbillman:
Example:

```sh
ftl serve --background && ftl status
info: FTL running in background with pid: 26990

error: Unary RPC failed: unavailable: dial tcp [::1]:8892: connect: connection refused: /xyz.block.ftl.v1.ControllerService/Status
{
  "controllers": [
    {
      "key": "C01HQBDJSK0C249VV2QBN3B856J",
      "endpoint": "http://localhost:8892"
    }
  ],
  "runners": [
...
```

We'll probably want a way in the future to not print the `error` in the status request, but it shouldn't stop the Integration/CI flow from working here.

### DIFF
--- a/cmd/ftl/cmd_serve.go
+++ b/cmd/ftl/cmd_serve.go
@@ -13,12 +13,15 @@ import (
 	"syscall"
 	"time"
 
+	"connectrpc.com/connect"
 	"github.com/alecthomas/kong"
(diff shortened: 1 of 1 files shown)


wesbillman:
Fixes #962

PR #989, opened 2024-02-27
chore: delete kotlin runtime ftl-generator

worstell:


alecthomas (approved):
OMG that is so satisfying!

PR #1107, opened 2024-03-18
feat: handle enums in jsonschema

### LINKED ISSUE #1099: Add `Enum` handling to `jsonschema`
It doesn't seem like this is currently being tested in `jsonschema_test.go`

worstell:
fixes #1099

### DIFF
--- a/backend/schema/jsonschema.go
+++ b/backend/schema/jsonschema.go
@@ -19,42 +19,40 @@ func DataToJSONSchema(sch *Schema, ref Ref) (*jsonschema.Schema, error) {
 		return nil, fmt.Errorf("unknown data type %s", ref)
 	}
 
-	// Collect all data types.
-	dataTypes := sch.DataMap()
-
 	// Encode root, and collect all data types reachable from the root.
 	refs := map[RefKey]*Ref{}
-	root := nodeToJSSchema(sch, data, refs)
+	root := nodeToJSSchema(data, refs)
 	if len(refs) == 0 {
 		return root, nil
 	}
 
-	// Resolve and encode all data types reachable from the root.
+	// Resolve and encode all types reachable from the root.
 	root.Definitions = map[string]jsonschema.SchemaOrBool{}
-	for key, r := range refs {
-		data, ok := dataTypes[RefKey{Module: key.Module, Name: key.Name}]
-		if !ok {
-			return nil, fmt.Errorf("unknown data type %s", key)
-		}
+	for _, r := range refs {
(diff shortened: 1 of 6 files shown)


wesbillman (approved):
Nice!!

Does the console work ok with this change as well? I wonder if we can deploy a module with enums and see how the editor works in the enum case. :)

backend/schema/jsonschema.go
```diff
@@ -19,42 +19,38 @@ func DataToJSONSchema(sch *Schema, ref Ref) (*jsonschema.Schema, error) {
...
+				root.Definitions[r.String()] = jsonschema.SchemaOrBool{TypeObject: nodeToJSSchema(n, refs)}
 			}
-			data = monomorphisedData
-
-			ref := fmt.Sprintf("%s.%s", r.Module, refName(r))
-			root.Definitions[ref] = jsonschema.SchemaOrBool{TypeObject: nodeToJSSchema(sch, data, refs)}
-		} else {
-			root.Definitions[r.String()] = jsonschema.SchemaOrBool{TypeObject: nodeToJSSchema(sch, data, refs)}
+		case *Enum:
+			root.Definitions[r.String()] = jsonschema.SchemaOrBool{TypeObject: nodeToJSSchema(n, refs)}
```
wesbillman, line 47: Is `Enum` the only `ref` we need to special case here? What makes enum special here? (mostly just so I understand better)

worstell: this logic is essentially universal: 
```
if len(r.TypeParameters) > 0 {
  monomorphisedData, err := n.Monomorphise(r)
  if err != nil {
    return nil, err
  }

  ref := fmt.Sprintf("%s.%s", r.Module, refName(r))
  root.Definitions[ref] = jsonschema.SchemaOrBool{TypeObject: nodeToJSSchema(monomorphisedData, refs)}
} else {
  root.Definitions[r.String()] = jsonschema.SchemaOrBool{TypeObject: nodeToJSSchema(n, refs)}
}
```

but `len(r.TypeParameters) > 0` will only evaluate true when the type is `Data`. this was just my attempt at making this section a bit more elegent — `Monomorphise` is a member function on the `Data` class, so without this switch we'd need to do a type assertion that the resolved type is `Data` to perform this call: `n.Monomorphise(r)`.

the switch also explicitly fails for unhandled Decl types which is nice (*Config, *Database, *Secret, *Verb)

alecthomas (approved):
Nice

PR #1110, opened 2024-03-19
fix: generated external modules don't import from metadata refs

worstell:
when we generate an external module, we stub the verb bodies. we shouldn't add imports which are only dependencies of the verb expression body because this import may be unused once the body is stubbed, causing build errors.


wesbillman (approved):
Nice! Thank you! :)

matt2e:
Does this solve this issue? https://github.com/TBD54566975/ftl/issues/1095

matt2e:
nevermind, that other issue still happens

PR #1147, opened 2024-03-27
fix: local enum refs use qualified module name

worstell:


wesbillman (approved):
Nice! This will help on the other repo.

PR #1162, opened 2024-04-03
feat: post to language server from log stream

worstell:


buildengine/engine.go
```diff
@@ -30,8 +30,6 @@ type schemaChange struct {
 }
 
 type Listener struct {
-	OnBuildComplete  func(project Project, err error)
-	OnDeployComplete func(project Project, err error)
 }
```
wesbillman, line 33: We can probably remove the whole Listener thing here since we're not using it anymore.

wesbillman (approved):

buildengine/engine.go
worstell: cool, wasn't sure if we were still going to use with a new function to refresh diagnostics with in your change but will remove!

PR #1180, opened 2024-04-05
feat: pass kotlin build errors to the language server

### LINKED ISSUE #1123: normalize stderr output across go/kotlin
need to ensure stderr output for FTL build errors is structured in a normalized way so that it can be ingested by the LSP. we will use this output to provide the LSP with positions for document highlighting

### LINKED ISSUE #1159: Ability to handle errors from any language runtime
Ideas: For handling errors from all language runtimes

- Create GRPC server and runtimes push errors back as they build
- Come up with some protocol that we can use to talk with builds over stderr
output json and that json has the structured errors in them
probably the easiest approach

worstell:
although we've temporarily suspended development on the kotlin runtime, this was a useful proof-of-concept to demonstrate how we can build non-Go runtimes to interact with the language server

fixes #1123
fixes #1159

Screenshot 2024-04-04 at 3 23 15 PM

### DIFF
--- a/internal/exec/exec.go
+++ b/internal/exec/exec.go
@@ -2,7 +2,6 @@ package exec
 
 import (
 	"context"
-	"fmt"
 	"os"
 	"os/exec" //nolint:depguard
 	"syscall"
@@ -59,7 +58,7 @@ func (c *Cmd) RunBuffered(ctx context.Context) error {
 
 	err := c.Run()
 	if err != nil {
-		fmt.Printf("%s", outputBuffer.Bytes())
+		log.FromContext(ctx).Infof("%s", outputBuffer.Bytes())
 		return err
 	}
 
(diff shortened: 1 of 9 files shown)


internal/exec/exec.go
```diff
@@ -59,7 +58,7 @@ func (c *Cmd) RunBuffered(ctx context.Context) error {
 
 	err := c.Run()
 	if err != nil {
-		fmt.Printf("%s", outputBuffer.Bytes())
+		log.FromContext(ctx).Infof("%s", outputBuffer.Bytes())
```
worstell, line 61: cc @wesbillman

alecthomas (commented):
Nice, I like this approach. Let's convert the Go runtime to use this mechanism in a followup well so they're consistent.

worstell:
> Nice, I like this approach. Let's convert the Go runtime to use this mechanism in a followup well so they're consistent.

good call!

PR #1195, opened 2024-04-08
fix: handle nil pointers in deepcopy

matt2e:
Previously a deepcopy involving nil pointers would end up with a pointer to a zero value.

Needed as part of https://github.com/TBD54566975/ftl/pull/1155


alecthomas (approved):
It would be good if this had some tests?

PR #1257, opened 2024-04-14
chore: clean up string interpolation

matt2e:
I think I was getting caught out by copilot mimicking nearby odd string interpolation, so I figured a clean up might help avoid it in the future.


alecthomas (approved):
Yeah these were the result of a mechanical refactoring unfortunately

PR #1287, opened 2024-04-17
fix: return all errors from module build

worstell:


buildengine/build.go
```diff
@@ -45,15 +45,16 @@ func buildModule(ctx context.Context, sch *schema.Schema, module Module) error {
 	}
 
 	if err != nil {
+		errs := errors.UnwrapAllExcludingIntermediaries(err)
```
worstell, line 48: added `UnwrapAllExcludingIntermediaries` because we're basically decomposing this error and then rejoining it (L:59 `errors.Join(errs...)`), so the presence of the intermediate errors creates a bunch of duplication/spam

PR #1289, opened 2024-04-17
fix: maybe fix flaky watch test

worstell:
increasing the watch period frequency relative to the test sleep periods, in case we're encountering races causing the flakiness


worstell:
cc @wesbillman 🤷‍♀️

PR #1317, opened 2024-04-22
feat: module tests can declare config/secrets/DSNs and add verb mocks

matt2e:
Part of: https://hackmd.io/zNiBPIhVQN2hkeNHg2r8YA

This PR makes it easy to set up the following:
- config
- secrets
- DSNs
- mocks for verbs

```go
"github.com/TBD54566975/ftl/go-runtime/ftltest"

func TestEcho(t *testing.T) {
    ctx := ftltest.Context (
        ftltest.WithConfig("default", "anonymous"),
        ftltest.WithSecret("example", "test123"),
        ftltest.WithDSN("db", ftltest.DBTypePostgres, "..."),
        ftltest.WhenVerb(time.Time, func(ctx context. Context, req time. TimeRequest) (time.TimeResponse, error) {
            return time.TimeResponse{Time: stdtime.Date(2021, 9, 1, 0, 0, 0, 0, stdtime.UTC)}, nil
        },
    )

    ...
}
```

Another improvement is when doing unit tests, a simple error message will be returned when attempting to make a `ftl.Call` when there isn't any rpc client set up. This error will be returned:
> time.time: no mock found


go-runtime/ftl/call_overrider.go
```diff
@@ -0,0 +1,21 @@
+package ftl
+
+import (
+	"context"
+)
+
+type CallOverrider interface {
```
matt2e, line 7: I didn't want this to be public to module code, but wasn't sure of a better place to put it...

go-runtime/ftl/types.go
```diff
@@ -43,8 +43,8 @@ func (v *Ref) UnmarshalText(text []byte) error {
 	return nil
 }
 
-func (v *Ref) String() string { return v.Module + "." + v.Name }
-func (v *Ref) ToProto() *schemapb.Ref {
+func (v Ref) String() string { return v.Module + "." + v.Name }
```
matt2e, line 46: this is because Ref is used everywhere but printf wasn't formatting it nicely

go-runtime/ftl/ftltest/mock.go
```diff
@@ -0,0 +1,42 @@
...
+
+	"github.com/TBD54566975/ftl/backend/protos/xyz/block/ftl/v1/ftlv1connect"
+)
+
+type mockFunc func(ctx context.Context, req any) (resp any, err error)
+
+// mockProvider keeps a mapping of verb references to mock functions.
+//
+// It implements the CallOverrider interface to intercept calls with the mock functions.
+type mockProvider struct {
```
deniseli, line 18: Would it be worth renaming this to mockVerbProvider in case we want to support mocking of entities other than verbs down the road?

go-runtime/ftl/ftltest/ftltest.go
```diff
@@ -3,18 +3,94 @@ package ftltest
...
 
-// Context suitable for use in testing FTL verbs.
-func Context() context.Context {
+type DBType int32
+
+const (
+	DBTypePostgres = DBType(modulecontext.DBTypePostgres)
+)
+
+// Context suitable for use in testing FTL verbs with provided options
```
deniseli, line 22: super nitty nit: would you mind adding some example code to this comment so that folks don't need to grep around to learn how to use this? Basically exactly what you put in your PR description:

```
ctx := ftltest.Context (
        ftltest.WithConfig("default", "anonymous"),
        ftltest.WithSecret("example", "test123"),
        ftltest.WithDSN("db", ftltest.DBTypePostgres, "..."),
        ftltest.WithFakeVerb(time.Time, func(ctx context. Context, req time. TimeRequest) (time.TimeResponse, error) {
            return time.TimeResponse{Time: stdtime.Date(2021, 9, 1, 0, 0, 0, 0, stdtime.UTC)}, nil
        },
    )
```

Definitely don't need to include every single option in the example if you want to keep the comment shorter, either. Thanks :)

deniseli (approved):
super useful addition! thanks :D

PR #1332, opened 2024-04-25
fix: runInBackground returns error

deniseli:
Fixes https://github.com/TBD54566975/ftl/issues/1170

The new terminal output, when I force an error to return from `runInBackground`, looks like:

```
$ build/release/ftl serve --background
ftl: error: mock error!
```

### DIFF
--- a/cmd/ftl/cmd_serve.go
+++ b/cmd/ftl/cmd_serve.go
@@ -55,7 +55,9 @@ func (s *serveCmd) Run(ctx context.Context) error {
 			_ = KillBackgroundServe(logger)
 		}
 
-		runInBackground(logger)
+		if err := runInBackground(logger); err != nil {
+			return err
+		}
 
 		err := waitForControllerOnline(ctx, s.StartupTimeout, client)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


cmd/ftl/cmd_serve.go
```diff
@@ -55,7 +55,10 @@ func (s *serveCmd) Run(ctx context.Context) error {
 			_ = KillBackgroundServe(logger)
 		}
 
-		runInBackground(logger)
+		if err := runInBackground(logger); err != nil {
+			logger.Errorf(err, "runInBackground failed")
```
wesbillman, line 59: You might be able to remove the log message since the cmd will error as well. Otherwise, I think we might get duplicate error messages like this:
Screenshot 2024-04-25 at 10 57 11 AM

wesbillman (approved):

cmd/ftl/cmd_serve.go
deniseli: removed

alecthomas (commented):
Nice

PR #1337, opened 2024-04-26
fix: simplify config/secret duplication validation logic

deniseli:
Fixes https://github.com/TBD54566975/ftl/issues/1121

* Addresses comments in PR https://github.com/TBD54566975/ftl/pull/1336
* Adds additional duplication error case for configs/secrets with the same name but different types, which also simplifies the overall validation logic


alecthomas (approved):

matt2e (approved):
LGTM 👍

Issue #1339, opened 2024-04-26
inline global secrets don't seem to work


mistermoe:
## Repro

1. Set the following secrets
```bash
❯ echo -n abcd | ftl secret set keychain --keychain
❯ echo -n abcd | ftl secret set inline --inline
```

2. Get them as a sanity check
```bash
❯ ftl secret get keychain
"abcd"
❯ ftl secret get inline
"abcd"
```

3. Check resulting `ftl-project.toml`

> [!NOTE]
> yours may have more. i reproduced this in a brand new project

```toml
ftl-min-version = ""

[global]
  [global.secrets]
    inline = "inline://ImFiY2Qi"
    keychain = "keychain://keychain"

[executables]
  ftl = ""

[commands]
```

4. chuck the following into an ftl module:
```go
package debug

import (
	"context"

	"github.com/TBD54566975/ftl/go-runtime/ftl"
)

var inlineSecHandle = ftl.Secret[string]("inline")
var keychainSecHandle = ftl.Secret[string]("keychain")

type RivetingRequest struct{}
type RivetingResponse struct{}

//ftl:export
func Rivet(ctx context.Context, req RivetingRequest) (RivetingResponse, error) {
	log := ftl.LoggerFromContext(ctx)
	inlineSec := inlineSecHandle.Get(ctx)
	keychainSec := keychainSecHandle.Get(ctx)

	log.Infof("inline secret: %s", inlineSec)
	log.Infof("keychain secret: %s", keychainSec)

	return RivetingResponse{}, nil
}
```

5. pop open ftl console and send a request to verb

Output:
```bash
❯ ftl dev modules/debug --recreate
info: Starting FTL with 1 controller(s)
info:debug: Building module
info:debug: Deploying module
info:controller0: Deployed dpl-debug-cfk4q5bt7h77l7g
info: All modules deployed, watching for changes...
error:debug: panic in verb debug.rivet: failed to get secret "debug.inline": key "inline" not found
```

> [!IMPORTANT]
> the same will happen with `keychain` secret if you comment out `inline` secret references

```bash
error:debug: panic in verb debug.rivet: failed to get secret "debug.keychain": key "keychain" not found
```

hmm actually now that i'm trying even with the module specifier, it doesn't seem to work:

```bash
❯ ftl secret set debug.inline --inline
Secret:

❯ ftl secret get debug.inline
"abcd"
```

still getting:

```bash
[...]

PR #1340, opened 2024-04-26
fix: fall back to global secrets was not working

matt2e:
fixes https://github.com/TBD54566975/ftl/issues/1339

### DIFF
--- a/common/configuration/in_memory_provider.go
+++ b/common/configuration/in_memory_provider.go
@@ -2,7 +2,6 @@ package configuration
 
 import (
 	"context"
-	"fmt"
 	"net/url"
 )
 
@@ -18,13 +17,13 @@ func NewInMemoryProvider[R Role]() *InMemoryProvider[R] {
 }
 
 func (p *InMemoryProvider[R]) Role() R     { var r R; return r }
-func (p *InMemoryProvider[R]) Key() string { return "grpc" }
+func (p *InMemoryProvider[R]) Key() string { return "inmemory" }
 
 func (p *InMemoryProvider[R]) Load(ctx context.Context, ref Ref, key *url.URL) ([]byte, error) {
 	if bytes, found := p.values[ref]; found {
 		return bytes, nil
 	}
-	return nil, fmt.Errorf("key %q not found", ref.Name)
+	return nil, ErrNotFound
 }
 
 func (p *InMemoryProvider[R]) Writer() bool {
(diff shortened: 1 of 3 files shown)


common/configuration/in_memory_provider.go
```diff
@@ -18,13 +18,13 @@ func NewInMemoryProvider[R Role]() *InMemoryProvider[R] {
...
 func (p *InMemoryProvider[R]) Role() R     { var r R; return r }
-func (p *InMemoryProvider[R]) Key() string { return "grpc" }
+func (p *InMemoryProvider[R]) Key() string { return "inmemory" }
 
 func (p *InMemoryProvider[R]) Load(ctx context.Context, ref Ref, key *url.URL) ([]byte, error) {
 	if bytes, found := p.values[ref]; found {
 		return bytes, nil
 	}
-	return nil, fmt.Errorf("key %q not found", ref.Name)
+	return nil, ErrNotFound
```
matt2e, line 27: This specific error must be used

common/configuration/manager.go
```diff
@@ -84,9 +84,8 @@ func (m *Manager[R]) getData(ctx context.Context, ref Ref) ([]byte, error) {
 	key, err := m.resolver.Get(ctx, ref)
 	// Try again at the global scope if the value is not found in module scope.
 	if ref.Module.Ok() && errors.Is(err, ErrNotFound) {
-		gref := ref
-		gref.Module = optional.None[string]()
-		key, err = m.resolver.Get(ctx, gref)
+		ref.Module = optional.None[string]()
```
matt2e, line 87: If global config/secret is found, we need the `ref` var to point to it as it is used to access it from the provider.
Previously, resolver would match the global ref, but we would then use the module specific ref with the provider.

common/configuration/in_memory_provider.go
```diff
@@ -18,13 +18,13 @@ func NewInMemoryProvider[R Role]() *InMemoryProvider[R] {
 }
 
 func (p *InMemoryProvider[R]) Role() R     { var r R; return r }
-func (p *InMemoryProvider[R]) Key() string { return "grpc" }
+func (p *InMemoryProvider[R]) Key() string { return "inmemory" }
```
matt2e, line 21: clean up for an old refactor

deniseli:
Thanks for the comments! Very helpful in understanding the changes :)

PR #1355, opened 2024-04-29
chore: revert obsolete sumtype decl

deniseli:
Reverts everything in https://github.com/TBD54566975/ftl/pull/1322 except for the addition of `=` to `Punct` in `backend/schema/parser.go`, which we will still be using for value enums.

Going forward, sumtypes will be implemented as type enums.


wesbillman (approved):
RIP 😂

PR #1389, opened 2024-05-02
fix: display db type after 'database'

worstell:
displaying attributes after the resource identifier for consistency with other resources


deniseli:
Yay!!

PR #1417, opened 2024-05-06
fix: fix several go-runtime JSON encoding issues

deniseli:
This PR:
* Rips out the existing textMarshaler and jsonMarshaler usage from encoding.go. We may want to add those back for https://github.com/TBD54566975/ftl/issues/1296 down the road, but we will need to be thoughtful about how we do that. Removing it for now keeps the logic much more predictable.
* Moves the (un)marshaling logic for `ftl.Option` out of `option.go` and into `encoding.go`.
* Special-cases both `time.Time` (the only stdlib type we currently support) and `ftl.Option`. Also `json.RawMessage` for _just_ encoding to preserve the existing `omitempty` behavior.
* Fixes some existing issues where the Pointer unmarshaling wasn't actually working correctly
* [eww] Adds a rather grotesque alternative to `Peek()` in `isNextTokenNull()` because json Decoder does not support Peek. 
* [eww] Makes the ftl.Option struct fields public so that they are settable by `encoding.go`.

Suggestions welcome for both counts of [eww] above :)

Fixes https://github.com/TBD54566975/ftl/issues/1247.
Addresses most of https://github.com/TBD54566975/ftl/issues/1262, except `omitempty` is only working for json.RawMessage for now.

### DIFF
--- a/go-runtime/encoding/encoding.go
+++ b/go-runtime/encoding/encoding.go
@@ -4,20 +4,16 @@ package encoding
 
 import (
 	"bytes"
-	"encoding"
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"io"
 	"reflect"
+	"strings"
+	"time"
+	"unicode"
 
 	"github.com/TBD54566975/ftl/backend/schema/strcase"
-)
-
-var (
-	textMarshaler   = reflect.TypeOf((*encoding.TextMarshaler)(nil)).Elem()
-	textUnmarshaler = reflect.TypeOf((*encoding.TextUnmarshaler)(nil)).Elem()
-	jsonMarshaler   = reflect.TypeOf((*json.Marshaler)(nil)).Elem()
-	jsonUnmarshaler = reflect.TypeOf((*json.Unmarshaler)(nil)).Elem()
 )
 
 func Marshal(v any) ([]byte, error) {
@@ -31,37 +27,29 @@ func encodeValue(v reflect.Value, w *bytes.Buffer) error {
 		w.WriteString("null")
 		return nil
 	}
+
 	t := v.Type()
+
+	// Special-cased types
 	switch {
-	case t.Kind() == reflect.Ptr && t.Elem().Implements(jsonMarshaler):
-		v = v.Elem()
-		fallthrough
-
-	case t.Implements(jsonMarshaler):
-		enc := v.Interface().(json.Marshaler) //nolint:forcetypeassert
-		data, err := enc.MarshalJSON()
+	case t == reflect.TypeFor[time.Time]():
+		data, err := json.Marshal(v.Interface().(time.Time))
 		if err != nil {
 			return err
 		}
 		w.Write(data)
 		return nil
 
-	case t.Kind() == reflect.Ptr && t.Elem().Implements(textMarshaler):
-		v = v.Elem()
-		fallthrough
-
-	case t.Implements(textMarshaler):
-		enc := v.Interface().(encoding.TextMarshaler) //nolint:forcetypeassert
-		data, err := enc.MarshalText()
-		if err != nil {
-			return err
-		}
-		data, err = json.Marshal(string(data))
+	case t == reflect.TypeFor[json.RawMessage]():
+		data, err := json.Marshal(v.Interface().(json.RawMessage))
 		if err != nil {
 			return err
 		}
 		w.Write(data)
 		return nil
+
+	case isOption(v.Type()):
+		return encodeOption(v, w)
 	}
 
 	switch v.Kind() {
@@ -105,6 +93,24 @@ func encodeValue(v reflect.Value, w *bytes.Buffer) error {
 	default:
 		panic(fmt.Sprintf("unsupported type: %s", v.Type()))
 	}
+}
+
+var ftlOptionTypePath = "github.com/TBD54566975/ftl/go-runtime/ftl.Option"
+
+func isOption(t reflect.Type) bool {
+	return strings.HasPrefix(t.PkgPath()+"."+t.Name(), ftlOptionTypePath)
+}
+
+func encodeOption(v reflect.Value, w *bytes.Buffer) error {
+	if v.NumField() != 2 {
+		return fmt.Errorf("value cannot have type ftl.Option since it has %d fields rather than 2: %v", v.NumField(), v)
+	}
+	optionOk := v.Field(1).Bool()
+	if !optionOk {
+		w.WriteString("null")
+		return nil
+	}
+	return encodeValue(v.Field(0), w)
 }
 
 func encodeStruct(v reflect.Value, w *bytes.Buffer) error {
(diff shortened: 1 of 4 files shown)


go-runtime/encoding/encoding.go
```diff
@@ -213,49 +220,41 @@ func Unmarshal(data []byte, v any) error {
...
+		return decodeOption(d, v)
 	}
 
 	switch v.Kind() {
 	case reflect.Struct:
 		return decodeStruct(d, v)
 
 	case reflect.Ptr:
-		if token, err := d.Token(); err != nil {
+		allBytes, err := io.ReadAll(d.Buffered())
```
worstell, line 242: maybe move this into `isNextTokenNull`?

go-runtime/encoding/encoding.go
```diff
@@ -278,6 +277,59 @@ func decodeValue(d *json.Decoder, v reflect.Value) error {
 	}
 }
 
+// isNextTokenNull implements a cheap/dirty version of `Peek()`, which json.Decoder does
+// not support. This function is used to conditionally execute logic for handling nulls
+// without popping the next token, so that downstream code for the non-null case is still
+// able to pop it. Remember to pop the next token (d.Token()) inside your conditional
```
worstell, line 283: nit: i wonder if we should also handle consuming this token in `isNextTokenNull` (and maybe rename/change signature as needed) so callers don't need to manage this

i don't feel strongly!

worstell (approved):

go-runtime/encoding/encoding.go
deniseli: ooooo callbacks! Done! Looks way better now :D

go-runtime/encoding/encoding.go
deniseli: nice catch - fixed!

go-runtime/encoding/encoding.go
```diff
@@ -278,6 +268,63 @@ func decodeValue(d *json.Decoder, v reflect.Value) error {
...
+			return err
+		}
+		return nil
+	}
+	return elseFn(d)
+}
+
+// isNextTokenNull implements a cheap/dirty version of `Peek()`, which json.Decoder does
+// not support.
+func isNextTokenNull(d *json.Decoder) (bool, error) {
```
alecthomas, line 293: This feels pretty hacky. I'd be tempted to write a thin layer over `json.Decoder` that supports peeking, rather than doing this.

go-runtime/ftl/option.go
```diff
@@ -5,95 +5,90 @@ import (
...
 	fmt.GoStringer
-	json.Marshaler
-	json.Unmarshaler
 }
 
 // An Option type is a type that can contain a value or nothing.
 type Option[T any] struct {
-	value T
-	ok    bool
+	Val  T
```
alecthomas, line 20: We _really_ shouldn't be doing this, it violates encapsulation completely and it's a public type exposed to end users.

go-runtime/encoding/encoding.go
```diff
@@ -31,37 +27,29 @@ func encodeValue(v reflect.Value, w *bytes.Buffer) error {
...
-		fallthrough
-
-	case t.Implements(textMarshaler):
-		enc := v.Interface().(encoding.TextMarshaler) //nolint:forcetypeassert
-		data, err := enc.MarshalText()
-		if err != nil {
-			return err
-		}
-		data, err = json.Marshal(string(data))
+	case t == reflect.TypeFor[json.RawMessage]():
```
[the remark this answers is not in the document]
deniseli: We use it in ingress: https://github.com/TBD54566975/ftl/blob/main/backend/controller/ingress/response.go#L19

Without this, the `omitempty` behavior doesn't happen, and [this Ingress test](https://github.com/TBD54566975/ftl/blob/main/backend/controller/ingress/handler_test.go#L24C6-L24C17) fails because `"error": ""` gets included in the response body, which also breaks the [status code](https://github.com/TBD54566975/ftl/blob/main/backend/controller/ingress/handler_test.go#L107).

That's a really good point... should I update encoding to handle the `omitempty` next, so that this special case for `json.RawMessage` can be removed?

go-runtime/encoding/encoding.go
```diff
@@ -107,6 +95,24 @@ func encodeValue(v reflect.Value, w *bytes.Buffer) error {
 	}
 }
 
+var ftlOptionTypePath = "github.com/TBD54566975/ftl/go-runtime/ftl.Option"
+
+func isOption(t reflect.Type) bool {
+	return strings.HasPrefix(t.PkgPath()+"."+t.Name(), ftlOptionTypePath)
+}
+
+func encodeOption(v reflect.Value, w *bytes.Buffer) error {
```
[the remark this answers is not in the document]
deniseli: Thanks for catching that... we originally did this to break a circular dependency between the `encoding` and `ftl` packages, but since `call`'s dependency prevents that anyway, this is now pointless. cc @worstell so we can discuss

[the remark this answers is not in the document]
deniseli: One option is to move the logic back to `option.go`, undo the broken encapsulation, and then special case `Option` explicitly in the `switch` in `encodeValue`.

go-runtime/ftl/option.go
deniseli: I added a proposal for fixing this here: https://github.com/TBD54566975/ftl/pull/1417#discussion_r1591761950

What do you think? Pending agreement, I'll fix forward using that strategy.

go-runtime/encoding/encoding.go
deniseli: It is very very gross. We spent a bit of time working on wrapping `json.Decoder`, but the issue is that we can't just wrap `Token()` (returning the last `Peek` value when defined instead of what's at the top of the buffer) and add `Peek` because `d.Decode` will pop tokens off the underlying buffer, and if we've popped a token into the wrapper object, the underlying Decoder's `Decode()` function won't use that. So unless there's an alternative I haven't thought of, we'd effectively have to reimplement the whole `json.Decoder`. Do you have any ideas?

go-runtime/encoding/encoding.go
```diff
@@ -31,37 +27,29 @@ func encodeValue(v reflect.Value, w *bytes.Buffer) error {
...
-		v = v.Elem()
-		fallthrough
 
-	case t.Implements(jsonMarshaler):
-		enc := v.Interface().(json.Marshaler) //nolint:forcetypeassert
-		data, err := enc.MarshalJSON()
+	// Special-cased types
+	switch {
+	case t == reflect.TypeFor[time.Time]():
+		data, err := json.Marshal(v.Interface().(time.Time))
```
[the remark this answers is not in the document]
deniseli: Whoops, thanks for catching that! Fixed in https://github.com/TBD54566975/ftl/pull/1424

go-runtime/encoding/encoding.go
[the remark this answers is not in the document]
alecthomas: That sounds like a reasonable approach.

go-runtime/encoding/encoding.go
alecthomas: Ah right, that makes sense.

I think we want to ditch `json.Decoder` altogether, but that's not appropriate for this PR.

go-runtime/encoding/encoding.go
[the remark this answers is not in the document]
worstell: ah actually we needed to move the implementation into encoding.go to prepare for a following change that will pass context through to the custom encoder. the json marshaling overrides for option must call into the custom encoder to marshal/unmarshal the optional’s underlying value, and we don’t have a context to pass from these callers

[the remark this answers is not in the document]
worstell: a potential solution with callbacks—adding an interface like this in `encoding.go`:
```
type OptionMarshaler interface {
	Marshal(marshalSome func ... ) ([]byte, error)
	Unmarshal(unmarshalSome func ... ) error
}
```
implemented in `optional.go` (this just subs the `ftlencoding` calls from our current JSON overrides in `option.go` with the callbacks), e.g.:
```
func (o Option[T]) Marshal(marshalSome func(v any) ([]byte, error)) ([]byte, error) {
	if o.ok {
		return marshalSome(o.value) // was `ftlencoding.Marshal(o.value)`
	}
	return []byte("null"), nil
}

func (o Option[T]) Unmarshal(data []byte, unmarshalSome func(data []byte, v any) error) error {
	if string(data) == "null" {
		o.ok = false
		return nil
	}
	if err := unmarshalSome(data, &o.value); err != nil { // was `ftlencoding.Unmarshal(data, &o.value)`
		return err
	}
	o.ok = true
	return nil
}
```

then `encoding.go` uses `OptionMarshaler.Marshal/Unmarshal` if the value implements `OptionMarshaler`. the callbacks it would pass would be the original `Marshal/Unmarshal` calls

Issue #1390, opened 2024-05-02
make sure `ftl.Map` handles changes in underlying values correctly


matt2e:
From slack:

>aat
> Something else re. testing 
> [@mtoohey](https://sq-tbd.slack.com/team/U06LEKDJX9P)
>  - ftl.Map will need to change such that its caching is contingent on the underlying value being the same, rather than blindly doing it

```
ctx := ftltest.Context(
    ftltest.WithConfig(example, "hello"),
)

//Call a verb that uses ftl.Map with example config
resp, err := Verb(ctx, req)

ctx = ftltest.Context(
    ftltest.WithConfig(example, "world"),
)

//When this execution of the verb is done, it should execute again because the underlying value is different (hello -> world)
resp, err = Verb(ctx, req)
```

PR #1434, opened 2024-05-08
fix: possible solution for ftl.Map with changing values

matt2e:
fixes https://github.com/TBD54566975/ftl/issues/1390

Currently this PR is half implemented to just get a sense of this is the right direction.

Changes:
- `ftl.HashableHandle` extends `ftl.Handle` with `Hash(ctx context.Context) []byte`
- When Get(ctx) is called on a map handle, it checks if the old hash matches the current hash. If they differ, it gets a new value.
    - If the handle does not support hashing, then it falls back to the original behaviour of only ever mapping once.
- See echo_test.go for examples with config and databases. I haven't implemented for other types just yet

Questions:
- is it only ftl types that we expect to work with `ftl.Map`? If not, it is more work for them to make it work well in tests or if they don't implement `HashableHandle` then they may see off behaviour in tests
- Is `HashableHandle` needed or should we make this part of Handle to make it more obvious that you have to implement `Hash(ctx)` to make it work...

**Other proposed solutions:**
> aat: I think ftl.Map would need to change such that it hashes ModuleContext somehow

I think that would break the once aim of map in this case:
```
ctx := ftl.Context(...)
x = mapHandle.Get(ctx) // map func called correctly

// modify the context somehow in an unrelated way
ctx = context.WithValue(ctx, aKey, aValue)

y = mapHandle.Get(ctx) // map func called but the underlying value hasn't changed
```

### DIFF
--- a/examples/go/echo/echo.go
+++ b/examples/go/echo/echo.go
@@ -3,7 +3,9 @@ package echo
 
 import (
 	"context"
+	"database/sql"
 	"fmt"
+	"unsafe"
 
 	"ftl/time"
 
@@ -11,6 +13,13 @@ import (
 )
 
 var defaultName = ftl.Config[string]("default")
+var defaultMap = ftl.Map(defaultName, func(ctx context.Context, configStr string) (string, error) {
+	return configStr + " mapped", nil
+})
+var db = ftl.PostgresDatabase("echo")
+var dbMap = ftl.Map(db, func(ctx context.Context, db *sql.DB) (uintptr, error) {
+	return uintptr(unsafe.Pointer(db)), nil
+})
 
 // An echo request.
 type EchoRequest struct {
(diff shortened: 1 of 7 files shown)


matt2e:
Hmmm another issue with the current logic:
```
        t.Setenv("FTL_POSTGRES_DSN_ECHO_ECHO", "fake")
	ctx1 := ftltest.Context()
	ctx2 := ftltest.Context()
	
	firstMapValue := dbMap.Get(ctx1)
	secondMapValue := dbMap.Get(ctx2)

	assert.Equal(t, firstMapValue, dbMap.Get(ctx1))
	assert.Equal(t, secondMapValue, dbMap.Get(ctx2))
```
MapHandle is only tracking the latest value...
This could happen when doing a series of test t.Run() in a test asynchronously I think...?

matt2e:
Closing this, after chatting with @alecthomas, the decision is to lose the once/singleton nature of ftl.Map as that is complicating things and was only meant as an optimisation not the purpose of ftl.Map

PR #1440, opened 2024-05-08
feat: support sum types on ingress

worstell:


integration/integration_test.go
```diff
@@ -212,6 +212,14 @@ func TestHttpIngress(t *testing.T) {
 			assert.Equal(t, jsonData(t, []obj{{"item": "a"}, {"item": "b"}}), resp.bodyBytes)
 			return nil
 		}),
+		// todo: figure out how to pass type registry into context for this test
```
worstell, line 215: this is tricky because defining context with the type registry in test setup would require reflecting over types (e.g. ftl/httpingress.TypeEnum) that are not available to the test file (these are only in the `httpingress.go` file later scaffolded in this test)

ideally we could just use the context from the module we build, but not sure best way to do this— was thinking maybe @matt2e’s recent changes might be relevant?

matt2e: In other integration tests, I've moved some of the logic into a module's tests, and then kicked it off from integration test. That way you get the built out version of the module.
Would that solve your problem?
- module test: https://github.com/TBD54566975/ftl/blob/main/integration/testdata/go/wrapped/wrapped_test.go#L16
- integration test kicking off the module tests: https://github.com/TBD54566975/ftl/blob/main/integration/integration_test.go#L231

worstell: nice ty!!

i actually realized the initial approach was fundamentally wrong because the request context on ingress won't ever have the type registry in it

then i realized i didn't need a type registry at all for ingress validation, it's only necessary in the `go-runtime` encoder. ingress validation is already looking at the schema so it doesn't need to fetch anything else. so i've gutted that whole parallel `schema.TypeRegistry` in this PR

PR #1441, opened 2024-05-08
feat: support omitempty in custom json de/encoder

deniseli:
Fixes https://github.com/TBD54566975/ftl/issues/1262

### DIFF
--- a/go-runtime/encoding/encoding_test.go
+++ b/go-runtime/encoding/encoding_test.go
@@ -32,6 +32,11 @@ func TestMarshal(t *testing.T) {
 	type inner struct {
 		FooBar string
 	}
+	type validateOmitempty struct {
+		ShouldOmit   string `json:",omitempty"`
+		ShouldntOmit string `json:""`
+		NotTagged    string
+	}
 	tests := []struct {
 		name     string
 		input    any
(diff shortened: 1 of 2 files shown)


go-runtime/encoding/encoding_test.go
```diff
@@ -58,6 +63,8 @@ func TestMarshal(t *testing.T) {
 		{name: "Pointer", input: &struct{ String string }{"foo"}, err: `pointer types are not supported: *struct { String string }`},
 		{name: "SumType", input: struct{ D discriminator }{variant{"hello"}}, expected: `{"d":{"name":"Variant","value":{"message":"hello"}}}`},
 		{name: "UnregisteredSumType", input: struct{ D unregistered }{variant{"hello"}}, err: `the only supported interface types are enums or any, not encoding_test.unregistered`},
+		{name: "OmitEmptyNotNull", input: validateOmitempty{"foo", "bar", "baz"}, expected: `{"shouldOmit":"foo","shouldntOmit":"bar","notTagged":"baz"}`},
+		{name: "OmitEmptyNull", input: validateOmitempty{}, expected: `{"shouldntOmit":"","notTagged":""}`},
```
wesbillman, line 67: Nice! I wonder if we should add some `omitempty` to test data `.go` files in a test as well to make sure it's working end-to-end?

deniseli: Great idea! I'll do that now

Issue #1439, opened 2024-05-08
Remove usage of json.RawMessage via `go-runtime/encoding` package


deniseli:
`json.RawMessage` is not a type that FTL supports. The only standard lib type we currently support is `time.Time`.

However, currently, ingress requests and responses have `json.RawMessage` fields in their structs. They are configured using the `json` struct tag to be omitted from the output JSON when undefined. 

Now that we have support for omitempty (https://github.com/TBD54566975/ftl/issues/1262), we no longer need to type these fields as `json.RawMessage`. They are encoded as `[]byte` underlying, but byte arrays are serialized to base64 strings by our custom encoder, rather than a human readable value.

I haven't explored this fully to prove this works, but a reasonable option could be to convert the `json.RawMessage` fields to `string`, which would be human-readable.

Make sure to remove the special case for `json.RawMessage` in go-runtime/encoding/encoding.go as a part of this: [link](https://github.com/TBD54566975/ftl/blob/23b18faca8bbb0d46a5eb32d16f401172d37239d/go-runtime/encoding/encoding.go#L63)

PR #1454, opened 2024-05-09
fix: replace json.RawMessage with underlying []byte type in ingress response

deniseli:
Fixes https://github.com/TBD54566975/ftl/issues/1439

Alec was right - the `[]byte` array was the right solution here! The response does come back encoded the same way it was when it was typed `json.RawMessage`. I confirmed in `backend/controller/ingress/handler_test.go` (which would fail if we made the `encoding.go` change without the `response.go` change) and also cleaned up the test a bit.


alecthomas (approved):

deniseli:
Tragedy strikes! This is breaking the ingress integration tests - all the status codes are getting set to 500. Reverting for now

PR #1464, opened 2024-05-11
fix: split request and response area evenly

wesbillman:
Screenshot 2024-05-10 at 5 17 58 PM
Screenshot 2024-05-10 at 5 18 01 PM


deniseli:
Nice!!

Issue #1475, opened 2024-05-14
Declaring an enum after a struct can make the struct lose the field type


matt2e:
Module code:
```go
//ftl:enum
type Season string

const (
	Summer Season = "summer"
	Autumn Season = "autumn"
	Winter Season = "winter"
	Spring Season = "spring"
)

//ftl:data
type Info struct {
	Season Season
	Hair   HairColor
}

//ftl:enum
type HairColor string

const (
	Blonde HairColor = "blonde"
	Black  HairColor = "black"
	Brown  HairColor = "brown"
	White  HairColor = "white"
	Other  HairColor = "other"
)
```
Expected schema:
```
...
data Info {
  season order.Season
  hair order.HairColor
}
```

Actual schema:
```
...
data Info {
  season order.Season
  hair string // <--- !!!
}
```

matt2e:
(This is in the same area as type aliases)

Issue #1476, opened 2024-05-14
Declaring an enum const before the enum skips the value


matt2e:
Module code
```go
const (
	Onboarded UserState = "onboarded"
)

//ftl:enum
type UserState string

const (
	Registered UserState = "registered"
	Active     UserState = "active"
	Inactive   UserState = "inactive"
)
```
Expected vs actual schema:
```
enum UserState: String {
  // 🛑 Not included: Onboarded = "onboarded"
  Registered = "registered"
  Active = "active"
  Inactive = "inactive"
}
```

PR #1493, opened 2024-05-15
feat: add type aliases, refactor go schema parsing

matt2e:
https://github.com/TBD54566975/ftl/issues/1238
Adds the ability to declare type aliases in modules
```
//ftl:typealias
type UserId string
```

There is a lot of overlap between typealiases and enums. They both redefine an existing type as a new type, and therefore need to make sure our schema parsing does not skip this defined type by looking at the underlying type.
There were also issues if implicit exports, enum cases (and more?)  were encountered in the ast tree before we found the typealias or enum definition.
This PR solves that by doing an initial pass just to find all typealias and enum declarations for the module and then doing the normal pass.

Previously typealiases were allowed without any declaration but they would just fall back to the underlying type.
This PR makes this an error as we do not know if this type should be an enum or a type alias. We may want to discuss this more.

fixes https://github.com/TBD54566975/ftl/issues/1475 https://github.com/TBD54566975/ftl/issues/1476 https://github.com/TBD54566975/ftl/issues/1477 https://github.com/TBD54566975/ftl/issues/1492 https://github.com/TBD54566975/ftl/issues/1449


matt2e:
making this a draft because its failing tests and I'm done for the night

matt2e:
Actually tests are succeeding, now just lint failing (sumtypes). Will submit and fix that tomorroq

backend/schema/validate.go
```diff
@@ -279,10 +279,11 @@ func ValidateModule(module *Module) error {
 		}
 		switch n := n.(type) {
 		case *Ref:
-			if scopes.Resolve(*n) == nil && n.Module == "" {
+			mdecl := scopes.Resolve(*n)
+			if mdecl == nil && n.Module == "" {
 				merr = append(merr, errorf(n, "unknown reference %q", n))
 			}
-			if mdecl := scopes.Resolve(*n); mdecl != nil {
+			if mdecl != nil {
```
wesbillman, line 286: oh nice! :)

go-runtime/compile/schema_test.go
```diff
@@ -367,10 +434,13 @@ func TestErrorReporting(t *testing.T) {
...
-		`100:2-23: cannot attach enum value to BadValueEnum because it is a variant of type enum TypeEnum, not a value enum`,
-		`106:2-40: cannot attach enum value to BadValueEnumOrderDoesntMatter because it is a variant of type enum TypeEnum, not a value enum`,
-		`115:1-26: parent enum "ExportedTypeEnum" is exported, but directive "ftl:data" on "PrivateData" is not: all variants of exported enums that have a directive must be explicitly exported as well`,
-		`119:21-60: config and secret names must be valid identifiers`,
+		`101:2-24: cannot attach enum value to BadValueEnum because it is a variant of type enum TypeEnum, not a value enum`,
+		`108:2-41: cannot attach enum value to BadValueEnumOrderDoesntMatter because it is a variant of type enum TypeEnum, not a value enum`,
+		`117:1-26: parent enum "ExportedTypeEnum" is exported, but directive "ftl:data" on "PrivateData" is not: all variants of exported enums that have a directive must be explicitly exported as well`,
+		`121:21-60: config and secret names must be valid identifiers`,
+		`127:1-26: only one directive expected for type alias`,
+		`133:2-2: type is not declared as an ftl type alias or enum: NonFTLAlias`,
```
wesbillman, line 442: I was a bit confused by this error message at first. I wonder if shuffling it around would make it more clear 🤔 

```
type "NonFTLAlias" is not declared as an ftl type alias or enum 
```

go-runtime/compile/testdata/failing/failing.go
```diff
@@ -117,3 +119,16 @@ type PrivateData struct{}
 func (PrivateData) exportedTypeEnum() {}
 
 var invalidConfig = ftl.Config[string]("not an identifier")
+
+// There can be only one
```
wesbillman, line 123: Gif

wesbillman (approved):
Looks good! Love all the tests :)

alecthomas:
Added @worstell as she's been touching this code a lot recently!

go-runtime/compile/schema.go
```diff
@@ -495,16 +593,15 @@ func visitGenDecl(pctx *parseContext, node *ast.GenDecl) {
 		if err != nil {
 			pctx.errors.add(err)
 		}
-		if maybeVisitTypeEnumVariant(pctx, node, directives) {
-			return
-		}
+		maybeVisitTypeEnumVariant(pctx, node, directives)
```
alecthomas, line 596: Is this intentional?

go-runtime/compile/schema.go
```diff
@@ -1278,6 +1433,95 @@ func (p *parseContext) pathEnclosingInterval(start, end token.Pos) (pkg *package
...
 }
 
+func (p *parseContext) isPathInPkg(path string) bool {
+	if path == p.pkg.PkgPath {
+		return true
+	}
+	return strings.HasPrefix(path, p.pkg.PkgPath+"/")
+}
+
+func (p *parseContext) getEnumForTypeName(name string) (*schema.Enum, optional.Option[*types.Interface], bool) {
```
alecthomas, line 1443: Can you document this? I have no clue what the `bool` here does. If it's indicating whether the enum is being returned, just use an optional for that.

alecthomas: Also, a valid alternative to comments in some circumstances is to name the return values. eg. if this was `enumPresent bool`

go-runtime/compile/schema_test.go
alecthomas: Agreed, but also in general I tend to leave out the typename iff the error is pointing directly at it already, as it's redundant information (not sure if that is the case here).

alecthomas (approved):
Huge!

Who knew typealias support would be such a PITA?!

go-runtime/compile/schema_test.go
matt2e: Switched it to:
> type is not declared as an ftl enum or type alias

go-runtime/compile/schema.go
matt2e: This allowed this case: https://github.com/TBD54566975/ftl/issues/1492
Updating it now to collect errors for subsequent type enums it conforms to

matt2e: Actually no. This is there because even though something is a variant we still need to continue to process it's declaration. eg: a type alias that conforms to am enum.

PR #1521, opened 2024-05-17
fix: fix crash extracting struct with type parameter

matt2e:
FTL panics when compiling code like this:
```
//ftl:data
type DataWithType[T any] struct {
	Value T
}
```

`VisitStruct` was iterating through type parameters and type parameter values, but there are no values in this case.


matt2e:
Actually, putting this on hold, I think https://github.com/TBD54566975/ftl/pull/1522 might have unblocked what I needed.

matt2e:
It wasn't blocking me in the end, but submitting again because I do think its a legit issue.

PR #1534, opened 2024-05-17
fix: GetModuleContext: use config from context instead of from envar

### LINKED ISSUE #1352: `ftl-project.toml` loaded even if a different project file is specified
If we have 2 `ftl-project.toml` files (notice `op://` used in first file). And we specify with `ftl dev ...` that we want to use `ftl-project-inline.toml` it seems like the code is still loading the base `ftl-project.toml` file even though we didn't specify it. This leads to issues where 1Password needs to be configured even though we're online using `inline` secrets.

```sh
ftl dev examples/go --recreate --config=ftl-project.toml,ftl-project-inline.toml

info:controller0: Web console available at: http://localhost:8892
info: ftl-project-inline.toml startup command ⚡️
info:time: Building module
info:echo: Building module
info:time: Deploying module
info:echo: Deploying module
error:echo: main: error: could not get config: internal: could not marshal module context: failed to get config map: no provider for scheme "op"
warn:echo: Plugin failed to start, terminating pid 69894
```

```toml
# ftl-project.toml
module-dirs = ["examples/go"]
external-dirs = []

[global]

[modules]
[modules.echo]
[modules.echo.configuration]
default = "op://ImFub255bW91cyI"

[commands]
startup = ["echo 'ftl-project.toml startup command ⚡️'"]
```

```toml
# ftl-project-inline.toml
module-dirs = ["examples/go"]
external-dirs = []

[global]

[modules]
[modules.echo]
[modules.echo.configuration]
default = "inline://ImFub255bW91cyI"

[commands]
startup = ["echo 'ftl-project-inline.toml startup command ⚡️'"]
```

safeer:
Fixes #1352

With this change, the behavior is:
- `ftl dev examples/go`
  - uses `ftl-project.toml` in the gitroot if it exists
- `FTL_CONFIG=ftl-project-inline-a.toml ftl dev examples/go`
  - uses `ftl-project-inline-a.toml`
- `ftl dev examples/go --config=ftl-project-inline-b.toml`
  - uses `ftl-project-inline-b.toml`
- `FTL_CONFIG=ftl-project-inline-a.toml ftl dev examples/go --recreate --config=ftl-project-inline-b.toml`
  - uses `ftl-project-inline-b.toml`

### DIFF
--- a/backend/controller/controller.go
+++ b/backend/controller/controller.go
@@ -657,17 +657,12 @@ nextModule:
 func (s *Service) GetModuleContext(ctx context.Context, req *connect.Request[ftlv1.ModuleContextRequest]) (*connect.Response[ftlv1.ModuleContextResponse], error) {
 	name := req.Msg.Module
 
-	cm, err := cf.NewDefaultConfigurationManagerFromConfig(ctx, cf.ConfigFromEnvironment())
-	if err != nil {
-		return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("could not get configs: %w", err))
-	}
+	cm := cf.ConfigFromContext(ctx)
+	sm := cf.SecretsFromContext(ctx)
+
 	configs, err := cm.MapForModule(ctx, name)
 	if err != nil {
 		return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("could not get configs: %w", err))
-	}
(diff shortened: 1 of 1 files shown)


safeer:
Thanks @deniseli! I had a conversation with @wesbillman and @worstell, and we all agreed that having the CLI path override the envar makes the most sense, since it's more explicit. I'll hold off until we can talk about it during the sync :)

safeer:
Looks like the override actually happens separately in Kong :)

Issue #1496, opened 2024-05-15
Allow value enums to be variants of type enums


worstell:
```
//ftl:enum
type TypeEnum interface {
  typeEnum()
}

type ValueEnumType ValueEnum
func (ValueEnumType) typeEnum() {}


//ftl:enum
type ValueEnum string

const (
  A ValueEnum = "a"
)
```
should be extracted as:
```
enum ValueEnum: String {
   A = "a"
} 

enum TypeEnum {
   ValueEnumType foo.ValueEnum
}
```
right now, it extracts as the following:
```
enum ValueEnum: String {
   A = "a"
} 

enum TypeEnum {
   ValueEnumType string
}
```

however, this pattern will remain erroneous:
```
//ftl:enum
type TypeEnum interface {
  typeEnum()
}

//ftl:enum
type ValueEnum string
func (ValueEnum) typeEnum() {}

const (
  A ValueEnum = "a"
)
```
because `ValueEnum` cannot be both a type variant name and itself a value enum. we do not support mixing of type and value variants in a single enum

PR #1535, opened 2024-05-18
fix: support inlined or aliased structs as type enum variants

### LINKED ISSUE #1531: Support aliased and inlined struct type enum variants
both of the below will be supported ways of declaring struct type enum variants:
```
type TypeEnum interface {
   typeenum()
}

type Inlined struct {
}
func (Inlined) typeenum()

type Aliased Underlying {
}
func (Aliased) typeenum()

type Underlying struct {
}
```

producing the following schema:
```
enum TypeEnum {
   Inlined foo.Inlined
   Aliased foo.Underlying
}
```

if the struct is inlined, we will use its name as both the variant name and the type name. right now, we're only supporting the inlined case correctly

worstell:
small changes + fixes preempting the refactor

fixes #1531, #1527, #1496


alecthomas (approved):
Nice!

PR #1558, opened 2024-05-22
fix: allow subpackage types

matt2e:
Reverts two PRs:
- https://github.com/TBD54566975/ftl/pull/1530
- https://github.com/TBD54566975/ftl/pull/1535

This is a blunt fix to unblock based on our priorities. We should revisit https://github.com/TBD54566975/ftl/pull/1535 again later


deniseli (approved):

alecthomas (approved):
Ah you think this is the best approach vs. being surgical?

PR #1564, opened 2024-05-23
fix: treat type aliased subpackage types as underlying types

matt2e:
This issue was found when trying to upgrade ftl with another codebase:
- module has an ftl type that refers to `subpackage.CustomerType`
- subpackage has `type CustomerType string`

Previously this would get added to the schema as just `String`, but our recent changes had made it get added to the schema as `module.CustomerType`

We now detect this case and use the underlying type like we used to.


alecthomas (approved):

go-runtime/compile/testdata/parent/parent.go
```diff
@@ -0,0 +1,12 @@
+package parent
+
+import (
+	"context"
+	// Import the FTL SDK.
```
deniseli, line 5: nit: irrelevant comment

PR #1570, opened 2024-05-25
feat: add database-backed config resolver + provider for production

deniseli:
Fixes https://github.com/TBD54566975/ftl/issues/1548

This doesn't hook anything up yet - just adds the provider and resolver that we'll use in prod

Changes:
* Adds `DBConfig[Resolver|Provider]` to the `configuration` package, where the resolver does very little, and the provider executes most of the actual database SQL queries. For DB-backed configs, all the data needed to retrieve a config can be found in the `ref` without a separate key, so the URLs are all simple `db://` stubs that exist only to satisfy the existing provider and resolver interfaces, which expect URLs to be passed as keys.
* Refactors the 3 Err types, `isNotFound`, and `translatePGError` out of the `dal` package into a separate `dalerrors`
* Adds a public getter for the `db` field in `dal.DAL` so that we can connect to the same DB in the database config resolver/provider


common/configuration/projectconfig_resolver_test.go
```diff
@@ -40,6 +40,37 @@ func TestSet(t *testing.T) {
 	})
 }
 
+func TestGetGlobal(t *testing.T) {
```
deniseli, line 43: This test is not actually related to my change. I just added it to give myself some extra confidence that I was building the right thing (i.e. the fallback-to-global behavior) and then decided to keep this in for some extra branch coverage.

common/configuration/db_provider.go
```diff
@@ -0,0 +1,30 @@
...
+
+var _ MutableProvider[Configuration] = DBProvider{}
+
+func (DBProvider) Role() Configuration { return Configuration{} }
+func (DBProvider) Key() string         { return "db" }
+
+func (d DBProvider) Writer() bool { return d.DB }
+
+func (DBProvider) Load(ctx context.Context, ref Ref, key *url.URL) ([]byte, error) {
+	return []byte(key.Host), nil
```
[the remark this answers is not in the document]
alecthomas: I'm a bit on the fence about whether it's a good idea to have the content encoded into the URL. Config values are arbitrary JSON, so they could be quite large. I think it might be preferable if the URL was just a reference, and the provider retrieves the value. It's an extra DB round trip, but I think that's okay.

backend/controller/dal/database_resolver.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+func (d DatabaseResolver) List(ctx context.Context) ([]configuration.Entry, error) {
+	configs, err := d.db.ListConfigs(ctx)
+	if err != nil {
+		return nil, err
+	}
+	entries := []configuration.Entry{}
+	for _, c := range configs {
+		u, err := url.Parse(c.Accessor)
```
alecthomas, line 55: The database definitely shouldn't need to store the raw URL.

backend/controller/dal/database_resolver.go
```diff
@@ -0,0 +1,81 @@
...
+func (d DatabaseResolver) Get(ctx context.Context, ref configuration.Ref) (*url.URL, error) {
+	module, moduleOk := ref.Module.Get()
+	if !moduleOk {
+		module = "" // global
+	}
+	accessor, err := d.db.GetConfig(ctx, module, ref.Name)
+	if err != nil {
+		// If we could not find this config within the module, then try getting
+		// again with scope set to global
+		if moduleOk {
```
alecthomas, line 37: I think we should push this down to the query so we can retrieve a global value or the module specific value in a single query, something like this:

```sql
SELECT value FROM module_config
WHERE
  (module IS NULL AND name = @name)
  OR (module = @module AND name = @name)
ORDER BY module NULLS LAST -- TIL about NULLS LAST, handy
LIMIT 1
```

alecthomas: Though as mentioned elsewhere we should not include the value in the URL, in which case the resolver probably doesn't even need to hit the database at all it can just return `db://[<module>.]<name>` - but this select would be used by the provider.

alecthomas (changes requested):
Good start, but let's not include the value in the URL, and the DB should store the raw value not a URL. The URL can be completely synthetic for this use case, only used as a reference.

backend/controller/dal/database_resolver.go
```diff
@@ -0,0 +1,81 @@
...
+		entries = append(entries, configuration.Entry{
+			Ref:      configuration.NewRef(c.Module, c.Name),
+			Accessor: u,
+		})
+	}
+	return entries, nil
+}
+
+func (d DatabaseResolver) Set(ctx context.Context, ref configuration.Ref, key *url.URL) error {
+	return d.db.SetConfig(ctx, moduleAsString(ref), ref.Name, key.String())
```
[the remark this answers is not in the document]
deniseli: ah, got it. Done

backend/controller/sql/schema/001_init.sql
```diff
@@ -445,6 +445,15 @@ CREATE TABLE fsm_transitions (
...
     async_calls_id BIGINT NOT NULL REFERENCES async_calls(id) ON DELETE CASCADE
-)
+);
+
+CREATE TABLE configs
+(
+    id           BIGINT NOT NULL GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+    module       TEXT   NOT NULL,
+    name         TEXT   NOT NULL,
+    accessor     TEXT   NOT NULL,
```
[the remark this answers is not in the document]
deniseli: Ahhh that makes sense. Accessor came from here: https://github.com/TBD54566975/ftl/blob/main/common/configuration/api.go#L32

I'll rename this to `value` and switch the type. Great idea - thanks!

backend/controller/sql/schema/001_init.sql
```diff
@@ -445,6 +445,15 @@ CREATE TABLE fsm_transitions (
     id BIGINT NOT NULL GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
     fsm_executions_id BIGINT NOT NULL REFERENCES fsm_executions(id) ON DELETE CASCADE,
     async_calls_id BIGINT NOT NULL REFERENCES async_calls(id) ON DELETE CASCADE
-)
+);
+
+CREATE TABLE configs
+(
+    id           BIGINT NOT NULL GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+    module       TEXT   NOT NULL,
```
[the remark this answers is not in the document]
deniseli: Ahhh good call - done

common/configuration/db_provider.go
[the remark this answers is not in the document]
deniseli: Yeah... what do you think about having the URL be "db://{{id}}" and then DBProvider.Load retrieves the config value by ID?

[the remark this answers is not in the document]
deniseli: Actually that wouldn't make sense in `DatabaseResolver.Set`, then, since neither the Ref nor the URL would encapsulate the actual config value we're storing... I'll go with the base64 encoding for now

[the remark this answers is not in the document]
alecthomas: That could work, but then you'll need two queries - one for the resolver to retrieve the ID and one for the provider to load the value. It's probably a bit more efficient to make it something like `db://[<module>.]<name>` - you actually won't need to parse it at all because the resolver passes the Ref down to the provider, it's just for users.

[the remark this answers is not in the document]
alecthomas: I don't think we should be base64 encoding anywhere.

backend/controller/dal/database_resolver.go
deniseli: It looks like there aren't any functions that take a `key *url.URL` and don't also get a `ref`, which already has the module and config name as structured data. In that case, can we just have the URL be effectively a meaningless stub and always operate off the ref?

deniseli: The biggest challenge seems to be the fact that `DatabaseResolver.Set` takes a URL as its only way to provide a value to `SetConfig` with. So we need to be able to retrieve the underlying value from the URL. I'll go with base64 encoding it for now...

deniseli: If we go with `db://[<module>.]<name>`, then what should `DatabaseResolver.Set` do? It only receives the ref and URL; no []byte value.

backend/controller/dal/database_resolver.go
deniseli: done!

common/configuration/db_provider.go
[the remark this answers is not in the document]
deniseli: I set this up to use b64 for now. My only concern with doing something like `db://[module.]name` is the question I mentioned below, around what `DatabaseResolver.Set` should do in that case.

common/configuration/db_provider.go
```diff
@@ -0,0 +1,30 @@
...
+}
+
+func (DBProvider) Store(ctx context.Context, ref Ref, value []byte) (*url.URL, error) {
+	return &url.URL{Scheme: "db", Host: string(value)}, nil
```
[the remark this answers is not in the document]
deniseli: base64 encoded here too

common/configuration/db_provider.go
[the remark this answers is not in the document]
deniseli: fwiw I am also very uncomfortable encoding the value into the URL haha

alecthomas:
I was thinking about this a bit more, and I think we should move this to the configuration package. The DAL methods obviously will have to stay in the `dal` package, but that's okay. To consume the DAL without having to import the entire thing, write an interface containing just the DAL methods the provider needs, and use that in the provider.

common/configuration/db_provider.go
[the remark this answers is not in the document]
deniseli: Done! URL is now just "db://" and all DB accesses except for `List` have been moved to the provider.

backend/controller/dal/database_resolver.go
deniseli: Done! Resolving this thread to consolidate discussion on the other one

deniseli:
@alecthomas sweet that makes sense! I'll split up the files so that constructors stay in the DAL package and the rest can move to `configuration`. It does feel quite gunky in the code as is.

backend/controller/dal/dal.go
```diff
@@ -281,14 +267,20 @@ func (t *Tx) Rollback(ctx context.Context) error {
 	return tx.Rollback(ctx)
 }
 
+// GetDB returns the private database interface, which is used by specific, designated
+// packages alongside the dalerrors package to talk to the same database.
+func (d *DAL) GetDB() sql.DBI {
+	return d.db
+}
+
```
wesbillman, line 275: We should remove this so that people can't bypass the protection the `DAL` provides. Instead we can use the `interface` approach in the comments below

common/configuration/db_config_provider.go
```diff
@@ -0,0 +1,50 @@
...
+
+// DBConfigProvider is a configuration provider that stores configuration in its key.
+type DBConfigProvider struct {
+	DB    bool `help:"Write configuration values to the database." group:"Provider:" xor:"configwriter"`
+	dalDB sql.DBI
+}
+
+var _ MutableProvider[Configuration] = DBConfigProvider{}
+
+func NewDBConfigProvider(db sql.DBI) DBConfigProvider {
```
wesbillman, line 19: instead of using `db sql.DBI` here and exposing direct access to the `db` we should probably just use the `interface` as @alecthomas suggested. If I'm reading his comment correctly, that might end up looking similar to the approach here in [cronjobs.go](https://github.com/TBD54566975/ftl/blob/main/backend/controller/cronjobs/cronjobs.go#L68-L73).

Then, your new stuff can use your new `DAL` which would only expose the methods needed by your new provider. It will likely make things a bit easier to test as well. 👍

common/configuration/db_config_resolver.go
```diff
@@ -0,0 +1,61 @@
...
+	"net/url"
+
+	"github.com/TBD54566975/ftl/backend/controller/dalerrors"
+	"github.com/TBD54566975/ftl/backend/controller/sql"
+	"github.com/TBD54566975/ftl/internal/slices"
+)
+
+// DBConfigResolver loads values a project's configuration from the given database.
+type DBConfigResolver struct {
+	db sql.DBI
```
wesbillman, line 14: This should use your new `DAL` interface as well. which will remove the need to do things like `TranslatePGError` hopefully.

backend/controller/controller.go
```diff
@@ -32,6 +32,7 @@ import (
 
 	"github.com/TBD54566975/ftl/backend/controller/cronjobs"
 	"github.com/TBD54566975/ftl/backend/controller/dal"
+	"github.com/TBD54566975/ftl/backend/controller/dalerrors"
```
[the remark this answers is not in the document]
deniseli: undid this whole change since we talked about the interface :D

backend/controller/dal/dal.go
deniseli: done :)

common/configuration/db_config_provider.go
deniseli: done :)

common/configuration/db_config_resolver.go
deniseli: done!

common/configuration/db_config_provider.go
```diff
@@ -0,0 +1,56 @@
...
+	"context"
+	"net/url"
+
+	"github.com/TBD54566975/ftl/backend/controller/dal"
+	"github.com/alecthomas/types/optional"
+)
+
+// DBConfigProvider is a configuration provider that stores configuration in its key.
+type DBConfigProvider struct {
+	DB  bool `help:"Write configuration values to the database." group:"Provider:" xor:"configwriter"`
```
alecthomas, line 13: I think we won't need to integrate with the CLI, so you can probably omit this.

common/configuration/db_config_provider.go
```diff
@@ -0,0 +1,56 @@
...
+	err := d.dal.SetModuleConfiguration(ctx, ref.Module, ref.Name, value)
+	if err != nil {
+		return nil, dal.TranslatePGError(err)
+	}
+	return &url.URL{Scheme: "db"}, nil
+}
+
+func (d DBConfigProvider) Delete(ctx context.Context, ref Ref) error {
+	err := d.dal.UnsetModuleConfiguration(ctx, ref.Module, ref.Name)
+	return dal.TranslatePGError(err)
```
alecthomas, line 55: No need to translate them, the DAL already does it.

alecthomas (approved):
Nice!

common/configuration/db_config_provider.go
deniseli: awesome! undone :)

backend/controller/dal/async_calls.go
```diff
@@ -81,7 +81,7 @@ func (d *DAL) AcquireAsyncCall(ctx context.Context) (call *AsyncCall, err error)
 	ttl := time.Second * 5
 	row, err := tx.db.AcquireAsyncCall(ctx, ttl)
 	if err != nil {
-		err = translatePGError(err)
+		err = TranslatePGError(err)
```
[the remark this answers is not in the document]
deniseli: fixed!

common/configuration/db_config_provider.go
deniseli: awesome!

backend/controller/dal/dal_test.go
```diff
@@ -400,3 +401,93 @@ func assertEventsEqual(t *testing.T, expected, actual []Event) {
...
+		}
+
+		// Insert entries in a separate order from how they should be returned to
+		// test sorting logic in the SQL query
+		err := dal.db.SetModuleConfiguration(ctx, sortedList[1].Module, sortedList[1].Name, []byte(`""`))
+		assert.NoError(t, err)
+		err = dal.db.SetModuleConfiguration(ctx, sortedList[2].Module, sortedList[2].Name, []byte(`""`))
+		assert.NoError(t, err)
+		err = dal.db.SetModuleConfiguration(ctx, sortedList[0].Module, sortedList[0].Name, []byte(`""`))
+		assert.NoError(t, err)
```
wesbillman, line 484: It looks like we're still using the `dal.db` here instead of a method on the `dal` itself. If we go straight to the db, I think it will bypass the usefulness of the dal. You might want to duplicate the pattern here: https://github.com/TBD54566975/ftl/blob/main/backend/controller/cronjobs/cronjobs.go#L69 which calls a `dal` method here: https://github.com/TBD54566975/ftl/blob/main/backend/controller/dal/dal.go#L990

deniseli: Ah, thanks for catching that! I'll make that change now.

PR #1601, opened 2024-05-29
fix: query ModuleConfiguration via DAL

deniseli:
Fixes https://github.com/TBD54566975/ftl/pull/1570#discussion_r1618128693


wesbillman (approved):
Nice! Thanks :)

PR #1633, opened 2024-06-03
fix: guard for panic

michaelneale:
candidate for https://github.com/TBD54566975/ftl/issues/1404


wesbillman (approved):
Nice! Thanks @michaelneale!

PR #1647, opened 2024-06-04
chore: remove jest from frontend

wesbillman:
Since we have no real tests in `frontend` currently, and they are painful to maintain, removing them for now. We can add a proper setup once we're ready to invest in frontend tests.

Also, removed a bunch of old unused packages


mistermoe:
this is the best thing that's happened to FTL console

Issue #1518, opened 2024-05-16
refactor go-runtime schema extraction


worstell:
this file has gotten out of hand!!!!
it's quite difficult to reason through

refactor will split out schema concepts into separate scans (e.g. parse type enums, parse value enums, parse verbs, etc.)

reference https://pkg.go.dev/golang.org/x/tools/go/analysis

PR #1700, opened 2024-06-07
chore: refactor go schema extraction

worstell:
initial step towards #1518

- integrates with ([forked](https://github.com/TBD54566975/golang-tools)) go/analysis to run static analysis tasks for schema extraction
- implements first extraction task, type aliases
- refactors existing build stack to use the new extractor, combining results with legacy so we can incrementally migrate
- removes typealias extraction logic from legacy code


go-runtime/compile/parser.go
```diff
@@ -1,5 +1,8 @@
 package compile
 
+// TODO: This file is now duplicated in go-runtime/schema/analyzers/parser.go. It should be removed
+// from here once the schema extraction refactoring is complete.
```
worstell, line 4: just duplicated this file to avoid having to export everything from it's new location (so the legacy extractor can continue using it) only to make it all private again once the refactor is complete

go-runtime/compile/schema.go
```diff
@@ -106,44 +107,34 @@ type ParseResult struct {
 	Errors []*schema.Error
 }
 
-// ExtractModuleSchema statically parses Go FTL module source into a schema.Module.
-func ExtractModuleSchema(dir string, sch *schema.Schema) (optional.Option[ParseResult], error) {
+func legacyExtractModuleSchema(dir string, sch *schema.Schema, out *analyzers.ExtractResult) error {
```
worstell, line 110: got a bit hairy attempting to run the legacy extractor as a completely isolated process since things are so entangled

instead modified the legacy code to accept an output pointer as param (`out *analyzers.ExtractResult`). the initial caller provides the partial result from the prior extraction step, which gives the legacy extractor access to details (native names and decls) that it needs. the output/result is updated in-place during the legacy extraction

worstell:
so many `go.mod` changes 😭  separated them into a separate commit for easier reviewing

alecthomas (approved):
This is amazing Lizzy. This will be _so_ much cleaner. 

A couple of thoughts/questions/comments:

This is not an immediate concern, but if we could switch to using the analysis package's [Diagnostic](https://pkg.go.dev/golang.org/x/tools/go/analysis) system, we'd get cleaner integration with the package.

Typically with the analysis package each pass will produce a distinct type that other analysers can require, and they will then be run in the correct order, and values passed through. It seems we could utilise this system rather than having a single `extractorContext` passed through and mutated, and I _think_ (though am not sure) that we would benefit from parallelisation then. For example, the `TypeAliasExtractor` could output a `[]*schema.TypeAlias`.

Another one for later, but I think we could also utilise the facts system. My first thought along those lines is that a first pass could add a `schema.Ref` fact to every node that is a reference to an FTL type or value. This would allow latter passes to just retrieve these facts directly from the node, rather than having to recompute them.

But yeah, overall, absolutely awesome.

alecthomas:
> so many `go.mod` changes 😭 separated them into their own commit for easier reviewing

Fear not, I just found this which made it actually quite okay :)
image

worstell:
> This is not an immediate concern, but if we could switch to using the analysis package's [Diagnostic](https://pkg.go.dev/golang.org/x/tools/go/analysis) system, we'd get cleaner integration with the package.
> 

ah in place of shoving the schema errors into the result? will check it out!

> Typically with the analysis package each pass will produce a distinct type that other analysers can require, and they will then be run in the correct order, and values passed through. It seems we could utilise this system rather than having a single `extractorContext` passed through and mutated, and I _think_ (though am not sure) that we would benefit from parallelisation then.
> 

currently the extractors actually do produce their own results (rather than mutate a single result), they just share the same type (`extractorContext`, maybe a confusing name)— so i believe we're already getting parallelization except for dependencies, which may end up being just finalizer -> everything else

> For example, the `TypeAliasExtractor` could output a `[]*schema.TypeAlias`.

i initially had the same thought, actually, to return []*schema.TypeAlias directly. i moved to the current approach to handle transitive visibility, e.g.:

pass A: `export typealias Alias foo.Thing`, `export data Thing{}`
pass B: `data Thing{}`
final result (merged): `export typealias Alias foo.Thing`, `export data Thing{}`

basically everyone producing a partial view of the world; then these get merged and the highest visibility Decl wins out in the case of conflicts. now in thinking about it more, this does feel cleaner as a separate calculation in finalizer that looks at the complete schema and uses refs to update visibility as needed. will take a look at updating tomorrow

> Another one for later, but I think we could also utilise the facts system. My first thought along those lines is that a first pass could add a `schema.Ref` fact to every node that is a reference to an FTL type or value. This would allow latter passes to just retrieve these facts directly from the node, rather than having to recompute them.
> 

amazing idea!!

go-runtime/compile/build.go
```diff
@@ -226,6 +224,36 @@ func Build(ctx context.Context, moduleDir string, sch *schema.Schema, filesTrans
 	return exec.Command(ctx, log.Debug, mainDir, "go", "build", "-o", "../../main", ".").RunBuffered(ctx)
 }
 
+// ExtractModuleSchema statically parses Go FTL module source into a schema.Module.
+func ExtractModuleSchema(dir string, sch *schema.Schema) (optional.Option[analyzers.ExtractResult], error) {
+	maybeResult, err := extract.Extract(dir)
+	if err != nil {
+		return optional.None[analyzers.ExtractResult](), err
+	}
+	result, ok := maybeResult.Get()
```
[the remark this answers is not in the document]
worstell: i think i got a bit optional-happy 😅

worstell:
@alecthomas this has been updated with the following:
- uses diagnostics instead of manually propagating schema errors
- extractors just return list of relevant Decls (also native names for now, but we might be able to get rid of this down the line by managing it as a separate pass)
- Decl visibility is updated transitively in finalizer — at the moment we're also running the same logic in `compile/build.go` after the legacy extractor runs, but we will remove this once we're migrated and just rely on the finalizer

PR #1739, opened 2024-06-11
fix: require a project toml for `ftl dev` and `ftl serve`

### LINKED ISSUE #1669: `ftl dev` and `ftl serve` should error if no `ftl-project.toml` file is found (or specified)
We want to require the usage of `ftl-project.toml` files now.

The user can override this with `--config=<custom>-project.toml`, but `ftl dev` and `ftl serve` should error if no project file is provided (or automatically detected)

safeer:
Fixes #1669 by maintaining the list of configuration files that generated a project's `Config`, and adding a non-empty pre-condition for `ftl dev` and `ftl serve`.

This seemed cleaner than plumbing through the ConfigFlag to sub-commands, but I could use another opinion..


safeer:
Going a different route

PR #1743, opened 2024-06-12
fix: pubsub schema fixes

matt2e:
includes fixes for the following cases:
- fix panic when validating if a call is valid if the call was to an inbuilt type (which has no package)
- fix issue finding a topic's name when declaring a subscription
    - the topic's var has a nil `.Obj` field, possibly because it was defined in a different file. Added a map from variable declaration position to the schema.Topic, and used that instead
- fix issue disambiguating between an illegal direct verb call (eg: `external.DoSomething(ctx)`, and an allowed field access (eg: `external.Data{}.Name`)

### DIFF
--- a/go-runtime/compile/schema_test.go
+++ b/go-runtime/compile/schema_test.go
@@ -286,9 +286,6 @@ func TestExtractModuleSchemaNamedTypes(t *testing.T) {
 		t.SkipNow()
 	}
 	assert.NoError(t, prebuildTestModule(t, "testdata/named", "testdata/namedext"))
-	if testing.Short() {
-		t.SkipNow()
-	}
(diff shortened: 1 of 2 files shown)


go-runtime/compile/schema_test.go
```diff
@@ -286,9 +286,6 @@ func TestExtractModuleSchemaNamedTypes(t *testing.T) {
 		t.SkipNow()
 	}
 	assert.NoError(t, prebuildTestModule(t, "testdata/named", "testdata/namedext"))
-	if testing.Short() {
```
matt2e, line 289: These were redundant as they are repeated a few lines earlier

PR #1769, opened 2024-06-13
fix: add `--insecure` to skip TLS certificate verification

### LINKED ISSUE #1738: Add an option to disable certificate checking on the ftl cli
Allow the ftl cli to make insecure SSL connections in the same manner that curl does via the --insecure flag

Configure the GRPC client to not do cert checks via configuring the HTTP client. Check the connect docs.

safeer:
Fixes #1738


safeer:
Screenshot 2024-06-13 at 2 01 43 PM
gosec does not like this change

alecthomas (approved):

PR #1771, opened 2024-06-13
fix: don't attempt to create ftl-project.toml in fs root

### LINKED ISSUE #1770: `ftl init` fails with `error: read-only file system`
Error when attempting to initialize a FTL Go module:
```
ftl init go . mymodule --log-level=DEBUG
debug: Unary RPC failed: unavailable: dial tcp [::1]:8892: connect: connection refused: /xyz.block.ftl.v1.AdminService/Ping
debug: Initializing FTL Go module mymodule in .
debug: Creating a new project config file at "/ftl-project.toml"
ftl: error: open /ftl-project.toml3523669785: read-only file system
```

safeer:
Fixes #1770

### DIFF
--- a/cmd/ftl/cmd_init.go
+++ b/cmd/ftl/cmd_init.go
@@ -64,7 +64,7 @@ func (i initGoCmd) Run(ctx context.Context, parent *initCmd) error {
 	}
 
 	if _, ok := internal.GitRoot(i.Dir).Get(); !ok {
-		return fmt.Errorf("directory %s is not a git repository", i.Dir)
+		return fmt.Errorf("directory %s is not in a git repository, run 'git init' at the root of your project", i.Dir)
 	}
 
 	logger := log.FromContext(ctx)
--- a/cmd/ftl/cmd_init.go
+++ b/cmd/ftl/cmd_init.go
@@ -63,6 +63,10 @@ func (i initGoCmd) Run(ctx context.Context, parent *initCmd) error {
 		return fmt.Errorf("module name %q is invalid", i.Name)
 	}
 
+	if _, ok := internal.GitRoot(i.Dir).Get(); !ok {
+		return fmt.Errorf("directory %s is not a git repository", i.Dir)
+	}
+
 	logger := log.FromContext(ctx)
 	logger.Debugf("Initializing FTL Go module %s in %s", i.Name, i.Dir)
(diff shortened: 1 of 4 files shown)


cmd/ftl/cmd_init.go
```diff
@@ -71,7 +71,7 @@ func (i initGoCmd) Run(ctx context.Context, parent *initCmd) error {
 	if err := updateGitIgnore(i.Dir); err != nil {
 		return err
 	}
-	if err := projectconfig.MaybeCreateDefault(ctx); err != nil {
+	if err := projectconfig.MaybeCreateDefault(ctx, i.Dir); err != nil {
```
wesbillman, line 74: I think we might want to create the project file in the gitRoot folder vs. where this command was run. I just tried this in the FTL repo and it added the project file to `examples/go` since that's where I ran the `ftl init....` command. But I would guess we would want the file to be created at the root of the ftl repo.

wesbillman (commented):
Nice! Thanks for taking a stab at this one!

cmd/ftl/cmd_init.go
safeer: I agree; think we want the search for `ftl-project.toml` to end at the gitroot instead of walking up the tree to the fs root, then it would create it there if it doesn't exist. Was planning on bringing this up during standup, but I'm also wondering about the general assumption that the project is in a git repo. My test project wasn't and FTL was very angry.

safeer: @wesbillman added an explicit check for a git repo during `init`

wesbillman: Nice. Thanks!

Maybe we should just list the assumptions and make sure the code does what we expect. So maybe the process is something like this in order (please feel free to make changes or add assumptions here)

1. Check for existing ftl-project.toml.
2. Create new ftl-project.toml at the gitRoot if gitRoot is found
3. Create new ftl-project at the location where the `ftl init` command was run

I don't suspect we would ever want to create this at the FS root.

wesbillman: Chatted with @safeer and he's going to put together an RFC for changing the CLI to have more obvious commands for creating ftl projects and modules.

wesbillman (approved):

PR #1773, opened 2024-06-13
refactor: deduplicate cron parsing

gak:


wesbillman (approved):
Nice! I was not aware of `~EOL)*` before :)

PR #1777, opened 2024-06-13
fix: WhenMap() now typesafe, ftl.Map returns ptr

matt2e:
- `ftltest.WhenMap(...)` included a func that returned `any` when it should be strongly types with the map's return type
- `ftl.Map` was returning `MapHandle` where as `handle.Get()` was defined on `*MapHandle` and `ftltest.WhenMap(...)` expected a pointer as well. I've made it all consistently use `*MapHandle` so you don't need to do `&handle` anywhere

### DIFF
--- a/go-runtime/ftl/ftltest/fake.go
+++ b/go-runtime/ftl/ftltest/fake.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/TBD54566975/ftl/backend/schema"
 	"github.com/TBD54566975/ftl/common/configuration"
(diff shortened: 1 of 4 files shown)


alecthomas (commented):
AFAICT this doesn't add any additional type safety? This just seems to move the tow safety boundary from WhenMap to one later deeper.

wesbillman (approved):

matt2e:
Not so much added safety checks, more of a QoL when writing tests. Previously returning the wrong type within `WhenMap()` would fail at runtime, but now it will fail at build time.

PR #1778, opened 2024-06-14
fix: ftl dev notifies when all deployments are up to date

### LINKED ISSUE #1706: Log `All modules deployed` after rebuilds
From @jiyoontbd 

Feature request: after a hot-reload, would it be terribly difficult to show the same message at the end `info: All modules deployed, watching for changes...` like what I see when I first boot up pfi by typing `just dev`? After a hot reload, I’m not always sure it’s done redeploying and it’s ok to proceed with more changes.

matt2e:
fixes #1706
when both of these things happen:
- we update a deployment successfully
- the engine idles for long enough for the file watcher to do 2 cycles
then we log "All modules deployed, watching for changes..."

### DIFF
--- a/buildengine/engine.go
+++ b/buildengine/engine.go
@@ -278,10 +278,18 @@ func (e *Engine) watchForModuleChanges(ctx context.Context, period time.Duration
 	})
 
 	// Watch for file and schema changes
+	didUpdateDeployments := false
 	for {
+		var completedUpdatesTimer <-chan time.Time
+		if didUpdateDeployments {
+			completedUpdatesTimer = time.After(period * 2)
+		}
 		select {
 		case <-ctx.Done():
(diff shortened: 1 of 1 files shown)


buildengine/engine.go
```diff
@@ -278,10 +278,18 @@ func (e *Engine) watchForModuleChanges(ctx context.Context, period time.Duration
 	})
 
 	// Watch for file and schema changes
+	didUpdateDeployments := false
 	for {
+		var completedUpdatesTimer <-chan time.Time
+		if didUpdateDeployments {
```
matt2e, line 284: this if statement just allows us to not aggressively poll when in an idle state, rather than putting `time.After` directly in the select statement

alecthomas (approved):

PR #1825, opened 2024-06-18
fix: move startup command exec from Dev to cmd_serve

### LINKED ISSUE #1343: We need a way to run [startup] command from ftl-project tomls when running tests
Pairing with @wesbillman. This came up. adding an issue here so we don't forget

This might be implemented short-term via `ftl serve` (which currently doesn't run the startup command)
Or longer-term perhaps we introduce a `ftl test` command that can be use when running tests.

cc: @matt2e this might be related to the testing journey you're about to embark on.

safeer:
Fixes #1343

### DIFF
--- a/cmd/ftl-controller/main.go
+++ b/cmd/ftl-controller/main.go
@@ -17,6 +17,7 @@ import (
 	"github.com/TBD54566975/ftl/backend/controller/dal"
 	"github.com/TBD54566975/ftl/backend/controller/scaling"
 	cf "github.com/TBD54566975/ftl/common/configuration"
+	"github.com/TBD54566975/ftl/common/projectconfig"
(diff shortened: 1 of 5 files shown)


cmd/ftl-controller/main.go
```diff
@@ -27,7 +27,7 @@ var cli struct {
 	ObservabilityConfig observability.Config `embed:"" prefix:"o11y-"`
 	LogConfig           log.Config           `embed:"" prefix:"log-"`
 	ControllerConfig    controller.Config    `embed:""`
-	ConfigFlag          []string             `name:"config" short:"C" help:"Paths to FTL project configuration files." env:"FTL_CONFIG" placeholder:"FILE[,FILE,...]"`
```
deniseli, line 30: nice catch

deniseli (approved):

PR #1873, opened 2024-06-25
fix: execute startup tasks after serving

worstell:
ensure controller is live and db container is provisioned before executing startup tasks


wesbillman:
This looks good to me! I wonder if there would ever be a time where we want to run startup tasks before starting the controller, etc. 🤔

worstell:
> This looks good to me! I wonder if there would ever be a time where we want to run startup tasks before starting the controller, etc. 🤔

ah interesting - i'm thinking startup tasks are more likely to involve provisioning resources or setting up state relating to the user code, rather than preconditions that would impact the actual FTL infrastructure, but lmk if you have other thoughts or if i'm missing something!

this fix was prompted because a downstream customer's `ftl serve` broke with the recent change migrating startup task executions over from `ftl dev`, because their tasks required a container be provisioned

PR #1882, opened 2024-06-26
feat: templated node creation

jonathanj-square:


extensions/vscode/src/commands/node-new.ts
```diff
@@ -23,8 +23,7 @@ export const nodeNewCommand = async (item: FtlTreeItem) => {
 
   //TODO: Add all the types here...
   // Also would be cool to have an httpingress type which would populate a valid //ftl:ingress ....
-  const nodeType = await vscode.window.showQuickPick(['verb', 'enum', 'pubsub', 'pubsub:subscription', 'fsm', 'database', 'config:string', 'config:struct', 'secret', 'cron'], {
-    title: 'Which type of node would you like to add',
+  const nodeType = await vscode.window.showQuickPick(['verb', 'enum', 'pubsub', 'pubsub:subscription', 'fsm', 'database', 'config:string', 'config:struct', 'secret', 'cron'], {    title: 'Which type of node would you like to add',
```
wesbillman, line 26: I think we should leave `title` on a new line here.

extensions/vscode/src/commands/node-new.ts
```diff
@@ -55,11 +54,11 @@ export const nodeNewCommand = async (item: FtlTreeItem) => {
   editor.revealRange(lastLineRange, vscode.TextEditorRevealType.Default)
 }
 
-const snippetForNodeType = (nodeType: string): string => {
+const snippetForNodeType = async (nodeType: string): Promise<string> => {
   //TODO: fill out for all node types.
```
wesbillman, line 58: We can remove this now that you're filling them out :)

extensions/vscode/src/commands/node-new.ts
```diff
@@ -94,17 +93,32 @@ const snippetForNodeType = (nodeType: string): string => {
...
-const verbSnippet = `type SampleRequest struct {
+const verbSnippetBuilder = async (): Promise<string> => {
+  const inputBoxOptions: vscode.InputBoxOptions = {
+    prompt: "Enter the verb's name",
+    placeHolder: 'Sample',
+  };
+  const name = await vscode.window.showInputBox(inputBoxOptions);
+
+  if (name) {
+    return verbSnippetTemplate.replace(/<name>/g, name)
```
wesbillman, line 104: We could make `verbSnippetTemplate` a func that takes in `name` as an argument

PR #1928, opened 2024-07-02
fix: make 1Password serial access less strict

matt2e:
Fixes this issue:
- Try deploying modules with a decent amount of secrets
- Initial deploys would work, and then those deployments would saturate 1password access while checking for updates every 1s
    - It takes about 1s for a call to `op` to return a result
- Subsequent deploys would struggle to succeed as secrets need to be gathered within 5s.

Also of note: always calling `op` serially would mean modules with ~5 secrets could never be deployed

Changes:
- `OnePasswordProvider` is less strict. If a call to `op` succeeded within a second ago then concurrent calls to `op` is allowed
- Controller only processes updates to secrets serially. This avoids the biggest cause of multiple 1Password prompts
    - This is skipped if 1Password is not a provider (ie: production)

Multiple 1Password prompts are possible, but hopefully minimized.
- In idle state, controllers will only be making 1 `op` call at a time (module update checks)
- In short bursts, more `op` calls can happen at a time
- Previously, after the computer was locked, we would prompt as many times as there were deployed modules with 1Password secrets

### DIFF
--- a/backend/controller/controller.go
+++ b/backend/controller/controller.go
@@ -191,6 +191,10 @@ type Service struct {
 
 	increaseReplicaFailures map[string]int
 	asyncCallsLock          sync.Mutex
+
+	// This lock is used to serialize access to secrets when checking for updates.
+	// Without this, it is too common for multiple 1Password prompts to be shown to the user at once.
+	secretUpdatesLock sync.Mutex
 }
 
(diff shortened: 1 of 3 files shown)


matt2e:
Ah correction:
> Also of note: always calling op serially would mean modules with ~5 secrets could never be deployed

This is still true because when building a module context we get each secret one after the other... 🤦‍♂️

PR #1929, opened 2024-07-02
fix: diagnostic errors were being discarded

### LINKED ISSUE #1849: Build error not showing up in VSCode
In this example I am trying to call a function that is not exported as a verb, but I imagine this is a more general problem:

image

gak:
Fixes #1849 

Error detection was far too aggressive, discarding legit schema.Errors.


alecthomas (approved):
I wish we could test the extension somehow :(

PR #1932, opened 2024-07-02
fix: show logs via command palette

gak:


extensions/vscode/package.json
```diff
@@ -60,12 +60,14 @@
         "category": "FTL"
       },
       {
-        "command": "ftl.showPanel",
-        "title": "Show Panel"
```
deniseli, line 64: Did you mean to remove Show Panel? Just confirming since I wasn't sure :)

deniseli (approved):
Nice!

extensions/vscode/package.json
gak: Yeah the bug was that it needs to be showLogs instead. I should have been more clear on why I changed it!

alecthomas (approved):

wesbillman (approved):
I did this in my hackathon, but I also want to move all these commands/handlers to a `./commands` folder to have this managed in a structured way. Hoping to get a PR with that refactor soon.

extensions/vscode/package.json
deniseli: Ahhh sweet thanks a bunch for the context :)

PR #1954, opened 2024-07-03
feat!: store all 1password secrets in a single entry

### LINKED ISSUE #1772: `secret set` 1password entry to warn humans

### LINKED ISSUE #1947: Use single 1Password entry for all secrets in a project
Use a separate field per secret

matt2e:
closes #1947
closes #1772
Creates an entry in 1Password called `<projectname>.secrets` with each secret stored in a password field called `<modulename>.<secretname>`
Username is set to a warning string as that is presented at the top of the 1Password UI.

This will break existing secrets stored in 1Password. Migration can be done using commands made available in: https://github.com/TBD54566975/ftl/pull/1982

### DIFF
--- a/common/configuration/1password_provider.go
+++ b/common/configuration/1password_provider.go
@@ -19,13 +19,18 @@ import (
 // OnePasswordProvider is a configuration provider that reads passwords from
 // 1Password vaults via the "op" command line tool.
 type OnePasswordProvider struct {
-	Vault string
+	Vault       string
+	ProjectName string
 }
 
 func (OnePasswordProvider) Role() Secrets { return Secrets{} }
 func (o OnePasswordProvider) Key() string { return "op" }
 func (o OnePasswordProvider) Delete(ctx context.Context, ref Ref) error {
 	return nil
+}
+
+func (o OnePasswordProvider) itemName() string {
+	return o.ProjectName + ".secrets"
 }
 
 // Load returns the secret stored in 1password.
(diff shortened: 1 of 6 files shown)


matt2e:
I don't plan on merging this in yet until I have more PRs ready that will help transition existing projects

common/configuration/1password_provider.go
```diff
@@ -28,21 +29,25 @@ func (o OnePasswordProvider) Delete(ctx context.Context, ref Ref) error {
 	return nil
 }
 
+func (o OnePasswordProvider) itemName() string {
+	return o.ProjectName + ".secrets"
```
deniseli, line 33: Does this need the `"ftl." +` prefix, or does that get added somewhere else?

common/configuration/1password_provider_test.go
```diff
@@ -42,33 +58,43 @@ func Test1PasswordProvider(t *testing.T) {
...
+	}
+
+	_, err = provider.getItem(ctx, vault)
 	assert.Error(t, err)
 
 	var pw1 = []byte("hunter1")
 	var pw2 = []byte(`{
 	  "user": "root",
-	  "password": "hunter🪤"
+	  "password": "hun\\ter🪤"
```
deniseli, line 75: hehehe nice!

deniseli (approved):

common/configuration/1password_provider.go
matt2e: I moved away from the `ftl.` prefix, but i hadn't updated the documentation. good catch!

PR #1955, opened 2024-07-03
fix: validate project config

matt2e:
Validates project config files.

This breaks previous cases where `ftl config set` and similar would be able to create the project toml if it wasn't already created. Now that a project's name is required, an invalid project config would have been created.

It is now a requirement to go through `ftl init <name> <dir>`

### DIFF
--- a/backend/controller/admin/admin_test.go
+++ b/backend/controller/admin/admin_test.go
@@ -62,7 +62,7 @@ func tempConfigPath(t *testing.T, existingPath string, prefix string) string {
 	var existing []byte
 	var err error
 	if existingPath == "" {
-		existing = []byte{}
+		existing = []byte("name = \"generated\"")
 	} else {
 		existing, err = os.ReadFile(existingPath)
 		assert.NoError(t, err)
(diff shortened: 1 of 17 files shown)


backend/controller/admin/admin_test.go
```diff
@@ -61,7 +61,7 @@ func tempConfigPath(t *testing.T, existingPath string, prefix string) string {
 	var existing []byte
 	var err error
 	if existingPath == "" {
-		existing = []byte{}
+		existing = []byte("name = \"generated\"")
```
wesbillman, line 64: nit: You might be able to use backticks here:
```go
`name = "generated"`
```

common/projectconfig/projectconfig.go
```diff
@@ -47,6 +47,26 @@ func (c Config) Root() string {
 	return filepath.Dir(c.Path)
 }
 
+// Validate checks that the configuration is valid.
+func (c Config) Validate() error {
```
wesbillman, line 51: Do we need a test for this one, or is it covered by other tests that use `Create`?

wesbillman (approved):
Do we need to update the root `ftl-project.toml` for FTL as well? 

I'm torn by the need to give the project a name when using `ftl init` but I guess it's inline with other tools like `npm init`.

matt2e:
> Do we need to update the root `ftl-project.toml` for FTL as well?
> 
> I'm torn by the need to give the project a name when using `ftl init` but I guess it's inline with other tools like `npm init`.

This was done in a previous PR: https://github.com/TBD54566975/ftl/pull/1952 👍

PR #1979, opened 2024-07-05
fix: console was crashing quietly, relative paths

gak:
- Now log why console crashed
- Use gitRoot based path for running console


alecthomas (approved):
Nice find.

PR #2008, opened 2024-07-08
fix: dont mark version as dirty temporarily and investigate git status

gak:


Justfile
```diff
@@ -51,6 +51,8 @@ build +tools: build-protos build-zips build-frontend
   #!/bin/bash
   shopt -s extglob
 
+  git diff
+
```
wesbillman, line 55: Is this for logging @gak?

backend/controller/cronjobs/sql/models.go
```diff
@@ -478,6 +478,14 @@ type ModuleConfiguration struct {
 	Value     []byte
 }
 
+type ModuleSecret struct {
+	ID        int64
+	CreatedAt time.Time
+	Module    optional.Option[string]
+	Name      string
+	Url       string
+}
```
wesbillman, line 487: I'm guessing this is unrelated, but needed to be checked in?

wesbillman (approved):

Justfile
gak: Yes, I need to see what is causing the repo to be dirty.

backend/controller/cronjobs/sql/models.go
gak: It is related to code generated in the CI causing the repo to be dirty. I'm under the impression these sqlc files needed to be checked in?

deniseli: Your impression is correct :) Looks like whoever added ModuleSecret to the db schema just didn't know they needed to run `just build-sqlc`.

PR #2037, opened 2024-07-10
fix: transitive type loop

worstell:


matt2e:
LGTM 👍 the new code looks good, but I can't tell what was wrong with the previous code 🤷‍♂️

PR #2038, opened 2024-07-10
chore: require build-sqlc to be run in pre-push if sql inputs are changed

### LINKED ISSUE #2013: Automate running `just build-sqlc`

safeer:
Fixes #2013 by adding a pre-push step to do a simple modified-time check on sql inputs and generated sqlc. Also fixes a small bug in `build-sqlc` where it couldn't find `.../{cronjobs}/...`

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -71,7 +71,11 @@ init-db:
 
 # Regenerate SQLC code (requires init-db to be run first)
 build-sqlc:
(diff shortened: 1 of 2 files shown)


worstell (approved):

safeer:
Closing this out; I think we should be smarter and do a diff instead of simple timestamps, but this isn't a priority

PR #2039, opened 2024-07-10
fix: sync stub references for each build group

wesbillman:
This PR modifies the build engine to generate stubs for all known schemas at build time. After a build group is built, we then sync the stub references to help users get auto completions and imports in their IDEs or text editors.

It essentially changes the way we gather schemas for the module(s) we're building.

In `go` syncing the stub references means that we update the `go.work` file for each built module to include ALL shared module stubs. This helps when a module is adding a shared module as a dependency.


worstell (approved):
yayyyyyyyy

Issue #2020, opened 2024-07-09
otel: implement `just otel`


deniseli:
Run [otel collector](https://opentelemetry.io/docs/collector/) in a docker container to collect local logs/metrics/traces/spans and show the stream as they come in in the terminal tab. This way, we can run ftl dev, trigger the events we plan to log, and see them come through the terminal window in real(ish) time.

Wes's command to start working with:
```
docker run \
  -p 4317:4317 \
  -p 55679:55679 \
  otel/opentelemetry-collector:0.104.0
```

deniseli:
#2078 is the last PR on this issue for now. We should ideally add [color](https://www.unix.com/unix-for-dummies-questions-and-answers/134824-using-sed-change-specific-words-color.html) + bolding to this as well, which can be done by piping to a very similar sed command, but moving forward without that for now to save time.

PR #2040, opened 2024-07-10
feat: implement just cmds `otel-stream`, `otel-stop`, and `otel-dev`

deniseli:
Part 1 of https://github.com/TBD54566975/ftl/issues/2020

This PR enables local testing of otel metric changes/additions. To use:

In tab 1, run `just otel-stream`.

In tab 2, run `just otel-dev`. You can pass in any args you'd run `ftl dev` with (e.g. `--recreate`). If you set `--log-level=DEBUG`, you should see `OTEL is enabled, exporting to localhost:4317` in your output. Any signals you trigger to log at this point should appear in the streaming output of tab 1.

When you're done, kill the `dev` command and then run `just otel-stop` in any tab other than tab 1.


deniseli:
Completely forgot to integrate https://github.com/tobert/otel-cli before making the PR so I'll do that now :p

deniseli:
Update: what we have here is already useful and needed for other folks to pick up tasks, so we're going to go ahead and merge it and I'll fix in the other improvements later.

To make this work locally, just open `internal/observability/client.go` and replace `otlpmetricgrpc.New(ctx)` with `otlpmetricgrpc.New(ctx, otlpmetricgrpc.WithInsecure(), otlpmetricgrpc.WithEndpoint(os.Getenv("OTEL_E\
XPORTER_OTLP_ENDPOINT")))`

Justfile
```diff
@@ -173,4 +173,28 @@ debug *args:
...
+# Stop the docker container running otel.
+otelContainerID := `docker ps -f ancestor=otel/opentelemetry-collector:0.104.0 | tail -1 | cut -d " " -f1`
+otel-stop:
+  docker stop "{{otelContainerID}}"
+
+# Run `ftl dev` with the given args after setting the necessary envar.
+otel-dev *args:
+  #!/bin/bash
+
+  gprcPort=$(cat docker-compose.yml | grep "OTLP gRPC" | sed 's/:.*//' | sed -r 's/ +- //')
```
gak, line 198: https://github.com/mikefarah/yq maybe easier in the future? (not an endorsement--I have not tried it!)

gak (approved):

Justfile
gak: Oh I see it's a comment, sorry that suggestion wouldn't work at all!

deniseli: oooo nice! I'll go ahead and merge with the current stuff so folks can use it but that would be much nicer. I'll try to get it working with the real parser in the next update

deniseli: Oh weird I just got your second comment on the comment gRPC string... yeah maybe it is actually safer to keep it like this. I'll think about it.

PR #2043, opened 2024-07-11
feat: ftl schema diff

### LINKED ISSUE #1989: `ftl schema diff` - diff local schema against a cluster
This will be necessary for production.

gak:
Fixes #1989

```
ftl schema diff --help
Usage: ftl schema diff <other-endpoint> [flags]

Print any schema differences between this cluster and another cluster. Returns
an exit code of 1 if there are differences.

Arguments:
  <other-endpoint>    Other endpoint URL to compare against.

  --endpoint=http://127.0.0.1:8892     FTL endpoint to bind/connect to ($FTL_ENDPOINT).
  --color    Enable colored output regardless of TTY.
```

- Compares `<other-endpoint>` vs `--endpoint` into a unified diff with color.
- Exits with 1 if there is a difference.
- Detects tty and will not emit ANSI unless `--color` is passed in.

image

### DIFF
--- a/cmd/ftl/cmd_schema.go
+++ b/cmd/ftl/cmd_schema.go
@@ -2,6 +2,7 @@ package main
 
 type schemaCmd struct {
 	Get      getSchemaCmd      `default:"" cmd:"" help:"Retrieve the cluster FTL schema."`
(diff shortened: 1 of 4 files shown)


worstell:
awesome

worstell (approved):

PR #2054, opened 2024-07-12
fix: expose otel port outside container

gak:


Justfile
```diff
@@ -196,5 +196,5 @@ otel-dev *args:
   #!/bin/bash
 
   gprcPort=$(cat docker-compose.yml | grep "OTLP gRPC" | sed 's/:.*//' | sed -r 's/ +- //')
```
wesbillman, line 198: I think `gprcPort` is defined above as well. Should we combine them?

wesbillman (approved):

deniseli (approved):
Yay!!

PR #2055, opened 2024-07-12
feat: asynchronous providers

### LINKED ISSUE #1949: Cache secrets at router
Allow each provider to sync (ASM, 1Password)

matt2e:
closes #1949

Changes:
- Splits `Provider` interface into:
    - `SynchronousProvider`: same as old interface, can load values from the data source on-demand
    - `AsynchronousProvider`: used by providers that prefer to sync with their data source
- `Manager` owns a `cache` which syncs asynchronous providers
    - When `Manager` is getting values from an asynchronous provider, it now accesses it via `cache`
    - When `Manager` is setting values for an asynchronous provider, it directly sets it with the provider and then notifies the cache
- `OnePasswordProvider` is now an `AsynchronousProvider` so that we don't cause multiple annoying authorization prompts
- `ASM` is now an `AsynchronousProvider` with it's previous `secretsCache` now being repurposed to work more generally.


common/configuration/cache.go
```diff
@@ -0,0 +1,281 @@
...
+			// treat this as a successful initial sync, so we don't ever block trying to access this provider
+			close(c.loaded)
+		})
+		return
+	}
+
+	c.lastSyncAttempt = optional.Some(clock.Now())
+	err := c.provider.Sync(ctx, c.values)
+	if err != nil {
+		logger.Errorf(err, "Error syncing %s", c.provider.Key())
```
gak, line 246: This happens when a provider fails? eg asm timeout? Does the backoff time out?

gak (approved):

common/configuration/cache.go
matt2e: This happens when a provider fails to sync (such as ASM timing out, or 1Password not getting authorized).
The backoff schedules the sync to try again in a future round of async provider syncs, starting at 1s later, exponentially growing and then maxing out at a 2 min maximum (`syncMaxBackoff`).
Does that answer your question @gak?

PR #2064, opened 2024-07-12
feat: migrate topics and subscriptions

worstell:


go-runtime/schema/common/common.go
```diff
@@ -509,6 +509,52 @@ func Deref[T types.Object](pass *analysis.Pass, node ast.Expr) (string, T) {
...
+		return optional.None[*ast.CallExpr]()
+	}
+	callExpr, ok := vs.Values[0].(*ast.CallExpr)
+	if !ok {
+		return optional.None[*ast.CallExpr]()
+	}
+	return optional.Some(callExpr)
+}
+
+func FuncPathEquals(pass *analysis.Pass, callExpr *ast.CallExpr, path string) bool {
```
wesbillman, line 533: nit: should we add comments to this func and the one above since they're public?

go-runtime/schema/data/analyzer.go
```diff
@@ -49,14 +49,10 @@ func extractData(pass *analysis.Pass, pos token.Pos, named *types.Named) optiona
 		Pos:  common.GoPosToSchemaPos(fset, pos),
 		Name: strcase.ToUpperCamel(named.Obj().Name()),
 	}
-	if md, ok := common.GetFactForObject[*common.ExtractedMetadata](pass, named.Obj()).Get(); ok {
-		if _, ok = md.Type.(*schema.Data); !ok && md.Type != nil {
-			return optional.None[*schema.Data]()
-		}
+	common.ApplyMetadata[*schema.Data](pass, named.Obj(), func(md *common.ExtractedMetadata) {
```
wesbillman, line 52: Nice!

go-runtime/schema/subscription/analyzer.go
```diff
@@ -0,0 +1,108 @@
...
+		// we will derive subscription name from generated variable name
+		moduleIdent, moduleOk := topicSelExp.X.(*ast.Ident)
+		if !moduleOk {
+			common.Errorf(pass, node, "subscription registration must have a topic")
+			return optional.None[*schema.Subscription]()
+		}
+		varName := topicSelExp.Sel.Name
+		topicRef = &schema.Ref{
+			Module: moduleIdent.Name,
+			Name:   strcase.ToLowerCamel(string(varName[0]) + varName[1:]),
```
wesbillman, line 92: is it safe to index into `varName` here without checking len?

wesbillman (approved):
Niiiiiiice

go-runtime/schema/subscription/analyzer.go
worstell: ah nice catch- i'm not sure, this logic is directly ported. i'll add a check to be safe

PR #2071, opened 2024-07-15
fix: no need to specify —opvault to access secrets

matt2e:
Changes:
- Asynchronous providers are given a list of known entries when syncing
- 1Password provider uses these entries to know which vaults to sync
- Removed logic for whether an asynchronous provider was active. This was there to make sure we didnt attempt syncing with 1Password for projects that weren't using 1Password. This is now solved by 1PasswordProvider only syncing known vaults.


gak (approved):
LGTM

Issue #2068, opened 2024-07-13
Could not deregister runner on new deployment


gak:
On deployment, `ftl-controller` can not deregister runners (maybe from the previous deploy).

```
Jul 14 00:34:38.761 ftl-controller Streaming RPC failed: runner callback failed: unavailable: failed to connect to runner: context deadline exceeded: /xyz.block.ftl.v1.ControllerService/RegisterRunner
Jul 14 00:34:38.761 ftl-controller Could not deregister runner http://xxx: not found
```

matt2e:
Looks like this is what's happening:
- Runner starts, starts stream to controller
- Controller receives first message of stream and tries to ping the runner https://github.com/TBD54566975/ftl/blob/main/backend/controller/controller.go#L572
    - Controller fails to ping runner, retries for 10s, then gives up (by cancelling context): https://github.com/TBD54566975/ftl/blob/main/backend/controller/controller.go#L612
    - Controller tries to deregister the runner https://github.com/TBD54566975/ftl/blob/main/backend/controller/controller.go#L565
        - This fails with `Could not deregister runner http://xxx: not found` because we have not yet reached the code that upserts the runner https://github.com/TBD54566975/ftl/blob/main/backend/controller/controller.go#L581

PR #2082, opened 2024-07-16
fix: only deregister runner if runner was ever inserted

matt2e:
fixes https://github.com/TBD54566975/ftl/issues/2068

Errors can occur after we were adding the `defer` and when we would try upserting the runner.
This PR makes it so we only add the defer once we have upserted the runner


gak (approved):
LGTM nice fix. This one is probably a bit tricky to repro in an integration test!

PR #2091, opened 2024-07-17
fix: only create config managers if needed to avoid syncing on every cmd

### LINKED ISSUE #2086: 1Password spamming prompts again
Happens after starting `ftl dev` with about ~12 prompts at first, and then more

matt2e:
fixes #2086
- Recently we added a feature for cf.Manager to sync providers (1Password & ASM)
- We should not create these managers if they are not used, otherwise we will prompt the user for 1Password access unnecessarily
- Previously we were creating config + secret managers in `main.go` for every command, even though it is only needed for these commands:
    - `ftl serve`
    - `ftl dev`
    - `ftl config ...` if there is no controller running
    - `ftl secret ...` if there is no controller running

### DIFF
--- a/cmd/ftl/cmd_config.go
+++ b/cmd/ftl/cmd_config.go
@@ -75,14 +75,14 @@ func setUpAdminClient(ctx context.Context, config projectconfig.Config) (ctxOut 
 		cr := cf.ProjectConfigResolver[cf.Configuration]{Config: config.Path}
 		cm, err := cf.NewConfigurationManager(ctx, cr)
 		if err != nil {
-			return client, err
+			return ctx, client, err
 		}
 		ctx = cf.ContextWithConfig(ctx, cm)
 
 		sr := cf.ProjectConfigResolver[cf.Secrets]{Config: config.Path}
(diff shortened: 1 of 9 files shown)


cmd/ftl/cmd_config.go
```diff
@@ -60,7 +64,38 @@ type configListCmd struct {
 	Module string `optional:"" arg:"" placeholder:"MODULE" help:"List configuration only in this module."`
 }
 
-func (s *configListCmd) Run(ctx context.Context, adminClient admin.Client) error {
+func setUpAdminClient(ctx context.Context, config projectconfig.Config) (ctxOut context.Context, client admin.Client, err error) {
```
deniseli, line 67: I'm not sure what the answer to this question is, just wondering: should we move this function, or part of this function, to another file? It is config-related, and it is cmd-related, but it's not really a cmd_config-_specific_ thing conceptually. I almost wonder if we should just have an analog to `backend/controller/admin/client.go` called `backend/controller/admin/local_client.go`...

deniseli: Just kidding, that file already exists :) In that case, I suppose my question is whether some or all of this logic belongs better in there. Not urgent at all of course

PR #2094, opened 2024-07-17
fix: remove pg_notify usage in favor of polling

### LINKED ISSUE #2087: Switch schema notifications from db to use polling

wesbillman:
Fixes #2087

### DIFF
--- a/backend/controller/controller.go
+++ b/backend/controller/controller.go
@@ -1,6 +1,7 @@
 package controller
 
 import (
+	"bytes"
 	"context"
 	sha "crypto/sha256"
 	"encoding/binary"
@@ -1516,7 +1517,7 @@ func (s *Service) updateControllersList(ctx context.Context) (time.Duration, err
 func (s *Service) watchModuleChanges(ctx context.Context, sendChange func(response *ftlv1.PullSchemaResponse) error) error {
 	logger := log.FromContext(ctx)
 	type moduleStateEntry struct {
-		hash        sha256.SHA256
+		hash        []byte
 		minReplicas int
 	}
 	moduleState := map[string]moduleStateEntry{}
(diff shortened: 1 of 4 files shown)


backend/controller/dal/notify.go
```diff
@@ -51,6 +53,88 @@ type event struct {
...
+}
+
+func (d *DAL) pollDeployments(ctx context.Context) {
+	logger := log.FromContext(ctx)
+	retry := backoff.Backoff{}
+
+	previousDeployments := make(map[string]deploymentState)
+
+	for {
+		delay := time.Millisecond * 500
```
wesbillman, line 82: Do we want this to be configurable maybe?

backend/controller/dal/notify.go
```diff
@@ -51,6 +53,88 @@ type event struct {
...
+	schemaHash  []byte
+	minReplicas int
+}
+
+func deploymentStateFromDeployment(deployment Deployment) (deploymentState, error) {
+	hasher := sha256.New()
+	data := []byte(deployment.Schema.String())
+	if _, err := hasher.Write(data); err != nil {
+		return deploymentState{}, fmt.Errorf("failed to hash schema: %w", err)
+	}
```
wesbillman, line 67: This is the hashing I had to change to. It seemed like `sha256.fromBytes` just didn't compute a different has at all, but maybe I was using it wrong.

backend/controller/controller.go
```diff
@@ -1575,16 +1576,19 @@ func (s *Service) watchModuleChanges(ctx context.Context, sendChange func(respon
...
-				if err != nil {
+
+				hasher := sha.New()
+				data := []byte(moduleSchema.String())
+				if _, err := hasher.Write(data); err != nil {
 					return err
 				}
+
 				newState := moduleStateEntry{
-					hash:        sha256.FromBytes(moduleSchemaBytes),
```
wesbillman, line 1583: This seemed to compute the same hash every time even if the schema changed.

gak (approved):

backend/controller/dal/notify.go
matt2e: I vote for getting it in as is and adding an option if there's a need

matt2e (approved):
LGTM!

PR #2102, opened 2024-07-17
fix: flaky asm tests fixed by wrapping async provider

### LINKED ISSUE #2092: ASM tests are flaky
logs: https://github.com/TBD54566975/ftl/actions/runs/9967125333/job/27540391199?pr=2091

matt2e:
fixes #2092
- mock clock was causing the same race conditions as this issue: https://github.com/TBD54566975/ftl/issues/1368#issuecomment-2087944187
- instead we are now wrapping the ASM provider with a manual sync provider
    - This wrapper allows us to trigger when syncs should happen in tests, and block until sync completes


matt2e:
@wesbillman if there's no PR concerns, feel free to merge if the flaky test is annoying

wesbillman (approved):

PR #2103, opened 2024-07-17
fix: setting secrets and configs error with aliased field names

wesbillman:
Fixes an error where 

```
type SomeSecret struct {
	FirstName string `json:"first-name"`
}
```

when set with `secret set` `{"first-name": "whatever"}`


backend/controller/ingress/ingress_test.go
```diff
@@ -401,9 +401,9 @@ func TestEnumValidation(t *testing.T) {
 		{&schema.Ref{Name: "TypeEnumRequest", Module: "test"}, obj{"message": obj{"name": "Invalid", "value": 0}},
 			"\"Invalid\" is not a valid variant of enum test.TypeEnum"},
 		{&schema.Ref{Name: "TypeEnumRequest", Module: "test"}, obj{"message": obj{"name": 0, "value": 0}},
-			`invalid type for enum "test.TypeEnum"; name field must be a string, was int`},
+			`failed to transform aliased fields: 0:0: expected 'name' field to be a string, got int`},
 		{&schema.Ref{Name: "TypeEnumRequest", Module: "test"}, obj{"message": "Hello"},
-			`malformed enum type test.TypeEnumRequest.message: expected structure is {"name": "<variant name>", "value": <variant value>}`},
+			`failed to transform aliased fields: 0:0: expected map, got string`},
```
wesbillman, line 406: @safeer we might want to come back and fix up these error messages if possible to make them more readable

mistermoe (approved):

backend/schema/jsonvalidate.go
```diff
@@ -211,6 +216,7 @@ func ValidateJSONValue(fieldType Type, path path, value any, sch *Schema) error
 	return nil
 }
 
+// ValidateRequestMap validates a given JSON map against the provided schema.
 func ValidateRequestMap(ref *Ref, path path, request map[string]any, sch *Schema) error {
```
deniseli, line 220: Note for future reference: 
Since this function is only called by `ingress.BuildRequestBody` and `ValidateJSONValue` above, we could have this function call `TransformFromAliasedFields` directly. Then, the caller doesn't need to know to do it, so we can prevent issues like this in the future.

Discussed on Slack - we'll leave this as is for now to keep everything as stable as possible

Issue #2084, opened 2024-07-16
Updating a secret with `--asm` will return an error, but appears to work


wesbillman:
When attempting to update an ASM secret, we get the following errors.

```sh
❯ ftl --endpoint=https://[redacted] secret set --asm [redacted].FTL_DSN_[redacted]_[redacted]
Secret:
ftl: error: unknown: failed to set secret: failed to set secret URL: could not set secret URL: duplicate key value violates unique constraint "module_secrets_module_name_key": conflict

❯ ftl --endpoint=https://[redacted] secret unset --asm [redacted].FTL_DSN_[redacted]_[redacted]

❯ ftl --endpoint=https://[redacted] secret set --asm [redacted].FTL_DSN_[redacted]_[redacted]
Secret:
ftl: error: unknown: failed to set secret: asm/leader: unable to store secret in ASM: operation error Secrets Manager: CreateSecret, https response error StatusCode: 400, RequestID: 498ad9de-f3d5-466d-a0b1-733ee040cb03, InvalidRequestException: You can't create this secret because a secret with this name is already scheduled for deletion.

❯ ftl --endpoint=https://[redacted] secret set --asm [redacted].FTL_DSN_[redacted]_[redacted]
Secret:
ftl: error: unknown: failed to set secret: failed to set secret URL: could not set secret URL: duplicate key value violates unique constraint "module_secrets_module_name_key": conflict
```

Likely coming from here when a secret already exists:

https://github.com/TBD54566975/ftl/blob/main/common/configuration/asm_leader.go#L139-L164

matt2e:
Putting this back into the `next` pile because I didn't get to it today

deniseli:
Looks like `duplicate key value violates unique constraint "module_secrets_module_name_key": conflict` comes from the `module_secrets` table's [uniqueness constraint](https://github.com/TBD54566975/ftl/blob/main/backend/controller/sql/schema/20240704103403_create_module_secrets.sql#L10) coupled with the query `SetModuleSecretURL` just running `INSERT INTO` without an `ON CONFLICT` [clause](https://stackoverflow.com/questions/63969970/how-to-update-or-insert-in-postgresql). This logic was largely copied from the `module_config`, so we'll need to fix that too (though lower priority).

deniseli:
[...]

PR #2105, opened 2024-07-17
fix: avoid insertion key dupls in `module_secrets` table by upserting on conflict

deniseli:
Addresses part of https://github.com/TBD54566975/ftl/issues/2084

Postgresql way to handle upserts: https://www.postgresqltutorial.com/postgresql-tutorial/postgresql-upsert/

This PR fixes the issue for secrets since that's causing issues for the client team, but not for config, which has the same issue. I confirmed the test fails before this change but passes with it.


matt2e (approved):
👌

PR #2109, opened 2024-07-18
feat: kubernetes dev deployment

gak:
See deployment/README.md for details and usage.

`ftl deploy` doesn't work currently because of cross compiling issues. Pushing this PR out now so it can be used at least.


matt2e (approved):
LGTM but i'm no k8s expert

PR #2113, opened 2024-07-18
fix: more errors wrapped around deploy/reserve runners

gak:


backend/controller/controller.go
```diff
@@ -1453,13 +1453,14 @@ func (s *Service) terminateRandomRunner(ctx context.Context, key model.Deploymen
...
 	client, err := s.reserveRunner(ctx, reconcile)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to reserve runner for %s %s: %w", reconcile.Key, reconcile.Module, err)
 	}
 
 	_, err = client.runner.Deploy(ctx, connect.NewRequest(&ftlv1.DeployRequest{DeploymentKey: reconcile.Key.String()}))
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to request deploy %s %s: %w", reconcile.Key, reconcile.Module, err)
```
matt2e, line 1461: I think module is redundant here because each deployment key is `dpl-modulename-randombytes`

backend/controller/controller.go
```diff
@@ -1471,14 +1472,19 @@ func (s *Service) reserveRunner(ctx context.Context, reconcile model.Deployment)
...
-		return clients{}, fmt.Errorf("failed to claim runners for %s: %w", reconcile.Key, err)
+		return clients{}, fmt.Errorf("failed to claim runners for %s %s: %w", reconcile.Key, reconcile.Module, err)
 	}
 
 	err = dal.WithReservation(reservationCtx, claim, func() error {
 		client = s.clientsForEndpoint(claim.Runner().Endpoint)
 		_, err = client.runner.Reserve(reservationCtx, connect.NewRequest(&ftlv1.ReserveRequest{DeploymentKey: reconcile.Key.String()}))
-		return err
+		if err != nil {
+			return fmt.Errorf("failed to request to reserve a runner for %s %s: %w", reconcile.Key, reconcile.Module, err)
```
matt2e, line 1482: I think we can get rid of the first `to` to make it read a bit better

matt2e: Let's add the endpoint here though as thats the missing piece of info we need to know if we think it should be able to connect ok

matt2e (approved):

PR #2116, opened 2024-07-19
fix: experimental fix for runner connection failures

matt2e:
We are storing clients per endpoint. During FTL upgrade we may see controllers try to connect to runners at an endpoint and fail but keep the client around. This may cause the RBAC issues we've seen in production.

This changes the cache to work based on runner key instead of endpoint so that a new runner at an old endpoint will not re-use an old client.


gak (approved):
LGTM

wesbillman (approved):

Issue #1744, opened 2024-06-12
Use an existing cron go package to reduce code complexity


gak:
Our cron code does work well, however it adds a fair amount of complexity to the code that we don't want to have to maintain when we can use existing functionality.

https://pkg.go.dev/github.com/hashicorp/cronexpr

Looks reasonably maintained and seems to support the various cron formats we currently support.

PR #2126, opened 2024-07-22
feat: use hashicorp/cronexpr for cron

stuartwdouglas:
This replaces our cron calculation with those provided by a 3rd party library.

Unfortunatly we still need to do our own parsing however, as we support additional syntax, and other parts of the code expect to be able to parse the Pattern using participle.

closess #1744


backend/controller/cronjobs/cronjobs.go
```diff
@@ -307,7 +307,7 @@ func (s *Service) watchForUpdates(ctx context.Context) {
 		now := s.clock.Now()
 		next := now.Add(time.Hour) // should never be reached, expect a different signal long beforehand
 		for _, j := range state.jobs {
-			if possibleNext, err := s.nextAttemptForJob(j, state, false); err == nil && possibleNext.Before(next) {
+			if possibleNext, err := s.nextAttemptForJob(j, state, now,false); err == nil && possibleNext.Before(next) {
```
wesbillman, line 310: You might need to setup your editor/formatter it looks like. This type of stuff `state, now,false` should autofix when you save

backend/protos/xyz/block/ftl/v1/schema/schema.proto
```diff
@@ -80,8 +80,7 @@ message EnumVariant {
   Value value = 4;
 }
 
-message Error {
-  enum ErrorLevel {
+message Error {  enum ErrorLevel {
```
wesbillman, line 83: Similar strange formatting here.

wesbillman:
This is awesome! 🤩

internal/cron/cron_test.go
```diff
@@ -233,7 +233,7 @@ func TestSeries(t *testing.T) {
 			"0 0 0 * * 1/2 *",
 			time.Date(2023, 12, 31, 23, 59, 0, 0, time.UTC),
 			time.Date(2024, 1, 31, 0, 0, 0, 0, time.UTC),
-			18,
+			14,
```
gak, line 236: Just wondering, why did these tests change?

gak (approved):

internal/cron/cron_test.go
stuartwdouglas: They changed due to differences in how the cron implementation behaves. The old implementation treated Sunday as both a 0 and a 7, so it was both even and odd. The new implementation just treats it as a 0.

PR #2130, opened 2024-07-22
chore: first metric (verb call counter)

deniseli:
Instrument our first metric. Next: check if this metric pipes through to DD correctly.

Locally, the output via `just otel-stream` looks like:
```
ScopeMetrics #1
ScopeMetrics SchemaURL: 
InstrumentationScope ftl.call 

Metric #0
Descriptor:
     -> Name: requests
     -> Description: Count of FTL verb calls via the controller
     -> Unit: 
     -> DataType: Sum
     -> IsMonotonic: true

NumberDataPoints #0
Data point attributes:
     -> ftl.module.name: Str(time)
     -> ftl.verb.name: Str(time)
StartTimestamp: 2024-07-23 00:25:50.782759 +0000 UTC
Timestamp: 2024-07-23 00:25:59.069408 +0000 UTC
Value: 1

NumberDataPoints #1
Data point attributes:
     -> ftl.module.name: Str(echo)
     -> ftl.verb.name: Str(echo)
StartTimestamp: 2024-07-23 00:25:50.782759 +0000 UTC
Timestamp: 2024-07-23 00:25:59.069408 +0000 UTC
Value: 1
	{"kind": "exporter", "data_type": "metrics", "name": "debug"}
2024-07-23T00:26:00.745Z	info	MetricsExporter	{"kind": "exporter", "data_type": "metrics", "name": "debug", "resource metrics": 1, "metrics": 10, "data points": 15}
2024-07-23
T00:26:00.745Z	info	ResourceMetrics #0
Resource SchemaURL: https://opentelemetry.io/schemas/1.25.0
Resource attributes:
     -> service.name: Str(echo)
     -> service.version: Str(HEAD)
     -> telemetry.sdk.language: Str(go)
     -> telemetry.sdk.name: Str(opentelemetry)
     -> telemetry.sdk.version: Str(1.27.0)
```

Interestingly, it looks like the metric logging races in a slightly inconsistent way. If you call a verb that calls another verb, you would expect just one verb to get bumped to 1 first, and then the other verb. Sometimes, it does behave that way. However, other times, the first time it appears in `otel-stream` _will already_ be with both verbs. So essentially, there is no guarantee that transient values actually show up in the stream, but the ultimate value will be correct.

This does appear the report the correct count in even error cases. e.g. I tried calling `echo` before the deploy was done, so that the verb call would fail without calling out to `time`. It would correctly log a single call in that case. And if you made another successful call afterwards, the metric would correctly increment to include both successful calls.

### DIFF
--- a/backend/controller/controller.go
+++ b/backend/controller/controller.go
@@ -824,17 +824,6 @@ func (s *Service) AcquireLease(ctx context.Context, stream *connect.BidiStream[f
 }
 
 func (s *Service) Call(ctx context.Context, req *connect.Request[ftlv1.CallRequest]) (*connect.Response[ftlv1.CallResponse], error) {
-	logger := log.FromContext(ctx)
-
-	requestCounter, err := otel.GetMeterProvider().Meter("ftl.verb").Int64Counter(
-		"requests",
-		metric.WithDescription("Count of FTL verb calls via the controller"))
-	if err != nil {
-		logger.Errorf(err, "Failed to instrument otel metric `ftl.call.request`")
-	} else {
-		requestCounter.Add(ctx, 1, metric.WithAttributes(attribute.String("ftl.module.name", req.Msg.Verb.Module), attribute.String("ftl.verb.name", req.Msg.Verb.Name)))
-	}
-
 	return s.callWithRequest(ctx, req, optional.None[model.RequestKey](), "")
 }
 
@@ -944,6 +933,18 @@ func (s *Service) callWithRequest(
 	sourceAddress string,
 ) (*connect.Response[ftlv1.CallResponse], error) {
 	start := time.Now()
+
+	logger := log.FromContext(ctx)
+
+	requestCounter, err := otel.GetMeterProvider().Meter("ftl.call").Int64Counter(
+		"requests",
+		metric.WithDescription("Count of FTL verb calls via the controller"))
+	if err != nil {
+		logger.Errorf(err, "Failed to instrument otel metric `ftl.call.requests`")
+	} else {
+		requestCounter.Add(ctx, 1, metric.WithAttributes(attribute.String("ftl.module.name", req.Msg.Verb.Module), attribute.String("ftl.verb.name", req.Msg.Verb.Name)))
+	}
+
 	if req.Msg.Verb == nil {
 		return nil, connect.NewError(connect.CodeInvalidArgument, errors.New("verb is required"))
 	}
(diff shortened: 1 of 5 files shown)


backend/controller/controller.go
```diff
@@ -821,6 +823,18 @@ func (s *Service) AcquireLease(ctx context.Context, stream *connect.BidiStream[f
 }
 
 func (s *Service) Call(ctx context.Context, req *connect.Request[ftlv1.CallRequest]) (*connect.Response[ftlv1.CallResponse], error) {
+	logger := log.FromContext(ctx)
+
+	requestCounter, err := otel.GetMeterProvider().Meter("ftl_call").Int64Counter(
```
wesbillman, line 828: do we want this meter to be `ftl.call` to match the spreadsheet names?

backend/controller/controller.go
```diff
@@ -821,6 +823,18 @@ func (s *Service) AcquireLease(ctx context.Context, stream *connect.BidiStream[f
...
+		"verb_call",
+		metric.WithDescription("Count of FTL verb calls via the controller"),
+		metric.WithUnit("requests"))
```
wesbillman, line 831: I think `WithUnit` is supposed to use a UCUM based on the docs:
```go
// WithUnit sets the instrument unit.
//
// The unit u should be defined using the appropriate [UCUM](https://ucum.org) case-sensitive code.
func WithUnit(u string) InstrumentOption { return unitOpt(u) }
```

https://ucum.org/ucum#section-Alphabetic-Index-By-Name

internal/observability/client.go
```diff
@@ -40,7 +40,9 @@ func Init(ctx context.Context, serviceName, serviceVersion string, config Config
 		return nil
 	}
 
-	logger.Debugf("OTEL is enabled, exporting to %s", os.Getenv("OTEL_EXPORTER_OTLP_ENDPOINT"))
+	logger.Debugf("OTEL is enabled, exporting to %s at log level %s",
+		os.Getenv("OTEL_EXPORTER_OTLP_ENDPOINT"),
```
wesbillman, line 44: do you want this extra newline here?

backend/controller/controller.go
deniseli: yes :)

backend/controller/controller.go
deniseli: ahh thanks for catching that, looks like these are all measurement units, so I'll see if we can just omit unit.

deniseli: indeed we can!

internal/observability/client.go
deniseli: I generally find things like this improve readability, but we also don't do that anywhere else in the codebase, so I'll just remove it. Probably not very Golang of me :p

wesbillman (approved):

deniseli:

PR #2131, opened 2024-07-22
feat: metrics instrumentation

jonathanj-square:


backend/controller/observability/observability.go
```diff
@@ -0,0 +1,207 @@
...
+	"github.com/alecthomas/types/optional"
+
+	"github.com/TBD54566975/ftl/backend/controller/dal"
+	ftlv1 "github.com/TBD54566975/ftl/backend/protos/xyz/block/ftl/v1"
+	"github.com/TBD54566975/ftl/backend/schema"
+	"github.com/TBD54566975/ftl/internal/log"
+	"github.com/TBD54566975/ftl/internal/model"
+)
+
+const name = "ftl.xyz/ftl/runner"
```
wesbillman, line 19: I'm not sure this is correct. Many of these metrics can come from the controller if I'm reading this correctly. I also think `runner` specific metrics should be handled in `backend/runner/` packages probably.

backend/controller/observability/observability.go
```diff
@@ -0,0 +1,207 @@
...
+	meter      metric.Meter
+	attributes metricAttributeBuilders
+	calls      *callMetrics
+	fsm        *fsmMetrics
+}
+
+var (
+	metrics = observableMetrics{
+		meter: otel.Meter(name),
+		// TODO: move to a initialization method
```
wesbillman, line 50: I might be missing something here, but based on how I read the spreadsheet for these, it seems like we would have a `meter` called something like `ftl.call` then that meter would contain counter/etc. like `requests` and `latency`. I'm still trying to make the names/columns in the spreadsheet to their counterparts here.

wesbillman (commented):
I like the approach of a centralized place for metrics details 👍

wesbillman:
@jonathanj-square I wonder if we can make a simplified version of this that won't require `init()`, etc. I was thinking something like this:

Folder structure might be:
```
- backend/controller
  - observability
    - call.go
    - pubsub.go
    - fsm.go
    - etc....
```

Essentially, 1 file for each "signal" in the spreadsheet.

Then for each of these files we could define a `meter` and use it for all the metrics. This is a very basic not complete example based on the metric @deniseli recently added as a test.

```go
var meter = otel.GetMeterProvider().Meter("ftl.call")

func RecordCallRequest(ctx context.Context, req *connect.Request[ftlv1.CallRequest]) {
	logger := log.FromContext(ctx)
	requestCounter, err := meter.Int64Counter(
		"ftl.call.requests",
		metric.WithDescription("Count of FTL verb calls via the controller"))

	if err != nil {
		logger.Errorf(err, "Failed to instrument otel metric `ftl.call.requests`")
		return
	}

	requestCounter.Add(ctx, 1, metric.WithAttributes(
		attribute.String("ftl.module.name", req.Msg.Verb.Module),
		attribute.String("ftl.verb.name", req.Msg.Verb.Name),
	))
}
```

This approach might help us easily add new metrics and scan through each "feature" to see what metrics are recorded and what attributes they have.

backend/controller/observability/observability.go
```diff
@@ -0,0 +1,207 @@
...
+	RequestKey    model.RequestKey
+	StartTime     time.Time
+	DestVerb      *schema.Ref
+	Callers       []*schema.Ref
+	Request       *ftlv1.CallRequest
+	Response      optional.Option[*ftlv1.CallResponse]
+	CallError     optional.Option[error]
+}
+
+func init() {
```
deniseli, line 84: It'd be nice not having metrics across so many different domains bundled into this same file. This file is going to get pretty dang long over time, and then someone will have to refactor it into separate smaller files anyway.

backend/controller/observability/observability.go
```diff
@@ -0,0 +1,207 @@
...
+	metrics.calls.requests, _ = metrics.meter.Int64Counter("ftl.call.requests",
+		metric.WithDescription("number of verb calls"),
+		metric.WithUnit("{count}"))
```
deniseli, line 87: The unit names all have to come from here so I'm pretty sure we have to leave the counts empty: https://ucum.org/ucum#section-Alphabetic-Index-By-Name

backend/controller/observability/observability.go
```diff
@@ -0,0 +1,195 @@
...
+			moduleName: func(name string) attribute.KeyValue {
+				return attribute.String("ftl.module.name", name)
+			},
+			featureName: func(name string) attribute.KeyValue {
+				return attribute.String("ftl.feature.name", name)
+			},
+			destinationVerb: func(name string) attribute.KeyValue {
+				return attribute.String("ftl.verb.dest", name)
+			},
+		},
```
deniseli, line 58: It feels like a lot of this could be simplified/shortened to just `const` declarations of the attr names. Since below, it's already a full line, we would save a bit of complexity here and not lose anything below. Ex: 

```
moduleAttr := metrics.attributes.moduleName(fsm.Module)
```
to
```
moduleAttr := attribute.String(ATTR_MODULE, fsm.Module)
```

It's nice having the attributes associated with their metrics, but it's already associated below since we need to actually instantiate all the attrs below anyway. So we don't need to duplicate those associations up here

backend/controller/observability/observability.go
```diff
@@ -0,0 +1,195 @@
...
+	meter       metric.Meter
+	active      metric.Int64UpDownCounter
+	transitions metric.Int64Counter
+}
+
+type observableMetrics struct {
+	attributes metricAttributeBuilders
+	calls      *callMetrics
+	fsm        *fsmMetrics
+}
```
deniseli, line 43: Quite a bit of this is duplicative as well, so we could just get rid of the structs. We need to instantiate the metrics below anyway, so this is just another site we'd need to add a line whenever we need to add a new metric.

backend/controller/controller.go
```diff
@@ -1005,22 +1006,27 @@ func (s *Service) callWithRequest(
...
+	})
+
 	response, err := client.verb.Call(ctx, req)
 	var resp *connect.Response[ftlv1.CallResponse]
 	var maybeResponse optional.Option[*ftlv1.CallResponse]
 	if err == nil {
 		resp = connect.NewResponse(response.Msg)
 		maybeResponse = optional.Some(resp.Msg)
 	}
-	s.recordCall(ctx, &Call{
```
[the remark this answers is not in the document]
jonathanj-square: the call is still there its renamed. I've introduced metrics that require instrumentation at call start and call end so the name `recordCall` doesn't fit as well as `RecordCallBegin` and `RecordCallEnd`

backend/controller/observability/observability.go
jonathanj-square: yep good call, I'll incorporate that input as I break out the metrics into individual feature based metrics (e.g. observability/calls.go, etc)

backend/controller/observability/observability.go
deniseli: ahhhh I totally missed @wesbillman 's sweet comment [here](https://github.com/TBD54566975/ftl/pull/2131#issuecomment-2245740496) when I first wrote this. I love that idea! Lesdoit

backend/controller/controller.go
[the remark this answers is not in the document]
deniseli: Ahhh got it, that makes sense

backend/controller/observability/observability.go
jonathanj-square: There is a difference between the two options worth noting. Attributes have a type associated with them and the second option decouples name and type. Attributes do have a schema to them and the proposal opens up the possibility of accidental type mismatching - I'm not sure where the break would end up but since there is no schema registration up front any breakage would occur at run-time or collection time. The first option is meant to defend against that error class.

backend/controller/observability/observability.go
jonathanj-square: sounds good, breaking this up now

deniseli:
Related to @wesbillman 's comment above: here's what I put together real quick locally to play with using attrs for success/failure: https://github.com/TBD54566975/ftl/compare/dli/call-counts?expand=1

It's convenient being able to pass arrays of attributes around for logging the separate variant states of each metric. So it's beyond the scope of what you have now but we'll probably want to incorporate that before merging

backend/controller/observability/observability.go
deniseli: Great point. Fortunately, it looks like the typecheck is built into the attr package :) https://pkg.go.dev/go.opentelemetry.io/otel/attribute#String

deniseli: I'm pretty sure that should fail to build but let me confirm

deniseli: yeah confirmed:
```
backend/controller/observability/call.go:36:44: cannot use "msg" (untyped string constant) as int value in argument to attribute.Int
```

deniseli: Got it. Helper functions do make sense for that. Let's just make sure we're only creating those helper functions once for each attribute. My main concern here was with trying to capture the associations of metrics with attributes in the `observableMetrics` struct.

backend/controller/observability/observability.go
jonathanj-square: dropped and went with other naming feedback (e.g. `fsm.call`)

the xyz is a domain that I've seen used in conjunction with ftl. The domain/path was motivated by some of the example otel integration code I learned from while ramping. I personally prefer the naming scheme that you two have suggested here.

backend/controller/observability/observability.go
jonathanj-square: change applied w/ the observability.go file decomposition into feature specific files.

backend/controller/observability/observability.go
jonathanj-square: adopted

backend/controller/dal/fsm.go
```diff
@@ -57,11 +58,15 @@ func (d *DAL) StartFSMTransition(ctx context.Context, fsm schema.RefKey, executi
 		}
 		return fmt.Errorf("failed to start FSM transition: %w", err)
 	}
+
+	observability.RecordFsmTransitionBegin(ctx, fsm)
```
deniseli, line 62: nit: go convention is to fully capitalize all initialisms, so this would be `RecordFSMTransitionBegin`. Same elsewhere

backend/controller/observability/calls.go
```diff
@@ -0,0 +1,133 @@
...
+	Request       *ftlv1.CallRequest
+	Response      optional.Option[*ftlv1.CallResponse]
+	CallError     optional.Option[error]
+}
+
+func initCallMetrics() {
+	callInitOnce.Do(func() {
+		callRequests, _ = callMeter.Int64Counter("ftl.call.requests",
+			metric.WithDescription("number of verb calls"),
+			metric.WithUnit("{count}"))
```
deniseli, line 46: see previous comment on units

backend/controller/observability/calls.go
```diff
@@ -0,0 +1,133 @@
...
+		callFailures, _ = callMeter.Int64Counter("ftl.call.failures",
```
deniseli, line 47: Weren't we going to make this an attr of request instead of a separate metric?

backend/controller/observability/calls.go
```diff
@@ -0,0 +1,133 @@
...
+	StartTime     time.Time
+	DestVerb      *schema.Ref
+	Callers       []*schema.Ref
+	Request       *ftlv1.CallRequest
+	Response      optional.Option[*ftlv1.CallResponse]
+	CallError     optional.Option[error]
+}
+
+func initCallMetrics() {
+	callInitOnce.Do(func() {
```
deniseli, line 43: Why did you use a sync.Once here instead of just `var callRequests = ...` at the package scope?

backend/controller/observability/observability.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"go.opentelemetry.io/otel/attribute"
+)
+
+type metricAttributeBuilders struct {
+	moduleName      func(name string) attribute.KeyValue
+	featureName     func(name string) attribute.KeyValue
+	destinationVerb func(name string) attribute.KeyValue
+}
```
deniseli, line 11: This is a bit excessive, could just do:

```
func attrModuleName(name string) attribute.KeyValue {...}
func attrFeatureName(name string) attribute.KeyValue {...}
...
```

Otherwise, it's just more lines to change/maintain.

backend/runner/runner.go
```diff
@@ -39,6 +42,13 @@ import (
 	"github.com/TBD54566975/ftl/internal/unstoppable"
 )
 
+const name = "ftl.xyz/ftl/go/runner"
```
deniseli, line 45: let's go back to the naming scheme in your google sheet - that's way nicer :)

backend/runner/runner.go
```diff
@@ -39,6 +42,13 @@ import (
...
+
+var (
+	meter           = otel.Meter(name)
+	instanceCounter metric.Int64UpDownCounter
+)
```
deniseli, line 50: Could we pull this out into a separate file like you did for the controller? It would be best for all the otel stuff to live in dedicated files. That makes it more obvious to a reader what the pattern is, so when we add new features in the future, it's easy to copy the instrumentation pattern

backend/runner/runner.go
```diff
@@ -107,6 +117,18 @@ func Start(ctx context.Context, config Config) error {
 	go rpc.RetryStreamingClientStream(ctx, backoff.Backoff{}, controllerClient.RegisterRunner, svc.registrationLoop)
 	go rpc.RetryStreamingClientStream(ctx, backoff.Backoff{}, controllerClient.StreamDeploymentLogs, svc.streamLogsLoop)
 
+	instanceCounter, err = meter.Int64UpDownCounter("ftl.sys.runner.instance",
```
deniseli, line 120: Let's make all the metric names `{meter}.{counter}`. Rather than hardcode that naming pattern, we can just set `var meterName` in each individual otel file, and then all the counter names can be constructed as: `fmt.Sprintf("%s.requests", meterName)` (subbing `request` for whatever the counter name is)

backend/runner/runner.go
```diff
@@ -107,6 +117,18 @@ func Start(ctx context.Context, config Config) error {
...
+		metric.WithDescription("number of runner instances"),
+		metric.WithUnit("{count}"))
+
+	if err != nil {
+		panic(err)
```
deniseli, line 125: We probably don't want to crash the whole runner when otel instrumentation fails. Could you take a look at how the existing code users the logger from ctx?

backend/runner/runner.go
```diff
@@ -107,6 +117,18 @@ func Start(ctx context.Context, config Config) error {
...
+	}
+
+	moduleNameAttribute := attribute.String("ftl.module.name", "unknown-module")
```
deniseli, line 128: This should use your helper function to guarantee consistency, right?

cmd/ftl/cmd_serve.go
```diff
@@ -84,6 +86,10 @@ func (s *serveCmd) run(ctx context.Context, projConfig projectconfig.Config, ini
 		return errors.New(ftlRunningErrorMsg)
 	}
 
+	if err := observability.Init(ctx, "ftl-dev", ftl.Version, s.ObservabilityConfig); err != nil {
+		return fmt.Errorf("failed to initialize observability: %w", err)
+	}
+
```
deniseli, line 92: no longer necessary since Saf's change merged

backend/controller/observability/calls.go
```diff
@@ -0,0 +1,133 @@
...
+
+	var featureName attribute.KeyValue
+	var moduleName attribute.KeyValue
+	if len(call.Callers) > 0 {
+		featureName = metricAttributes.featureName(call.Callers[0].Name)
+		moduleName = metricAttributes.moduleName(call.Callers[0].Module)
+	} else {
+		featureName = metricAttributes.featureName("unknown")
+		moduleName = metricAttributes.moduleName("unknown")
+	}
```
deniseli, line 70:
```
moduleNameAttr := attrModuleName(lookupModuleName(call.Callers))
```

or even simpler:

```
callActive.Add(ctx, 1, metric.WithAttributes(
    attrModuleName(lookupModuleName(call.Callers)), 
    /* rest of the attrs */))
```

deniseli: same below

backend/controller/observability/calls.go
```diff
@@ -0,0 +1,133 @@
...
+			metric.WithDescription("number of in flight calls"),
+			metric.WithUnit("{count}"))
+		callLatency, _ = callMeter.Int64Histogram("ftl.call.latency",
+			metric.WithDescription("verb call latency"),
+			metric.WithUnit("{ms}"))
+	})
+}
+
+func RecordCallBegin(ctx context.Context, call *CallBegin) {
+	initCallMetrics()
```
wesbillman, line 60: I think I'm missing the context on why we need to have an `init` in these metric funcs. It would be awesome if they were just plain old functions that could define the metric and attributes, without having to initialize the other structures. It might also remove the need for `sync.Once` code as well. Maybe we can sync on why the init stuff is required.

backend/runner/runner.go
jonathanj-square: applied in new branch

backend/runner/runner.go
jonathanj-square: noted, this meter is also no longer in scope. will apply this feedback in other metric inits

backend/runner/runner.go
jonathanj-square: yep

backend/controller/observability/observability.go
jonathanj-square: switched to explicit functions in the other PR

backend/controller/observability/calls.go
jonathanj-square: The otel getting started guides pre-initialize their metrics and use the initialized metrics in their instrumentation. The initialization metric process is a black box (for me at least) so my preference is to avoid risking introducing heavy weight operations in instrumentation code

wesbillman: Gotcha! I'm cool with that approach too. I might suggest having like a `New` function on these and maybe an overall `New` that calls the individual "feature" metric files. Then we can just `New` it up when we need it vs. having to have these init funcs everywhere.

wesbillman:
@jonathanj-square do you still want this branch now that we have the other implementations?

PR #2134, opened 2024-07-23
fix: new client for each runner after 5 minutes

matt2e:
Looking into RBAC issue and this may give us a clue.

### DIFF
--- a/backend/controller/controller.go
+++ b/backend/controller/controller.go
@@ -184,8 +184,9 @@ var _ ftlv1connect.ControllerServiceHandler = (*Service)(nil)
 var _ ftlv1connect.VerbServiceHandler = (*Service)(nil)
 
 type clients struct {
-	verb   ftlv1connect.VerbServiceClient
-	runner ftlv1connect.RunnerServiceClient
+	verb       ftlv1connect.VerbServiceClient
+	runner     ftlv1connect.RunnerServiceClient
+	created_at time.Time
 }
 
 // ControllerListListener is regularly notified of the current list of controllers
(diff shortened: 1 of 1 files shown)


backend/controller/controller.go
```diff
@@ -1150,14 +1151,16 @@ func (s *Service) getDeployment(ctx context.Context, key string) (*model.Deploym
...
 	clientItem := s.clients.Get(key.String())
-	if clientItem != nil {
+	if clientItem != nil && clientItem.Value().created_at.Add(time.Minute*5).After(time.Now()) {
 		return clientItem.Value()
 	}
+	log.FromContext(ctx).Tracef("Creating new grpc clients for %s at %s", key, endpoint)
 	client := clients{
-		runner: rpc.Dial(ftlv1connect.NewRunnerServiceClient, endpoint, log.Error),
-		verb:   rpc.Dial(ftlv1connect.NewVerbServiceClient, endpoint, log.Error),
+		runner:     rpc.Dial(ftlv1connect.NewRunnerServiceClient, endpoint, log.Error),
```
stuartwdouglas, line 1161: Does this actually create a new client in terms of the underlying connections? rcp.Dial seems to use a static HTTP client, and AFAIK go clients manage their own connection pools, so I think we may have an implicit connection cache.

stuartwdouglas: https://github.com/TBD54566975/ftl/blob/787444ee9d8ec79cec02f645dfaea2c6726cdcda/internal/rpc/rpc.go#L32 is the client I am talking about.

matt2e:
```
// The [Client.Transport] typically has internal state (cached TCP
// connections), so Clients should be reused instead of created as
// needed.
```
It seems so yes. But it also seems that having one http.Client is correct usage 🤔

matt2e:
Closing as we were able to reproduce the RBAC issue for an endpoint that had never been used before in the environment

PR #2135, opened 2024-07-23
fix: remove global HTTP clients

stuartwdouglas:
The global HTTP clients act as an implicit connection cache, which appears to cause issues in some kube/istio environments.


stuartwdouglas:
closing, as it appears this is not a caching issue.

PR #2136, opened 2024-07-23
fix: kubernetes builds native deploy

gak:


wesbillman (approved):
I'm no `kubectl` expert but this looks good to me :)

PR #2138, opened 2024-07-23
feat: extract type by node + add schema fuzz test

worstell:
this change also allows us to support `type Foo = external.Bar` aliases


go-runtime/schema/extract_test.go
```diff
@@ -0,0 +1,324 @@
...
+			}
+			key := getSchemaType(symbol[4:idx])
+			value := getSchemaType(symbol[idx+1:])
+			return &schema.Map{Key: key, Value: value}
+		}
+		if strings.HasPrefix(symbol, "ftl.Option[") {
+			return &schema.Optional{Type: getSchemaType(symbol[11 : len(symbol)-1])}
+		}
+		panic(fmt.Sprintf("unexpected symbol: %s", symbol))
+	}
```
wesbillman, line 76: I don't have a ton of context here, but this seems a bit brittle. Is this just how things have to work here with string checking?

go-runtime/schema/extract_test.go
```diff
@@ -0,0 +1,324 @@
...
+}
+
+func generateSourceCode(symbol string) string {
+	var codeBuilder strings.Builder
+
+	codeBuilder.WriteString("package test\n\n")
+	codeBuilder.WriteString(`import (` + "\n")
+	codeBuilder.WriteString(`	"context"` + "\n")
```
wesbillman, line 84: Would it be better/easier to use raw string literals here instead of this `codeBuilder`? It seems a bit hard to write and update like this

wesbillman (approved):

go-runtime/schema/extract_test.go
worstell: true! looks better

go-runtime/schema/extract_test.go
```diff
@@ -0,0 +1,324 @@
...
+  export verb exportedFunc(%s) %s
+
+  verb func(%s) %s
+
+  verb sinkFunc(%s) Unit
+
+  verb sourceFunc(Unit) %s
+}
+`, typename, typename, typename, typename, typename, typename, typename, typename, valueenum, typeenum, typename,
+			typename, typename, typename, typename, typename, typename, typename) // there simply must be a better way
```
deniseli, line 316: Oh I think you could use https://pkg.go.dev/text/template#Must here too!

go-runtime/schema/data/analyzer.go
```diff
@@ -56,57 +50,68 @@ func extractData(pass *analysis.Pass, pos token.Pos, named *types.Named) optiona
...
+					jsonFieldName = tagParts[0]
+				}
+
+				if jsonFieldName != "" {
+					metadata = append(metadata, &schema.MetadataAlias{
+						Pos:   common.GoPosToSchemaPos(pass.Fset, node.Pos()),
+						Kind:  schema.AliasKindJSON,
+						Alias: jsonFieldName,
+					})
+				}
```
deniseli, line 105: Could we move this into a helper function (`parseTag` or something like that)?

deniseli (approved):

go-runtime/schema/extract_test.go
worstell: cleaned up a little and added some comments to make it clearer, e.g. indexing after `4` because `map[` is 4 characters log

go-runtime/schema/extract_test.go
```diff
@@ -0,0 +1,324 @@
...
+}
+
+// generate all possible permutations for slices, maps and ftl.Option[...]
+func generatePermutations(symbols []string) []string {
+	permutations := append([]string{}, symbols...)
+	for _, symbol := range symbols {
+		permutations = append(permutations, "[]"+symbol)
+		permutations = append(permutations, "ftl.Option["+symbol+"]")
+		for _, nestedSymbol := range symbols {
+			permutations = append(permutations, fmt.Sprintf("map[%s]%s", symbol, nestedSymbol))
```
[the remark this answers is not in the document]
worstell: ah yeah i omitted because the performance becomes exponentially worse and it seemed untenable

go-runtime/schema/extract_test.go
```diff
@@ -0,0 +1,324 @@
...
+			return &schema.Map{Key: key, Value: value}
+		}
+		if strings.HasPrefix(symbol, "ftl.Option[") {
+			return &schema.Optional{Type: getSchemaType(symbol[11 : len(symbol)-1])}
+		}
+		panic(fmt.Sprintf("unexpected symbol: %s", symbol))
+	}
+}
+
+func generateSourceCode(symbol string) string {
```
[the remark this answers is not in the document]
worstell: good idea!

PR #2140, opened 2024-07-23
feat: initializes otel metric collection for ftl dev and introduces a couple shared otel metric attributes

jonathanj-square:


deniseli (approved):
Niiiiice

PR #2144, opened 2024-07-24
feat: otel traces + interceptor

safeer:
otel interceptor that propagates spans and logs FTL attributes


internal/rpc/otel_interceptor.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+}
+
+type instruments struct {
+	duration     metric.Int64Histogram
+	requestSize  metric.Int64Histogram
+	responseSize metric.Int64Histogram
+}
+
+func createInstruments(meter metric.Meter) (instruments, error) {
```
deniseli, line 138: It would be nice if we didn't lose _all_ the instruments whenever one of their instantiations fails. What do you think about refactoring the code here and above to look something like below? This would also have the helpful maintainability effect of parceling the computation of each metric value into its own function, so it's easier to figure out what logic is used for what. `WrapUnary` is somewhat long at the moment, and it's easy to imagine how it could grow longer.

```
func (i *otelInterceptor) WrapUnary(next connect.UnaryFunc) connect.UnaryFunc {
	return func(ctx context.Context, request connect.AnyRequest) (connect.AnyResponse, error) {
		...
		recordDuration(ctx, attributes, requestStartTime)
		recordRequestSize(...)
		recordResponseSize(...)
		return response, err
	}
}

func recordDuration(ctx context.Context, attributes []attribute.KeyValue, requestStartTime time.Time) {
	recordHist(ctx, attributes, histDetails{
		name: "duration",
		unit: "ms",
		desc: "...",
		val: time.Since(requestStartTime).Milliseconds(),
	})
}

type histDetails struct {
	name string
	unit string
	desc string
	val  int64
}

func recordHist(ctx context.Context, attributes []attribute.KeyValue, details histDetails) {
	logger := log.FromContext(ctx) // just log the error instead of exiting the function early
	hist, err := meter.Int64Histogram(details.name, metric.WithUnit(details.unit), metric.WithDescription(details.desc))
	if err != nil {
		logger.Errorf(...)
	}
	hist.Record(ctx, details.val, metric.WithAttributes(attributes...))
}
```

internal/rpc/otel_interceptor.go
```diff
@@ -0,0 +1,156 @@
...
+	"go.opentelemetry.io/otel/trace"
+	"google.golang.org/protobuf/proto"
+)
+
+const (
+	otelFtlRequestKey = "ftl.requestKey"
+	otelFtlVerbRef    = "ftl.verb.ref"
+	otelFtlVerbModule = "ftl.verb.module"
+	otelFtlEventName  = "ftl.message"
+)
```
deniseli, line 24: Niiiice! For reference, Jon's [PR from yesterday](https://github.com/TBD54566975/ftl/pull/2140) added a few attributes wrapped in functions. I really like your way of doing const keys here since these attributes all have type string, and this is simpler/shorter than doing functions here, but we should clearly define the pattern at some point so it's clear when to follow/use each. Possibly: for interceptor trace attributes, which are kind of their own thing, we follow this pattern (const keys), but for metrics scattered across many directories, we'll use the functions from the linked PR?

internal/rpc/otel_interceptor.go
safeer: Good idea!

internal/rpc/otel_interceptor.go
safeer: Ahh I didn't see Jon's go in. I don't mind using that function pattern at all, happy to switch it, the consistency might be nice

deniseli: We ended up changing his file back to consts so what you have here is perfect :D

internal/rpc/otel_interceptor.go
deniseli: We ended up deciding to panic on metric instantiation error after all to make sure we don't miss any errors in development, but the rest of this still holds :)

internal/rpc/otel_interceptor.go
```diff
@@ -0,0 +1,420 @@
...
+	"go.opentelemetry.io/otel/trace"
+	"google.golang.org/protobuf/proto"
+
+	"github.com/TBD54566975/ftl/internal/log"
+)
+
+const (
+	otelFtlRequestKey       = "ftl.requestKey"
+	otelFtlVerbRef          = "ftl.verb.ref"
+	otelFtlVerbModule       = "ftl.verb.module"
```
deniseli, line 29: nittiest of nits: Jon appended `Attribute` to each of these variable names so I started doing that too. Kinda wondered about just doing `Attr` instead of brevity but either works

internal/rpc/otel_interceptor.go
```diff
@@ -0,0 +1,420 @@
...
+}
+
+func getAttributes(ctx context.Context, rpcSystemKey string) []attribute.KeyValue {
+	logger := log.FromContext(ctx)
+	attributes := []attribute.KeyValue{
+		semconv.RPCSystemKey.String(rpcSystemKey),
+	}
+	requestKey, err := RequestKeyFromContext(ctx)
+	if err != nil {
+		logger.Debugf("failed to get request key: %s", err)
```
deniseli, line 66: Since request key is so valuable, do you think we should raise this log level to Warn or even Error? I'm not sure personally

internal/rpc/otel_interceptor.go
```diff
@@ -0,0 +1,420 @@
...
+			if msg, ok := request.Any().(proto.Message); ok {
+				requestSize = proto.Size(msg)
+			}
+		}
+
+		span.AddEvent(otelMessageEvent,
+			trace.WithAttributes(
+				requestSpan,
+				otelMessageIDKey.Int(1),
+				otelMessageSizeKey.Int(requestSize),
```
deniseli, line 114: How does this work when requestSize is undefined? I assume message size attr just won't show up, but figured I'd ask just in case it's wonky :)

internal/rpc/otel_interceptor.go
```diff
@@ -0,0 +1,420 @@
...
+		if request != nil {
+			if msg, ok := request.Any().(proto.Message); ok {
+				requestSize = proto.Size(msg)
+			}
+		}
+
+		span.AddEvent(otelMessageEvent,
+			trace.WithAttributes(
+				requestSpan,
+				otelMessageIDKey.Int(1),
```
deniseli, line 113: What's message ID for?

internal/rpc/otel_interceptor.go
```diff
@@ -0,0 +1,420 @@
...
+	}
+	requestSize, err := meter.Int64Histogram("request.size", metric.WithUnit("By"))
+	if err != nil {
+		panic(fmt.Errorf("failed to create request size metric: %w", err))
+	}
+	responseSize, err := meter.Int64Histogram("response.size", metric.WithUnit("By"))
+	if err != nil {
+		panic(fmt.Errorf("failed to create response size metric: %w", err))
+	}
+	requestsPerRPC, err := meter.Int64Histogram("requests_per_rpc", metric.WithUnit("1"))
```
deniseli, line 288: Did you find any docs on how unit is supposed to be set for counts? I wasn't sure if we should just leave it out. "1" makes sense intuitively but I'm mostly wondering if this is how we're supposed to do it for all the count metrics

internal/rpc/otel_interceptor.go
```diff
@@ -0,0 +1,420 @@
...
+		attributes := getAttributes(ctx, request.Peer().Protocol)
+		traceOpts := []trace.SpanStartOption{
+			trace.WithAttributes(attributes...),
+			trace.WithSpanKind(spanKind),
+		}
+		tracer := otel.GetTracerProvider().Tracer(request.Spec().Procedure)
+		ctx, span := tracer.Start(ctx, name, traceOpts...)
+		defer span.End()
+
+		var requestSize int
```
deniseli, line 103: Should this default to 0 if there's no request object? If not, we should probably mention in the description of this metric that it's only recorded for RPCs with a non-null request.

internal/rpc/otel_interceptor.go
```diff
@@ -0,0 +1,420 @@
...
+		duration:        duration,
+		requestSize:     requestSize,
+		responseSize:    responseSize,
+		requestsPerRPC:  requestsPerRPC,
+		responsesPerRPC: responsesPerRPC,
+	}
+}
+
+func statusCodeAttribute(protocol string, err error) attribute.KeyValue {
+	statusCodeKey := fmt.Sprintf("rpc.%s.status_code", protocol)
```
deniseli, line 306: Do we need to lowercase protocol, or is it always guaranteed to be consistent case?

deniseli (approved):
This is awesome :D

internal/rpc/otel_interceptor.go
safeer: The err case is only really hit if there is a request key that we're unable to parse. It is indicative of an issue though, so good call bumping the sev

internal/rpc/otel_interceptor.go
safeer: Doing this for consistency across the rpc interceptors. The individual messages are all shown in the span event, and their ID's must be unique, in the unary case there is one request and one response

internal/rpc/otel_interceptor.go
safeer: Fixed, good catch on defining requestSize :)

internal/rpc/otel_interceptor.go
safeer: Yup, docs say that for counts the unit should be `"1"` 🤷

internal/rpc/otel_interceptor.go
safeer: Using the raw value from the Connection and Spec, so for consistency I'm not lowering here, but in our use it's always just `"grpc"` anyway

backend/controller/controller.go
```diff
@@ -1022,6 +1022,7 @@ func (s *Service) callWithRequest(
 		}
 	}
 
+	ctx = rpc.WithRequestName(ctx, requestKey)
```
wesbillman, line 1025: nit: Should this be `rpc.WithRequestKey`?

internal/rpc/otel_interceptor.go
```diff
@@ -0,0 +1,420 @@
...
+	}...)
+	if protomsg, ok := msg.(proto.Message); ok {
+		size := proto.Size(protomsg)
+		attrs = append(attrs, otelMessageSizeAttr.Int(size))
+		s.sendSize.Record(ctx, int64(size), metric.WithAttributes(attrs...))
+	}
+
+	span := trace.SpanFromContext(ctx)
+	span.AddEvent(otelMessageEvent, trace.WithAttributes(attrs...))
+	return err // nolint:wrapcheck
```
wesbillman, line 384: no we want to `nolint` all these `err` or should we wrap them?

internal/rpc/otel_interceptor.go
wesbillman: That's so interesting. Seems like some sort of shorthand, but it is what they recommend :)

wesbillman (approved):

deniseli:

internal/rpc/otel_interceptor.go
safeer: Good question; the interceptor just logs errors as they pass through, and it didn't feel quite right to attach this context to the trace; all errors that flow through rpc would then have some `otel error:` string attached

Issue #2147, opened 2024-07-24
Reduce build times and frontend JS size by removing excess dependencies.


gak:
Currently it takes ~10s to build release on a M2, and 16.1 MB to serve in dev.

All FTL builds and most CI rely on this install.

gak:
Looks like this might be adding a lot of deps that we don't need: https://github.com/jsonnext/codemirror-json-schema/issues/125

gak:
- https://github.com/shikijs/shiki/issues/658

gak:
Removing codemirror deps and temporarily showing a blank div brings the build down to 3s and 1.7 MB.

wesbillman:
@gak I just tried this locally on this branch and on `main` and didn't notice a massive difference:
```
this-branch: just build-all  22.41s user 9.17s system 144% cpu 21.893 total
main: just build-all  28.29s user 9.80s system 139% cpu 27.295 total
```

Definitely reduces the size though. I'm ok with pulling this out for now, but we might have to clean up the new version of the form a bit since this might be the most used feature on the console 😂

wesbillman:
Also, without json-schema validation in the UI it might be really difficult for users to know what fields they need to send on a request. Maybe we need to solve that in another way, but there currently isn't a great way to see what should and should not be there. 🤔

gak:
I am bringing it back, just with less deps. It will at least save the size of the deploy at least.

wesbillman:
Nice! Sounds great 👍

wesbillman:
I think this was "fixed" with #2183

gak:
Ah of course, closing!

PR #2153, opened 2024-07-24
fix: temporarily remove heavy dependency in verb request code editor

gak:
Related to #2147

It was taking some time go patch the dependency, so I put this out in the mean time to reduce build times.


wesbillman:
Oh shoot! I responded on the issue instead of here 🤦 https://github.com/TBD54566975/ftl/issues/2147

gak:
Going to pause this change for now since it doesn't seem to affect CI that much and we have other plans like tree shaking. Plus it's a very useful feature to just take away!

PR #2154, opened 2024-07-24
feat: extract type by node + add schema fuzz test (#2138)

### LINKED ISSUE #2145: allow type aliases for ingress req/response if their value is builtin.HttpRequest/Response

worstell:
user requested support for `type Foo = other.Bar` pattern when declaring type aliases. this allows them to leverage member functions of the underlying type.

after this change, the above pattern as well as the prior pattern, `type Foo other.Bar` will be supported for type aliasing.

also fixes #2145


wesbillman (approved):
Thanks for fixing this up!! Looks awesome

PR #2155, opened 2024-07-24
feat: metric integration for fsm

jonathanj-square:


backend/controller/observability/attributes.go
```diff
@@ -0,0 +1,5 @@
+package observability
+
+const (
+	fsmRefAttribute string = "ftl.fsm.ref"
```
wesbillman, line 4: Given this is only used by `fsm.go` it could go there instead of this shared `attributes` file (unless there's plans to share it in the future).

backend/controller/observability/observability.go
```diff
@@ -0,0 +1,5 @@
+package observability
+
+func Init() {
+	InitFSMMetrics()
+}
```
wesbillman, line 5: Maybe this should be something like:

```go
func InitControllerObservability(ctx context.Context) {
    err := internalobs.Init(ctx, .....)
    // check errors
    err := InitFSMMetrics(...)
    // check errors
}
```

internal/observability/client.go
```diff
@@ -64,6 +65,8 @@ func Init(ctx context.Context, serviceName, serviceVersion string, config Config
 	meterProvider := metric.NewMeterProvider(metric.WithReader(metric.NewPeriodicReader(otelMetricExporter)), metric.WithResource(res))
 	otel.SetMeterProvider(meterProvider)
 
+	observability.Init()
+
```
wesbillman, line 69: The internal package should not depend on backend/controller/... things at all (or runner, etc.). This would create internal dependencies on `controller` and `runner`. Instead the `backend` stuff should use `internal` as needed.

internal/observability/metrics/attributes.go
```diff
@@ -1,18 +1,5 @@
 package metrics
```
wesbillman, line 1: I think this file should move up one level and be at the `internal/observability` level instead of `internal/observability/metrics` that way we can use shared attributes in traces as well.

backend/controller/observability/fsm.go
```diff
@@ -0,0 +1,38 @@
...
+
+var fsmMeter = otel.Meter("ftl.fsm")
+
+var fsmCounters = struct {
+	instancesActive metric.Int64UpDownCounter
+}{}
+
+// TODO error logging and handling
+func InitFSMMetrics() {
+	fsmCounters.instancesActive, _ = fsmMeter.Int64UpDownCounter(
```
wesbillman, line 23: `func InitFSMMetrics()` should return an error so it can be checked.

backend/controller/observability/attributes.go
wesbillman: If we don't currently have any shared `attributes` then we can leave this file off for now and add it back if needed in the future.

wesbillman (commented):
Nice! I think the structure is getting there!

backend/controller/observability/fsm.go
jonathanj-square: agreed; just going for a quicker turn around on other aspects of the PR by leaving that as a TODO comment, will return to it before exiting draft mode on this PR

backend/controller/observability/attributes.go
jonathanj-square: moved

backend/controller/observability/observability.go
jonathanj-square: adopted that name

internal/observability/client.go
jonathanj-square: moved to `controller.go:New`

internal/observability/metrics/attributes.go
jonathanj-square: moved

backend/controller/observability/fsm.go
jonathanj-square: added the error handling

backend/controller/controller.go
```diff
@@ -226,6 +227,10 @@ func New(ctx context.Context, pool *pgxpool.Pool, config Config, runnerScaling s
 	}
 	config.SetDefaults()
 
+	if err := observability.InitControllerObservability(); err != nil {
+		log.FromContext(ctx).Warnf("failed to initialize controller observability: %v", err)
+	}
```
wesbillman, line 232: This works for now, but we might want to consider combining this and  the `observability.Init()` that's currently in `ftl-controller/main` `ftl-runner/main` and `cmd_serve` files now. We can iterate on that in the future.

wesbillman (approved):
Nice! This structure looks good to me!

backend/controller/controller.go
jonathanj-square: will combine in a fast follow change that completes the fsm metric instrumentation

deniseli:

PR #2163, opened 2024-07-25
fix: improve lease log level and include key

### LINKED ISSUE #2159: Include lease name in log for when lease expires for a module
Current log message from the runner: `Lease heartbeat terminated: canceled: context canceled`
It would be helpful to debug this by knowing the actual name of the lease
Same for `Renewing lease` log messages from the controller

### LINKED ISSUE #2160: No need for leases to log the happy case: `Lease heartbeat terminated: canceled: context canceled`
It seems that when logs are obtained by module code and the released we log a warning that the lease heartbeat has been terminated due to a canceled context.

... or there may be a legit issue being exposed here. Need to research more.

matt2e:
fixes #2159
fixes #2160

- When module lease heartbeats stopped, it always logged a warning, even if the lease was released properly.
- All module lease logs now include the lease key for easy log lookups with controller lease logs
- Previously, `lease.err` was not set properly because the struct was not referenced via a pointer so the state was not shared.
    - I added `state: *leaseState` to get around this (rather than making the lease handle a pointer) because it also made clear what the mutex is protecting. Not a strong opinion though, happy to do the other approach if people feel that it's cleaner.

### DIFF
--- a/go-runtime/ftl/leases.go
+++ b/go-runtime/ftl/leases.go
@@ -46,7 +46,7 @@ func (l LeaseHandle) Err() error {
 	l.state.mutex.Lock()
 	defer l.state.mutex.Unlock()
 	if err, ok := l.state.err.Get(); ok {
-		return err
+		return err //nolint:wrapcheck
 	}
 	return nil
 }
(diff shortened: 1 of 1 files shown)


matt2e:
Accidentally included this change in this PR https://github.com/TBD54566975/ftl/pull/2165 🤦‍♂️🤦‍♂️🤦‍♂️

Issue #2026, opened 2024-07-09
otel: change the go-runtime SDK to use a constant string (or configurable?)


deniseli:
**_THIS IS LOOOOWWWW PRIORITY!_** (because it's both not in use and opt-in)

In [go-runtime/ftl/observability/metrics.go](https://github.com/TBD54566975/ftl/blob/main/go-runtime/ftl/observability/metrics.go#L17), we use the verb name as the metric name, which would result in a seeing a quite high number of metrics. This is not ideal; we don't want FTL to have a reputation for being expensive to instrument logging for.

As an alternative, users are free to invoke the `otel` SDK directly in their code, and the global `otel` configuration/setup we've done will apply.

- [x] Delete the file for now
- [ ] Redesign it later

deniseli:
@wesbillman yeah agreed, getting rid of the file for now seems like the right call.

deniseli:
Current thoughts: it'd be nice to make the go-runtime SDK observability interface more declarative than what otel forces you into. We can simply wrap otel for now and give the users better functions for interfacing with it. For any niche features that we can't support, there's always the backdoor of invoking otel directly in the user code.

### Design Thoughts Part 1: Attributes

It'd be nice to pass attributes via struct (so the types are built-in) instead directly using the `attribute` package that otel provides. Example:

Suppose you're a user, and one of your verbs calls out to a somewhat flaky external API. You want to instrument some metrics on your verb with attributes recording the details of that API call:
```
type MyVerbAttrs struct {
    failureMode ftl.Option[string] `attr:"failure_mode"`
    retryCount  int                `attr:"my_api_call.retry_count"`
}

func(...) (...) {
    ...
    ftl.RecordCount(ctx, ftl.Metric(...), MyVerbAttrs{...})
    ...
}
```

The Go SDK could use `reflect` to produce:
```
attrs := []attribute.KeyValue{
    attribute.String("ftl.user_defined.failure_mode", "my_failure_mode_1"), // only present when option.ok is true
    attribute.Int64("ftl.user_defined.my_api_call.retry_count.bucket", 0),
}
```

deniseli:
### Design Thoughts Part 2: Metric instantiation

Meters and metrics don't _have_ to be instantiated at initialization time. In fact, the user code will be cleaner if they don't need to think about that. The SDK could just provide a set of functions to create and manage metrics:

`ftl.CounterMetric[T](signalName string, unit, description ftl.Option[string]) (metric.Int64Counter, error)`

[...]

PR #2182, opened 2024-07-26
chore: delete observability package from Go SDK

deniseli:
Part 1 of https://github.com/TBD54566975/ftl/issues/2026

This deletes the whole observability package from the Go SDK, since no one is using it yet. We'll replace it with an interface that better aligns with our internal otel usage patterns once that's more mature.


wesbillman (approved):
Love it! We can add this stuff back in once we have a better idea of what (if anything) might be useful. In the meantime, customers can use `otel`s normal API to track all the metrics they want

PR #2201, opened 2024-07-30
feat: introduce runner metrics

jonathanj-square:
This change introduces the following runner metrics:

* `ftl.runner.startup.failures`
* `ftl.runner.registration.heartbeats`
* `ftl.runner.failures`

Metric output from local run:

---

ScopeMetrics #1
ScopeMetrics SchemaURL:
InstrumentationScope ftl.runner

Metric #0
Descriptor:
     -> Name: ftl.runner.registration.heartbeats
     -> Description: the number of successful runner (re-)registrations
     -> Unit:
     -> DataType: Sum
     -> IsMonotonic: true

NumberDataPoints #0
Data point attributes:
     -> ftl.deployment.key: Str(unknown)
     -> ftl.runner.state.name: Str(reserved)
StartTimestamp: 2024-07-30 23:15:10.131588 +0000 UTC
Timestamp: 2024-07-30 23:15:55.135067 +0000 UTC
Value: 4

NumberDataPoints #1
Data point attributes:
     -> ftl.deployment.key: Str(dpl-echo-32feoiv4rvtuqc66)
     -> ftl.runner.state.name: Str(assigned)
StartTimestamp: 2024-07-30 23:15:10.131588 +0000 UTC
Timestamp: 2024-07-30 23:15:55.135067 +0000 UTC
Value: 39

NumberDataPoints #2
Data point attributes:
     -> ftl.deployment.key: Str(dpl-time-2qb3im47gl8zgfb0)
     -> ftl.runner.state.name: Str(assigned)
StartTimestamp: 2024-07-30 23:15:10.131588 +0000 UTC
Timestamp: 2024-07-30 23:15:55.135067 +0000 UTC
Value: 39

NumberDataPoints #3
Data point attributes:
     -> ftl.deployment.key: Str(unknown)
     -> ftl.runner.state.name: Str(idle)
StartTimestamp: 2024-07-30 23:15:10.131588 +0000 UTC
Timestamp: 2024-07-30 23:15:55.135067 +0000 UTC
Value: 119


backend/runner/observability/runner.go
```diff
@@ -0,0 +1,95 @@
...
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+	"go.opentelemetry.io/otel/metric/noop"
+
+	ftlv1 "github.com/TBD54566975/ftl/backend/protos/xyz/block/ftl/v1"
+)
+
+const (
+	runnerMeterName              = "ftl.runner"
+	runnerDeploymentKeyAttribute = "ftl.deployment.key"
```
deniseli, line 19: Since this one isn't runner-specific, should we move it to internal/observability/attributes.go? We'll definitely want to use it for deploy/controller metrics too

backend/runner/observability/runner.go
```diff
@@ -0,0 +1,95 @@
...
+	if result.registrationHeartbeats, err = result.meter.Int64Counter(
+		counter,
+		metric.WithDescription("the number of successful runner (re-)registrations")); err != nil {
+		result.registrationHeartbeats, errs = handleInitErrors(counter, err, errs)
+	}
+
+	counter = fmt.Sprintf("%s.registration.failures", runnerMeterName)
+	if result.registrationFailures, err = result.meter.Int64Counter(
+		counter,
+		metric.WithDescription("the number of failures encounter while attempting to runner registration")); err != nil {
```
deniseli, line 55: nit: the number of failures encounter**ed** while attempting to **register a runner**

backend/runner/observability/runner.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	return result, errs
+}
+
+func (m *RunnerMetrics) Registered(ctx context.Context, key *string, state ftlv1.RunnerState) {
+	keyAttr := "unknown"
+	if key != nil {
+		keyAttr = *key
+	}
```
deniseli, line 66: We should ideally make `key` an `optional.Option` instead of using a pointer here. Same below. It's annoying because we do have `key` as a pointer already in the runner code, but going forward it's better to avoid pointers to represent nullability :/

deniseli (commented):
Thanks for cranking this out! I'll hold off on stamping approval until we're done talking through everyone's thoughts on the lint rule suppression

backend/runner/runner.go
```diff
@@ -420,17 +424,20 @@ func (s *Service) registrationLoop(ctx context.Context, send func(request *ftlv1
...
 
 	// Wait for the next heartbeat.
 	delay := s.config.HeartbeatPeriod + time.Duration(rand.Intn(int(s.config.HeartbeatJitter))) //nolint:gosec
 	logger.Tracef("Registered with Controller, next heartbeat in %s", delay)
+	observability.Runner.Registered(ctx, deploymentKey, state)
 	select {
 	case <-ctx.Done():
 		err = context.Cause(ctx)
 		s.registrationFailure.Store(optional.Some(err))
+		observability.Runner.RegistrationFailure(ctx, deploymentKey, state)
```
safeer, line 440: aside: seeing these make me wonder whether we should be pushing up the actual errors to DD as well

wesbillman: Yeah that has come up a couple times. TBH I'm not sure either. Maybe we should bring this up with the larger team and @bradleydwyer since he has quite a bit of experience here as well.

deniseli: +1, looking forward to hearing what everyone has to say

backend/runner/observability/runner.go
jonathanj-square: nice catch

backend/runner/observability/runner.go
jonathanj-square: applied

backend/runner/observability/runner.go
jonathanj-square: that looks much better, thanks for the suggestion

backend/runner/observability/runner.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+const (
+	runnerMeterName              = "ftl.runner"
+	runnerDeploymentKeyAttribute = "ftl.deployment.key"
+	runnerStateNameAttribute     = "ftl.runner.state.name"
+)
+
+type RunnerMetrics struct {
+	meter                  metric.Meter
```
wesbillman, line 24: Are we using this reference to the `meter`?

jonathanj-square: not outside of init so this has been moved to init as a local var

wesbillman (approved):

PR #2202, opened 2024-07-30
fix: otel rpc metrics also as span attributes

safeer:
also adds an array of sent and received message sizes (in addition to the unary fields) for the streaming interceptors

### DIFF
--- a/internal/rpc/otel_interceptor.go
+++ b/internal/rpc/otel_interceptor.go
@@ -193,6 +193,8 @@ func (i *otelInterceptor) WrapStreamingClient(next connect.StreamingClientFunc) 
 			attributes:        attributes,
 			receiveSizeMetric: instruments.responseSize,
 			sendSizeMetric:    instruments.requestSize,
+			receiveSizes:      []int64{},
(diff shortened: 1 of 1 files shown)


internal/rpc/otel_interceptor.go
```diff
@@ -24,15 +24,25 @@ import (
...
+	otelMessageEventIDAttr       = attribute.Key("message.id")
+	otelMessageEventSizeAttr     = attribute.Key("message.uncompressed_size")
+	otelMessageEventTypeAttr     = attribute.Key("message.type")
+	otelMessageEventTypeSent     = "SENT"
+	otelMessageEventTypeReceived = "RECEIVED"
+	otelMessageSentSizesAttr     = attribute.Key("rpc.message.sent.sizes_bytes")
+	otelMessageReceivedSizesAttr = attribute.Key("rpc.message.received.sizes_bytes")
+)
+
+const (
```
deniseli, line 40: nit: should we merge these `const` blocks?

deniseli (approved):

PR #2204, opened 2024-07-30
fix: just tidy works with fuzz test

worstell:


alecthomas (approved):
LGTM but maybe we should copy the files to a temp directory before building, similar to how the integration tests work?

PR #2207, opened 2024-07-30
feat: remove error message attribute

deniseli:

Remove error messages from attribute logging to avoid leaking private info into metrics


alecthomas (approved):

jonathanj-square (commented):
nice catch!

PR #2219, opened 2024-07-31
feat: log name of the calling verb to PublishEvent in otel metric attributes

deniseli:
Add `ftl.pubsub.publish.caller.verb.name` attribute to `ftl.pubsub.published` metric. Only the verb name, not module, is needed because the module will be the same as the topic's module, which is already logged (`ftl.module.name`)

Please let me know if there are DB changes I missed that should be included in this PR :)

```
Metric #0
Descriptor:
     -> Name: ftl.pubsub.published
     -> Unit: 1
     -> DataType: Sum
     -> IsMonotonic: true

NumberDataPoints #0
Data point attributes:
     -> ftl.module.name: Str(echo)
     -> ftl.pubsub.publish.caller.verb.name: Str(Echo)
     -> ftl.pubsub.topic.name: Str(echotopic)
StartTimestamp: 2024-07-31 17:28:47.451585 +0000 UTC
Timestamp: 2024-07-31 17:29:12.44603 +0000 UTC
Value: 1
```


jonathanj-square (approved):
If the calling verb is, by design, only allowed to come from the same module then this lgtm. 

When we add support for multiple modules running in a cluster I wonder if that relationship will hold

deniseli:
> If the calling verb is, by design, only allowed to come from the same module then this lgtm.
> 
> When we add support for multiple modules running in a cluster I wonder if that relationship will hold

Yup indeed, they are currently required to come from the same module: https://github.com/TBD54566975/ftl/blob/fbd2d1a48fa8c398cdb82392dace05c2237f3b2f/go-runtime/internal/impl.go#L73

deniseli:
This PR doesn't persist the caller, so if we want to retrieve that later, we'll need to add `caller` to the `topic_events` table

deniseli:
> This PR doesn't persist the caller, so if we want to retrieve that later, we'll need to add `caller` to the `topic_events` table. That would let us record the `caller` attribute on all the other pubsub metrics too, which would be awesome. I'll add a todo for myself

https://github.com/TBD54566975/ftl/pull/2231

PR #2237, opened 2024-08-01
fix: setting global config and secrets when using db config provider 

mistermoe:
# Problem
Setting global config when using db config provider allows creating the same config name over and over again. e.g.

```bash
❯ ftl config set KEY "OKAY1" --db

❯ ftl config get KEY 
"OKAY1"

❯ ftl config set KEY "OKAY2" --db

❯ ftlprod config get KEY
"OKAY1"
```

running `ftl config list` reveals the problem:
```bash
❯ ftlprod config list
KEY
KEY
```

This problem occurs because of the compound unique constraint [here](https://github.com/TBD54566975/ftl/blob/8dd18e78e9404832f378101e3cee6f1a8b514f66/backend/controller/sql/schema/20231103205514_init.sql#L518). 

In PostgreSQL, `NULL` values are treated as distinct, so when you have a unique constraint on multiple columns where one of the columns can be `NULL`, each `NULL` is treated as a different value. This means that multiple rows can have `NULL` in the module column and the same value in the name column without violating the unique constraint.

# Fix
This PR proposes treating a `NULL` module value as `''` when creating the unique compound constraint 

> [!NOTE]
> I believe the following would also work
> ```sql
> 
> ALTER TABLE module_configuration
>     ALTER COLUMN module SET DEFAULT '';
> ```

> [!NOTE]
> I went with `''` to prevent colliding with potential module names and i chose _not_ to default module to `''` to prevent application code from treating `''` as a not nil value though go does treat `''` as the zero value for a string but im not sure how that works with `optional.Option`.


mistermoe:
@deniseli i went ahead and refactored the DAL Configuration tests to isolate each individual test run. The way they were previously written led to side effects from previous tests causing subsequent test failures. I believe i've captured all of the test cases you previously had. definitely feel free to add any that i may have missed!

deniseli:
@mistermoe fantastic, thanks for doing that!

mistermoe:
@deniseli i went ahead and applied this fix for secrets as well because the same problem existed there as well

deniseli (approved):
:D Context: https://sq-tbd.slack.com/archives/C04P86U1B4N/p1722534396323009

wesbillman (approved):
Nice! Thanks @mistermoe !!!

wesbillman:
I wonder if it would make sense in a future PR to make the global vs. module more explicit instead of relying on this `COALESCE`? Maybe another column with an enum of `global` and `module` or something? Eventually it could be like `realm` maybe as well... 🤔

PR #2255, opened 2024-08-05
otel: add `project_name` and `is_user_service` attributes

### LINKED ISSUE #2023: otel: add attributes to all calls: `project_name` and `is_user_service`
`project_name`: from `ftl-project.toml`

`is_user_service`: We have the `serviceName` set [here](https://github.com/TBD54566975/ftl/blob/main/internal/observability/client.go#L50), which currently will be one of `ftl-controller`, `ftl-runner`, or `<module name>`. Technically, we could use `ftl-controller || ftl-runner` but it would be good to be a bit more explicit.
* To do this, add bool flag `IsInternal` to `observability.Config` so that `Init`’s callers can configure whether it’s user code or not.
* Naming: is_user_service, is_internal, or similar are all fine

safeer:
Fixes #2023 #1816

### DIFF
--- a/buildengine/testdata/type_registry_main.go
+++ b/buildengine/testdata/type_registry_main.go
@@ -35,7 +35,7 @@ func init() {
 }
 
 func main() {
-	verbConstructor := server.NewUserVerbServer("project", "other",
+	verbConstructor := server.NewUserVerbServer("ftl", "other",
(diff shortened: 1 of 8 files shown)


deniseli (approved):
awesome :D

PR #2256, opened 2024-08-05
feat: introduce cron job metrics

jonathanj-square:


backend/controller/cronjobs/cronjobs.go
```diff
@@ -260,20 +270,24 @@ func (s *Service) killOldJobs(ctx context.Context) (time.Duration, error) {
 		pattern, err := cron.Parse(stale.Schedule)
 		if err != nil {
 			logger.Errorf(err, "Could not kill stale cron job %q because schedule could not be parsed: %q", stale.Key, stale.Schedule)
+			observability.Cron.JobFailed(ctx, stale.Key, stale.DeploymentKey)
```
deniseli, line 273: Instead of bundling all the failures to kill a job with failures of the actual job itself, could we refactor this to the following?

- JobRun with attribute `.succeeded`
- JobKilled with attribute `.succeeded`

backend/controller/observability/cron.go
```diff
@@ -0,0 +1,106 @@
...
+		result.jobLatency, errs = handleInt64HistogramCounterError(counter, err, errs)
+	}
+
+	return result, errs
+}
+
+func (m *CronMetrics) JobExecutionStarted(ctx context.Context, job model.CronJobKey, deployment model.DeploymentKey) {
+	m.jobsActive.Add(ctx, 1, metric.WithAttributes(
+		attribute.String(observability.ModuleNameAttribute, job.Payload.Module),
+		attribute.String(cronJobRefAttribute, job.String()),
```
deniseli, line 71: Just confirming - is `job.String()` the ref to the cron job declaration?

backend/controller/observability/cron.go
```diff
@@ -0,0 +1,106 @@
...
+	m.jobsActive.Add(ctx, -1, metric.WithAttributes(
+		attribute.String(observability.ModuleNameAttribute, job.Payload.Module),
+		attribute.String(cronJobRefAttribute, job.String()),
+		attribute.String(observability.RunnerDeploymentKeyAttribute, deployment.String()),
+	))
+
+	m.jobLatency.Record(ctx, elapsed.Milliseconds(), metric.WithAttributes(
+		attribute.String(observability.ModuleNameAttribute, job.Payload.Module),
+		attribute.String(cronJobRefAttribute, job.String()),
+		attribute.String(observability.RunnerDeploymentKeyAttribute, deployment.String()),
```
deniseli, line 88: Since these attributes are identical to the ones above for `jobsActive`, would you mind deduplicating them into an `attrs` array? In fact, since they're duplicated across all these functions, could we make a helper function to construct `attrs` given `job, deployment`?

backend/controller/observability/cron.go
jonathanj-square: It's the module qualified verb name

deniseli: Ahhh in that case should we rename the attribute key to match?

backend/controller/cronjobs/cronjobs.go
jonathanj-square: I like the idea and would like to take it a little further. Removing the `jobFailures` counter does result in a loss of visibility that can be avoided if the `ftl.status.succeeded` attribute can take on a more generalized semantic. Changing from success/fail to "outcome" would help. Outcome is a string representing the canonical set of outcomes for a given domain. So in the metrics where the `ftl.status.succeed` is used today; outcome would have the values of `success` or `fail`. 

For cron jobs I would like outcomes covering the following: success, failed to start, execution failure, and killed. This would allow me to remove the `jobsKilled` metric and segment latency by outcome

jonathanj-square: updated

backend/controller/observability/cron.go
jonathanj-square: updated

backend/controller/observability/cron.go
jonathanj-square: updated

internal/observability/attributes.go
```diff
@@ -2,6 +2,16 @@ package observability
...
 const (
 	ModuleNameAttribute          = "ftl.module.name"
-	StatusSucceededAttribute     = "ftl.status.succeeded"
+	OutcomeStatusNameAttribute   = "ftl.outcome.status"
 	RunnerDeploymentKeyAttribute = "ftl.deployment.key"
+
+	SuccessStatus = "success"
+	FailureStatus = "failure"
 )
+
```
deniseli, line 11: What do you think about reprising the pattern you came up with ages ago and adding a helper function here that just returns `attribute.String(observability.OutcomeStatusNameAttribute, observability.SuccessOrFailureStatus(succeeded))`? Since so much of the code that uses this now is basically just duplicating that line

backend/controller/observability/cron.go
```diff
@@ -0,0 +1,108 @@
...
+
+func initCronMetrics() (*CronMetrics, error) {
+	result := &CronMetrics{}
+
+	var errs error
+	var err error
+
+	meter := otel.Meter(deploymentMeterName)
+
+	counter := fmt.Sprintf("%s.job.failures", cronMeterName)
```
deniseli, line 39: s/failures/completed/

backend/controller/observability/cron.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+func (m *CronMetrics) JobKilled(ctx context.Context, job model.CronJob) {
+	m.jobCompleted(ctx, job, cronJobKilledStatus)
+}
+
+func (m *CronMetrics) JobFailedStart(ctx context.Context, job model.CronJob) {
+	completionAttributes := cronAttributes(job, optional.Some(cronJobFailedStartStatus))
+
+	elapsed := time.Since(job.NextExecution)
```
deniseli, line 79: you can use `timeSinceMS` here and then you don't need the additional `.Milliseconds()` call below. Same below for every `time.Since`

deniseli (commented):
this is looking really nice :D

internal/observability/attributes.go
jonathanj-square: sgtm, applied

backend/controller/observability/cron.go
jonathanj-square: 😅 thanks for spotting that, updated

backend/controller/observability/cron.go
jonathanj-square: thanks, applied

internal/observability/attributes.go
deniseli: micro-nit: could we rename `SuccessOrFailureStatus` to `SuccessOrFailureStatusAttr` since that's what it returns?

deniseli (approved):
Niiiice can't wait to see this in DD :D

PR #2269, opened 2024-08-06
feat: bucketize latency attribute

deniseli:
This PR bucketizes the latency _attribute_ (NOT the latency _metric_! That would be very bad!)

For better filtering capabilities in datadog, we need to bucket our high-cardinality numeric attributes. This uses log base8 because:

1. That seems to give the most useful results for current production numbers we're seeing in DD - good functional granularity for analysis purposes without being so granular that it's annoying to sort through the buckets. Current values range from 62 to 20279, concentrated in the 100-1000 range.
2. Base 8 bucket labels are all powers of 2, so it's relatively easy for our brains to skim

Sample output:
```
-> ftl.async_call.time_since_scheduled_at_ms.bucket: Str([8,64))
```

### DIFF
--- a/backend/controller/observability/async_calls.go
+++ b/backend/controller/observability/async_calls.go
@@ -16,12 +16,12 @@ import (
 )
 
 const (
-	asyncCallMeterName                = "ftl.async_call"
-	asyncCallOriginAttr               = "ftl.async_call.origin"
-	asyncCallVerbRefAttr              = "ftl.async_call.verb.ref"
-	asyncCallTimeSinceScheduledAtAttr = "ftl.async_call.time_since_scheduled_at_ms"
-	asyncCallRemainingAttemptsAttr    = "ftl.async_call.remaining_attempts"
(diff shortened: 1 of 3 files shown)


wesbillman (approved):

alecthomas (commented):
Can we perform this bucketing in DD instead?

deniseli:
> Can we perform this bucketing in DD instead?

I couldn't find a way to do it. Oddly, DD has decent support for [advanced string filtering](https://docs.datadoghq.com/metrics/advanced-filtering/) and bools, but not even basics for numeric-type attrs. At least, I couldn't find any docs on it, and nothing intuitive I tried in the UI worked either. 

Worse, DD tries to show every single value of the attr in its filter dropdown, which fills up the dropdown to its max size and pushes all the other attrs out of view, so then you can't explore the values of any other attributes.

One option would be to reduce `base` to 4 or something, because we can add groups of allowed values for filtering... btw if anyone can find a good way to filter by `<` and `>` on int attrs in DD, I'd love to undo this

PR #2274, opened 2024-08-06
fix: s/ftl.async_call.ms_to_complete/ftl.call.ms_to_complete

deniseli:
whoops

```
Metric #1
Descriptor:
     -> Name: ftl.call.ms_to_complete
     -> Description: duration in ms to complete a verb call
     -> Unit: ms
     -> DataType: Histogram

HistogramDataPoints #0
Data point attributes:
     -> ftl.call.verb.ref: Str(echo.echo)
     -> ftl.module.name: Str(echo)
StartTimestamp: 2024-08-06 21:59:38.317129 +0000 UTC
Timestamp: 2024-08-06 22:00:33.317447 +0000 UTC
Count: 1
Sum: 17.000000
Min: 17.000000
Max: 17.000000
```

### DIFF
--- a/backend/controller/observability/calls.go
+++ b/backend/controller/observability/calls.go
@@ -43,7 +43,7 @@ func initCallMetrics() (*CallMetrics, error) {
 		return nil, wrapErr(signalName, err)
 	}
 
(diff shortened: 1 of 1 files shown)


safeer (approved):
😄

worstell (approved):

PR #2287, opened 2024-08-07
feat: propagate otel spans across pubsub

### LINKED ISSUE #2258: otel: ensure traces include span parenting for publish->subscription calls
Subscription span should be a child of the publisher's span. Check whether this is the case, and if not, create & propagate a new child span for the subscriber.

safeer:
This adds an `otel_context` column to both `topic_events` and `async_calls`.
- on publish, the current otel context is stored in `topic_events`
- `otel_context` is propagated to the `async_calls` row on dequeue and retries
- on execution of a row in `async_calls`, the current otel trace is replaced with `otel_context`

Pros:
- we can see related calls all in the same trace
- pattern can be reused for all async calls (i.e. fsm)

Cons:
- storing otel specific context in the db, feels a bit gross
- there is no root span tying the retries / subscriptions together, since the root would be the controller lifecycle. screenshot below.

An alternative would be to persist the `requestKey` in the tables instead of `otel_context` and attach that to subscribers / retries, and use that as a filter in DD, etc.

Screenshot 2024-08-07 at 12 59 43 PM

Closes #2258


backend/controller/sql/schema/20240807174508_pubsub_otel_context.sql
```diff
@@ -0,0 +1,10 @@
+-- migrate:up
+
+ALTER TABLE topic_events
+    ADD COLUMN otel_context JSONB;
+
+ALTER TABLE async_calls
+    ADD COLUMN otel_context JSONB;
```
wesbillman, line 7: I wonder if we should keep these generic like `trace_context` or maybe it's better to be `otel` specific in case we have another trace context in the future 🤷

safeer:
Updated to store the request key in `topic_events`, which becomes the parent request key in async calls, which are now also logged in the event table for use in the console. Also propagating the parent request key and logging it via otel for good measure.

backend/controller/controller.go
```diff
@@ -1382,6 +1390,26 @@ func (s *Service) executeAsyncCalls(ctx context.Context) (time.Duration, error)
...
+	// Extract the request key from the call and attach it as the parent request key
+	parentRequestKey := optional.None[model.RequestKey]()
+	if prk, ok := call.ParentRequestKey.Get(); ok {
+		if rk, err := model.ParseRequestKey(prk); err == nil {
+			parentRequestKey = optional.Some(rk)
+		} else {
+			logger.Tracef("Ignoring invalid request key: %s", prk)
+		}
+	}
+
```
wesbillman, line 1412: nit: Would it make sense to have this in `observability/async_calls` to keep it out of the controller? That would also remove the `otel` dependencies here.

wesbillman (approved):
Nice this is looking good!

PR #2301, opened 2024-08-08
fix: possible race in reset subscription

### LINKED ISSUE #2261: Possible race condition when resetting a subscription to head
Was noticed during PR review, but not sure how important it is:
https://github.com/TBD54566975/ftl/pull/2210

Possible case:
- Subscription is behind and currently consuming an event
- Reset subscription command comes in sets the state to idle and updates the cursor to head
- 2 new events are posted to the topic
- New event starts being consumed by the subscription
- Original event (from before the reset command) finishes being processed, sets the state to idle even though we are processing the other event still
- Then pub sub triggers the next event to be processed (without knowing the other event is being processed still)

worstell:
only update cursor, leave state as-is

fixes #2261


matt2e (approved):
Nice!

lsp/hoveritems.go
```diff
@@ -5,7 +5,7 @@ var hoverMap = map[string]string{
...
-	"//ftl:retry": "## Retries\n\nAny verb called asynchronously (specifically, PubSub subscribers and FSM states), may optionally specify a basic exponential backoff retry policy via a Go comment directive. The directive has the following syntax:\n\n```go\n//ftl:retry [<attempts>] <min-backoff> [<max-backoff>]\n```\n\n`attempts` and `max-backoff` default to unlimited if not specified.\n\nFor example, the following function will retry up to 10 times, with a delay of 5s, 10s, 20s, 40s, 60s, 60s, etc.\n\n```go\n//ftl:retry 10 5s 1m\nfunc Invoiced(ctx context.Context, in Invoice) error {\n  // ...\n}\n```\n",
```
matt2e, line 8: My fault 🤦‍♂️

PR #2310, opened 2024-08-09
feat: otel metrics for ingress

safeer:
```
InstrumentationScope ftl.ingress 

Metric #0
Descriptor:
     -> Name: ftl.ingress.requests
     -> Description: the number of ingress requests that the FTL controller receives
     -> Unit: 1
     -> DataType: Sum
     -> IsMonotonic: true

NumberDataPoints #0
Data point attributes:
     -> ftl.ingress.method: Str(GET)
     -> ftl.ingress.path: Str(/http/echo)
     -> ftl.ingress.run_time_ms.bucket: Str([16,32))
     -> ftl.ingress.verb.ref: Str(echo.getEcho)
     -> ftl.module.name: Str(echo)
StartTimestamp: 2024-08-12 17:44:51.923107 +0000 UTC
Timestamp: 2024-08-12 17:45:26.923594 +0000 UTC
Value: 1

Metric #1
Descriptor:
     -> Name: ftl.ingress.ms_to_complete
     -> Description: duration in ms to complete an ingress request
     -> Unit: ms
     -> DataType: Histogram

HistogramDataPoints #0
Data point attributes:
     -> ftl.ingress.method: Str(GET)
     -> ftl.ingress.path: Str(/http/echo)
     -> ftl.ingress.verb.ref: Str(echo.getEcho)
     -> ftl.module.name: Str(echo)
StartTimestamp: 2024-08-12 17:44:51.92311 +0000 UTC
Timestamp: 2024-08-12 17:45:26.923622 +0000 UTC
Count: 1
Sum: 27.000000
Min: 27.000000
Max: 27.000000
```

### DIFF
--- a/backend/controller/observability/ingress.go
+++ b/backend/controller/observability/ingress.go
@@ -1,0 +1,76 @@
+package observability
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+	"go.opentelemetry.io/otel/metric/noop"
+
+	schemapb "github.com/TBD54566975/ftl/backend/protos/xyz/block/ftl/v1/schema"
+	"github.com/TBD54566975/ftl/backend/schema"
+	"github.com/TBD54566975/ftl/internal/observability"
+	"github.com/alecthomas/types/optional"
+)
+
+const (
+	ingressMeterName         = "ftl.ingress"
+	ingressMethodAttr        = "ftl.ingress.method"
+	ingressPathAttr          = "ftl.ingress.path"
+	ingressVerbRefAttr       = "ftl.ingress.verb.ref"
+	ingressFailureModeAttr   = "ftl.ingress.failure_mode"
+	ingressRunTimeBucketAttr = "ftl.ingress.run_time_ms.bucket"
+)
+
+type IngressMetrics struct {
+	requests     metric.Int64Counter
+	msToComplete metric.Int64Histogram
+}
+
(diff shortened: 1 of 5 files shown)


backend/controller/observability/ingress.go
```diff
@@ -0,0 +1,76 @@
...
+
+	schemapb "github.com/TBD54566975/ftl/backend/protos/xyz/block/ftl/v1/schema"
+	"github.com/TBD54566975/ftl/backend/schema"
+	"github.com/TBD54566975/ftl/internal/observability"
+)
+
+const (
+	ingressMeterName         = "ftl.ingress"
+	ingressMethodAttr        = "ftl.ingress.method"
+	ingressPathAttr          = "ftl.ingress.path"
```
safeer, line 22: I'm curious if this is too high of a cardinality to log

deniseli: Depending on what info is contained in the paths, we could potentially chop the path and keep just the top level prefix, or otherwise parse out only the data that we find useful

deniseli: Oh huh if the path is just the module name, I guess it doesn't even matter if we do anything special here, because it won't add any cardinality beyond the verb ref. We can just leave this as is - it's probably easiest to understand this way

safeer: The path is entirely user defined, i.e. `//ftl:ingress GET /http/echo` -> `https://<domain:port>/http/echo`

But you're right it's equivalent to adding a verb ref, and it's pretty usefull; I'll leave it in.

deniseli: Ahh got it, that makes sense. Sounds good!

deniseli (approved):

PR #2331, opened 2024-08-12
chore: update versions in files

stuartwdouglas:
Running just build-all will cause these changes, commiting them so that building does not modify source controlled files.

Also add a check for this so we won't get into this state in future.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -168,6 +168,15 @@ jobs:
         run: just init-db
       - name: Rebuild All
         run: just build-all
+      - name: Check No SCM Changes
+        run: |
+          if [[ -n $(git status -s) ]]
+          then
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -168,6 +168,14 @@ jobs:
...
       - name: Rebuild All
         run: just build-all
+      - name: Check No SCM Changes
+        run: |
+          if [[ -n $(git status -s) ]]
+          then
+            echo "Running the build resulted in changes to git controlled files:"
+            git --no-pager diff
+            exit 1
+          fi
```
wesbillman, line 178: Nice! This is great. I also notice that `lsp/hoveritems.go` changes pretty often when I `just clean` and `just build-all` locally, so hopefully this will catch that case as well 👍

wesbillman (approved):

Issue #2266, opened 2024-08-06
ftltest race condition when waiting for pubsub to complete


matt2e:
from @mistermoe 
Steps to repro:
- Have 2 topics, each with a subscription
- Create a test that does the following:
    - the subscriber for the first topic publishes to the second topic
    - publish to the first topic
    - calls `ftltest.WaitForSubscriptionsToComplete(ctx)`

This is the error that can be seen sometimes: `panic: sync: WaitGroup is reused before previous Wait has returned`
Running the test with `--race` gives more info.

See branch `matt2e/ftltest-pubsub-race-condition` which already has a failing test set up for this

matt2e:
partner project has skipped their test in the meantime

PR #2332, opened 2024-08-12
fix: addresses event completion race condition in pubsub test utility 

jonathanj-square:
fixes: #2266

The race conditions was caused by two issues.

1. The WaitGroup down tick was not synchronized with the completion of the subscriber verb execution. Failure to wait for completion may result in a down time coming before a new event is dispatched (which ticks up the waitgroup) - which may result in the WaitGroup reaching zero too soon.
2.  Non-linear pubsub networks are not supported by simply counting up when an event is published and down after it is completed. The up count needs to match the number of live subscriptions (e.g. subscriptions with registered subscribers) 

This synchronization scheme does not cover asynchronous event dispatch (e.g. events dispatched via go routines within the subscriber verb) such scenarios introduce a race condition that cannot be resolved by black box external synchronization.


go-runtime/ftl/ftltest/testdata/go/pubsub/pubsub_test.go
safeer, line 1: Nice

safeer (approved):
Looks great!

PR #2339, opened 2024-08-13
feat: initial java <-> go integration test

stuartwdouglas:
Initial PR that tests simple interop. This will be expanded to cover as many different types and features as possible.


alecthomas (approved):

stuartwdouglas:
Ugh, that failure is because the Java artifacts have not been built before running the test, and it is using old cached ones from a previous PR. I need to have a think about how to handle this.

stuartwdouglas:
For now I am just running the java build the same way the binary is built. It might kinda suck when running from the IDE though needing to do the build each time, at some point we may want an env var or something.

PR #2368, opened 2024-08-14
fix: possible race on artefact upload

stuartwdouglas:
fixes: #2367


stuartwdouglas:
You mean to maintain an updated list of new artifacts in case there are multiple failures? I thought about I was not sure if it was worth extra complexity.

PR #2377, opened 2024-08-15
feat: add directive to enable lenient json

### LINKED ISSUE #2132: Be strict with ingressed model objects, optionally
Relates to the fix for #2117 and #2119 

We're currently lenient on inbound objects, which unblocks the client team, but they'd prefer to have it be optional. I.e. Jackson has: `@JsonIgnoreProperties(ignoreUnknown = true)`

worstell:
fixes #2132

adds support for `ftl:encoding` directive. the default `ftl:encoding` type is json and presently supports only the `lenient` option.

`//ftl:encoding lenient` or `//ftl:encoding json lenient` (which are equivalent) can annotate an ingress verb. when an ingress verb is annotated with this, it will accept requests with unknown/extra fields and ignore them.


worstell:
`just tidy` did a number so i recommend filtering out `go.mod/go.sum` 😅

alecthomas (approved):

matt2e (commented):
This is great!

PR #2401, opened 2024-08-16
fix: handle custom local names for widened types

### LINKED ISSUE #2397: typealiasing an external type breaks for imports with a local name
Steps to reproduce:
- Import an external package but give it a local name
- Create an FTL type alias that uses a type from this external package

Expected result:
- type alias of the widened external type works correctly

Actual result:
```
unsupported external type "[PATH]"; see FTL docs on using external types: tbd54566975.github.io/ftl/docs/reference/externaltypes/
            [PATH]: unsupported type "[VAR NAME]"
```

Some digging shows that the problem starts around here: https://github.com/TBD54566975/ftl/blob/main/go-runtime/schema/typealias/analyzer.go#L54

matt2e:
fixes #2397

`pass.Pkg.Imports()` did not always* include the local name alias from the import statement, so instead we need to do our checks via `PkgName` which does include it.

*it includes it in the example in our tests, but didn't in the case that our partner project was hitting...
I've added another test case that previously failed like theirs did.

### DIFF
--- a/go-runtime/schema/schema_test.go
+++ b/go-runtime/schema/schema_test.go
@@ -194,6 +194,9 @@ func TestExtractModuleSchemaTwo(t *testing.T) {
 	}
 	actual := schema.Normalise(r.Module)
 	expected := `module two {
+			typealias BackoffAlias Any
+        		+typemap go "github.com/jpillora/backoff.Backoff"
+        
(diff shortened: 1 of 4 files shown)


stuartwdouglas (approved):

alecthomas (approved):
Ship it!

PR #2434, opened 2024-08-20
fix: console verb page loading

wesbillman:
Less broke now! 😂 

I think with streaming queries, `isLoading` will always be true with tanstack since it's like streaming yo.

Screenshot 2024-08-19 at 5 22 22 PM


stuartwdouglas (approved):

deniseli (approved):
noice

PR #2467, opened 2024-08-21
chore: refactor JVM code

stuartwdouglas:
This moves the dicovery code for various API constructs into their own classes, and makes the whole process more composable.


wesbillman (approved):
LGTM although I'm not an expert in this area :)

Issue #2458, opened 2024-08-20
Select code editor component and finalize UX


wesbillman:
We are currently using [Codemirror](https://codemirror.net/) for our editor. 

In the past we have used [Monaco](https://microsoft.github.io/monaco-editor/), which was great and maybe the best option for us, but had some layout/resizing issues that need to be solved.

Another great option is [Ace](https://ace.c9.io/) editor.

A good example experience we should use as inspiration is the [Rust Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021). The layout and resizing is seamless and I really like the way the output window appears after a request is sent. 

- [x] Evaluate and select a text editor component
- [x] Integrate and implement necessary features
  - Json Schema Validation
  - Syntax highlighting
  - Light/Dark mode switching
  - etc.
- [x] Layout
  - Show output only if there is something to show
  - Make output resizeable

deniseli:
* Let's stick with codemirror6. It's a close race between that and monaco, but codemirror, esp v6, is way lighter weight and plays better with a modern js ecosystem. I messed around with fixing a few bugs to test out the dev experience and it's solid enough.
* Core features seem pretty solid already, but I'm putting up a final PR soon to handle localStorage better so that the request box gets populated with the default request whenever the verb schema changes.
* Layout was fixed in the PR above (#2472)

PR #2472, opened 2024-08-21
fix: clean up codemirror panel resizing

deniseli:
Issue: https://github.com/TBD54566975/ftl/issues/2458

Changes:
* Response panel is only displayed when `Body` tab (i.e. Request tab) is selected, instead of taking screen real estate away from the read-only schema tabs
* Border between request and response panels is replaced with a full-width draggable handle
    * Related: fixed the existing bugs in `components/ResizeableVerticalPanels.tsx`
* All codemirror components automatically take the full height of their container

https://github.com/user-attachments/assets/9594c696-1ddb-4412-bd13-722c73cd7662


wesbillman (approved):

wesbillman:
Nice! This is much better! 🥳

PR #2477, opened 2024-08-21
feat: hitting tab key inserts a tab in codemirror

deniseli:
Fixes https://github.com/TBD54566975/ftl/issues/2458

Before, tabbing would change the focus to the next element on the page. This adds the correct key binding so that hitting tab within the codemirror editor actually inserts 2 spaces.


wesbillman (approved):
Woot!

PR #2502, opened 2024-08-26
fix: avoid duplicate import aliases

### LINKED ISSUE #2469: declaring a typealias using a type from a package whose name is identical to the module name causes build errors
## Issue

declaring the following type alias:

```go
import "github.com/TBD54566975/tbdex-go/tbdex"

//ftl:typealias export
type TbdexMessage = tbdex.Message
```

and then using it in the `tbdex` module causes the module's codegen'd `main.go` to include two imports of the same name e.g.

```
package main

import (
	"context"

	"github.com/TBD54566975/tbdex-go/tbdex"

	"github.com/TBD54566975/ftl/backend/protos/xyz/block/ftl/v1/ftlv1connect"
	"github.com/TBD54566975/ftl/common/plugin"
	"github.com/TBD54566975/ftl/go-runtime/ftl/reflection"
	"github.com/TBD54566975/ftl/go-runtime/server"

	"ftl/tbdex"
)
```

running `ftl build` produces the following error:

```
info:tbdex: # main
./main.go:30:2: tbdex redeclared in this block
	./main.go:23:2: other declaration of tbdex
./main.go:65:27: too many errors
```

renaming the import in the source file like so:

```go
import tbdexsdk "github.com/TBD54566975/tbdex-go/tbdex"

//ftl:typealias export
type TbdexMessage = tbdexsdk.Message
```

produces a different error:

```
unsupported external type "github.com/TBD54566975/tbdex-go/tbdex.Message"; see FTL docs on using external types: tbd54566975.github.io/ftl/docs/reference/externaltypes/
```

## Potential Solution

when code-gen'ing a module's `main.go` i think it might be helpful to generate package aliases using a pattern that has a very low likelihood of someone actually using. e.g. `__tbdex "github.com/TBD54566975/tbdex-go/tbdex"`

matt2e:
fixes #2469
When scaffolding imports for external go packages, FTL will use the shortest unique part of the path as the alias. This allows package aliases to have neat and expected aliases in simple cases, and falling back to the next best option as conflicts arise.

For example:
```
import (
    unique "github.com/two2/foo/unique.Type"
    one_foo_bar "github.com/one/foo/bar.Type"
    two_foo_bar "github.com/two/foo/bar.Type"
)
```


alecthomas (approved):

go-runtime/compile/build.go
```diff
@@ -978,23 +975,128 @@ func goVerbFromQualifiedName(qualifiedName string) (goVerb, error) {
 	}, nil
 }
 
-// package and directory names are the same (dir=bar, pkg=bar): "github.com/foo/bar.A" => "github.com/foo/bar"
-// package and directory names differ (dir=bar, pkg=baz): "github.com/foo/bar.baz.A" => "baz github.com/foo/bar"
-func goImportFromQualifiedName(qualifiedName string) (string, error) {
+// returns the import path and directory name for an external type
+// package and directory names are the same (dir=bar, pkg=bar): "github.com/foo/bar.A" => "github.com/foo/bar", none
+// package and directory names differ (dir=bar, pkg=baz): "github.com/foo/bar.baz.A" => "github.com/foo/bar", "baz"
+func goImportFromQualifiedName(qualifiedName string) (importPath string, directoryName optional.Option[string], err error) {
```
gak, line 981: Nice, named return signature helps heaps with readability in this case.

go-runtime/compile/build.go
```diff
@@ -978,23 +975,128 @@ func goVerbFromQualifiedName(qualifiedName string) (goVerb, error) {
...
+		case *schema.Optional, *schema.Unit:
+			imports["github.com/TBD54566975/ftl/go-runtime/ftl"] = ""
+
+		case *schema.Topic:
+			if n.IsExported() {
+				imports["github.com/TBD54566975/ftl/go-runtime/ftl"] = ""
+			}
+
+		case *schema.TypeAlias:
+			if !aliasesMustBeExported || n.IsExported() {
```
gak, line 1032: I am not sure about the context here, so I'm probably wrong here, but thought I'd flag it in case.

Reading this if statement makes me think this block is for exported aliases, but its checking for when aliases must not be exported? Should it be `aliasesMustBeExported || n.IsExported()`?

gak (approved):
👍

go-runtime/compile/build.go
matt2e: I think what I want is:
```
if aliasesMustBeExported && !n.IsExported() {
  // early exit
}
```
When i wrote the original line, it also struck me as odd / confusing, so I'll replace it with the above as it's a lot clearer

PR #2505, opened 2024-08-26
chore: migrate deploy_test to integration tests in buildengine

jvmakine:
This tests the same things as the old test, but through slightly higher abstraction layer (cli instead of calling build engine functions directly)

The same functions as in the original test should get called via the cli in the integration test.

See https://github.com/TBD54566975/ftl/issues/1585


internal/buildengine/deploy_integration_test.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"testing"
+
+	in "github.com/TBD54566975/ftl/internal/integration"
+)
+
+func TestDeploy(t *testing.T) {
+	in.Run(t,
+		in.WithTestDataDir("testdata"),
```
jvmakine, line 13: We can remove this once all tests are integrated, and we can move the modules under corresponding language dirs.

alecthomas (approved):
Nice nice, that is much cleaner.

Issue #2487, opened 2024-08-23
Finish up tree view UI and logic


wesbillman:
- [x] Validate the use of `pullSchema` for the backend driving this UI
- [x] Handle open and collapsed states correctly
- [x] Handle selected items (show them selected in the tree and show a child page for them)
- [x] Finish up look and feel
- [x] Refactor components 
- [x] Make it expandable (resize width)
- [x] Make the module itself tappable. Maybe a button like `encore` has?
![Screenshot 2024-08-23 at 1 32 02 PM](https://github.com/user-attachments/assets/a04913a2-08d6-4e59-9820-020e7538e6e8)

PR #2523, opened 2024-08-28
feat: add module tree with routing to basic decl pages

deniseli:
Fixes https://github.com/TBD54566975/ftl/issues/2487

* New tree view
* Local storage persists the module expansion state and the width of the tree
* Base page added for all decl types, with slightly more built out for verb and data to show how it all hooks together

https://github.com/user-attachments/assets/36b9fb48-76ac-4fc7-aa34-54713675e462

Next: verb page code editor


wesbillman:
@deniseli you should be able to set up your editor to automatically lint/format on save so you don't get these type of CI errors.

```
./src/providers/routing-provider.tsx organizeImports ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  × Import statements could be sorted:
  
     4  4 │   import { DeploymentsPage } from '../features/deployments/DeploymentsPage'
     5  5 │   import { InfrastructurePage } from '../features/infrastructure/InfrastructurePage'
     6    │ - import·{·ModulesPage,·ModulesPanel·}·from·'../features/modules/ModulesPage'
     7    │ - import·{·ModulePanel·}·from·'../features/modules/ModulePanel'
        6 │ + import·{·ModulePanel·}·from·'../features/modules/ModulePanel'
        7 │ + import·{·ModulesPage,·ModulesPanel·}·from·'../features/modules/ModulesPage'
     8  8 │   import { DeclPanel } from '../features/modules/decls/DeclPanel'
     9  9 │   import { TimelinePage } from '../features/timeline/TimelinePage'
```

frontend/src/features/modules/ModulePanel.tsx
```diff
@@ -0,0 +1,12 @@
...
+
+export const ModulePanel = () => {
+  const { moduleName } = useParams()
+
+  const body = (
+    <div className='flex-1 py-2 px-4'>
+      <p>Module: {moduleName}</p>
+    </div>
+  )
+  return body
```
wesbillman, line 11: We can probably just return the body directly here instead of creating a new `const body`

frontend/src/features/modules/ModulesTree.tsx
```diff
@@ -1,60 +1,152 @@
...
+const DeclNode = ({ decl, href }: { decl: Decl; href: string }) => {
+  if (!decl.value || !decl.value.case || !decl.value.value) {
+    return []
+  }
+  const Icon = icons[decl.value.case] || CodeBracketSquareIcon
+  return (
+    <li className='my-1'>
+      <DisclosureButton
+        as='a'
+        href={href}
```
wesbillman: We should probably try to avoid using `a` here and use `{Link}` instead. This should help react router not have to reload the whole page and instead just the part that is changing.

frontend/src/features/modules/ModulesTree.tsx
```diff
@@ -1,60 +1,152 @@
...
+  topic: BellIcon,
+  typeAlias: DocumentDuplicateIcon,
+  secret: LockClosedIcon,
+  subscription: BellAlertIcon,
+  verb: BoltIcon,
+}
+
+type WithExport = { export?: boolean }
+
+const DeclNode = ({ decl, href }: { decl: Decl; href: string }) => {
```
wesbillman, line 42: Should we `memo` this to avoid rebuilds?

frontend/src/features/modules/ModulesTree.tsx
```diff
@@ -1,60 +1,152 @@
...
+      >
+        <Icon aria-hidden='true' className='size-4 shrink-0' />
+        {decl.value.value.name}
+        {(decl.value.value as WithExport).export === true ? <ExportBadge /> : []}
+      </DisclosureButton>
+    </li>
+  )
+}
+
+const ModuleSection = ({
```
wesbillman, line 65: Maybe `memo` this as well?

frontend/src/features/modules/ModulesTree.tsx
```diff
@@ -1,60 +1,152 @@
...
+      )}
+    </li>
+  )
+}
+
 export const ModulesTree = ({ modules }: { modules: ModuleTreeItem[] }) => {
+  const [searchParams, setSearchParams] = useSearchParams()
+  modules.sort((m1, m2) => Number(m1.isBuiltin) - Number(m2.isBuiltin))
+
+  const expandedModules = (searchParams.get('tree_m') || '').split(',')
```
wesbillman, line 114: I think this behavior is a bit different from what I had in mind. I don't think we need to be able to share the tree with a set of expanded modules. Instead, I think we should share the tree with the `selected` node. That way when I share a `url` with you, it will load the tree with that module selected (and show it selected) and show the detail page associated with that node.

frontend/src/features/modules/ModulesTree.tsx
deniseli: just finished refactoring this whole area :)

frontend/src/features/modules/ModulesTree.tsx
deniseli: Ahhhh ok that makes more sense. I was thinking this is pretty nonstandard. That makes this a lot simpler.

deniseli: replaced w localStorage

frontend/src/features/modules/ModulesTree.tsx
deniseli: Changed a ton of stuff around here but I think I memoized everything that doesn't need to be rerendered

frontend/src/features/modules/ModulesPage.tsx
```diff
@@ -1,21 +1,34 @@
...
+    <div className='flex-1 py-2 px-4'>
+      <p>Content</p>
+    </div>
+  )
+}
+
+export const ModulesPage = ({ body }: { body: React.ReactNode }) => {
   const schema = useSchema()
   const tree = useMemo(() => moduleTreeFromSchema(schema?.data || []), [schema?.data])
+  const [treeWidth, setTreeWidth] = useState(Number(localStorage.getItem('tree_w')) || 300)
```
wesbillman, line 19: nit: but we could make this a `const` so we don't typo in the future 😂 

Also, maybe call this `tree_width` just to be more clear?

frontend/src/features/modules/ModulesTree.tsx
```diff
@@ -1,60 +1,134 @@
...
+  ChevronRightIcon,
+  CircleStackIcon,
+  CodeBracketSquareIcon,
+  Cog6ToothIcon,
+  DocumentDuplicateIcon,
+  LockClosedIcon,
+  NumberedListIcon,
+  SquaresPlusIcon,
+} from '@heroicons/react/24/outline'
+import { TableCellsIcon } from '@heroicons/react/24/solid'
```
wesbillman, line 17: nit: Should we use `24/outline` for this one as well to be consistent? Also we can ignore this for now if that's easier since we'll probably use different icons soon.

frontend/src/features/modules/decls/VerbPanel.tsx
```diff
@@ -0,0 +1,67 @@
+import { useNavigate } from 'react-router-dom'
+import { Badge } from '../../../components/Badge'
+import type { Ref, Type, Verb } from '../../../protos/xyz/block/ftl/v1/schema/schema_pb'
+
+const DataRef = ({ heading, r }: { heading: string; r: Ref }) => {
```
wesbillman, line 5: ubernit: we can use `ref` instead of `r`. It took me a sec to know what `r` was here 🤦

frontend/src/features/modules/decls/VerbPanel.tsx
```diff
@@ -0,0 +1,67 @@
...
+    >
+      <div className='text-sm'>{heading}</div>
+      <span className='text-xs'>
+        {r.module}.{r.name}
+      </span>
+    </div>
+  )
+}
+
+const ioBlockClassName =
```
wesbillman, line 22: what is `ioBlock`? 

Ahh I see now, it's for showing request or response info stuff.

frontend/src/providers/routing-provider.tsx
```diff
@@ -15,7 +17,9 @@ const router = createBrowserRouter(
       <Route path='/' element={<Layout />}>
         <Route index element={<Navigate to='events' replace />} />
         <Route path='events' element={<TimelinePage />} />
-        <Route path='modules' element={<ModulesPage />} />
+        <Route path='modules' element={<ModulesPage body={<ModulesPanel />} />} />
+        <Route path='modules/:moduleName' element={<ModulesPage body={<ModulePanel />} />} />
+        <Route path='modules/:moduleName/:declCase/:declName' element={<ModulesPage body={<DeclPanel />} />} />
```
wesbillman, line 22: Nice!

frontend/src/features/verbs/VerbRequestForm.tsx
```diff
@@ -1,6 +1,6 @@
 import { useEffect, useState } from 'react'
 import { CodeEditor, type InitialState } from '../../components/CodeEditor'
-import { ResizableVerticalPanels } from '../../components/ResizeableVerticalPanels'
+import { ResizableVerticalPanels } from '../../components/ResizableVerticalPanels'
```
wesbillman, line 3: thanks for catching this! Words are hard

wesbillman (approved):

frontend/src/features/modules/ModulesPage.tsx
deniseli: whoops how embarassing, thanks for reminding me! 😂

frontend/src/features/modules/ModulesTree.tsx
deniseli: nice catch, yeah the outline version blends in a little better

frontend/src/features/modules/decls/VerbPanel.tsx
deniseli: Looks like `ref` is a reserved prop name in React because it gets handled differently and ends up being undefined whenever you reference it :(((

frontend/src/features/modules/decls/VerbPanel.tsx
deniseli: ah yeah - I couldn't come up with a better name that wasn't disgustingly long. Definitely open to suggestions if you come up with anything 😂

wesbillman: yeah it's a tricky one! we can just leave it for now since we'll probably follow up with verb/decl pages next.

deniseli: perfect

PR #2555, opened 2024-08-29
feat: console: add verb request editor + reset button

deniseli:
Adds new externally-controlled codemirror component with a reset button

Really basic, UI-side only. Doesn't hook into actually sending the call yet.

https://github.com/user-attachments/assets/f773bf49-6bb4-4ed9-862d-2f4b180b151b

### DIFF
--- a/frontend/src/components/CodeEditorV2.tsx
+++ b/frontend/src/components/CodeEditorV2.tsx
@@ -27,6 +27,7 @@ export const CodeEditor = ({
   schema?: string
   readOnly?: boolean
   onChange?: (text: string) => void
+  // A "Reset" button will be shown if and only if this prop is provided.
   defaultContent?: string
 }) => {
   const { isDarkMode } = useUserPreferences()
--- a/frontend/src/components/CodeEditorV2.tsx
+++ b/frontend/src/components/CodeEditorV2.tsx
@@ -1,0 +1,130 @@
+import { EditorState, type Extension } from '@codemirror/state'
+import { EditorView, drawSelection, gutter, highlightActiveLineGutter, hoverTooltip, keymap, lineNumbers } from '@codemirror/view'
+
+import { bracketMatching, defaultHighlightStyle, foldGutter, foldKeymap, indentOnInput, syntaxHighlighting } from '@codemirror/language'
+import { lintGutter } from '@codemirror/lint'
+import { lintKeymap } from '@codemirror/lint'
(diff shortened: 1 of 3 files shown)


frontend/src/components/CodeEditorV2.tsx
```diff
@@ -0,0 +1,130 @@
...
+  content: string
+  schema?: string
+  readOnly?: boolean
+  onChange?: (text: string) => void
+  defaultContent?: string
+}) => {
+  const { isDarkMode } = useUserPreferences()
+  const containerRef = useRef(null)
+  const viewRef = useRef<EditorView | null>(null)
+  const [lastReset, setLastReset] = useState(Date.now())
```
wesbillman, line 35: Should this be optional? I'm guessing we'll still use this editor for displaying `json` and `schema` stuff that wouldn't need a `reset` button.

frontend/src/components/CodeEditorV2.tsx
```diff
@@ -0,0 +1,130 @@
...
+import { atomone } from '@uiw/codemirror-theme-atomone'
+import { githubLight } from '@uiw/codemirror-theme-github'
+
+import { defaultKeymap, history, indentWithTab, redo, undo } from '@codemirror/commands'
+import { handleRefresh, jsonSchemaHover, jsonSchemaLinter, stateExtensions } from 'codemirror-json-schema'
+import { json5, json5ParseLinter } from 'codemirror-json5'
+import { useEffect, useRef, useState } from 'react'
+import { useUserPreferences } from '../providers/user-preferences-provider'
+
+export const CodeEditor = ({
```
wesbillman, line 19: Does it make sense to just ditch the old editor and replace the usages with this one instead?

wesbillman (approved):

frontend/src/components/CodeEditorV2.tsx
deniseli: Yes! The optionality is controlled by the `defaultContent` prop. I just have the `lastReset` state getting created no matter what because handling nullability on that gets really messy. I'll add a comment above that prop explaining it.

frontend/src/components/CodeEditorV2.tsx
deniseli: Sadly they handle the state slightly differently, so it'd be a more time consuming refactor than is worth it :/

PR #2557, opened 2024-08-29
feat: make the module tree scrollable + scroll to selected module

deniseli:
Final part of https://github.com/TBD54566975/ftl/issues/2487

https://github.com/user-attachments/assets/eb7ed1eb-d8ae-4a92-b56e-7bde8aed3a6b


wesbillman (approved):
So good!

Issue #2490, opened 2024-08-23
Review the `Page` and `Header` components


wesbillman:
These were added in the old design, are they still useful?

deniseli:
Sending out the first PR removing `Page.Header` soon. I'll wait on removing the `Page.Body` component until after rewriting the module(s) pages, since those still use it, and I'd rather not refactor the CSS if I'm about to rip it all out anyway.

PR #2560, opened 2024-08-29
feat: delete Deployments tab, module tree uses old deployments components

deniseli:
* Removes `Deployments` tab
* Removes all the extraneous page headers (https://github.com/TBD54566975/ftl/issues/2490)
* Fixes all the routing to point to `/modules/{moduleName}/...` instead of `/deployments/{deploymentKey}/...`
* Minor cleanup of deployment components' internal logic after replacing `deploymentKey` with `moduleName` in the outermost interface

https://github.com/user-attachments/assets/ad785b7c-3e88-4d8d-8edb-b14bed441c98


wesbillman (commented):
So nice to be able to reuse all this for now! Looks good!!

This does break the scrolling in the treeview/page so we should fix that first then I'm good to merge 👍 

Screenshot 2024-08-30 at 9 18 25 AM
Screenshot 2024-08-30 at 9 18 31 AM
Screenshot 2024-08-30 at 9 19 46 AM

deniseli:
> So nice to be able to reuse all this for now! Looks good!!
> 
> This does break the scrolling in the treeview/page so we should fix that first then I'm good to merge 👍
> 
> [Screenshots redacted]

Rebasing since the scrollability PR merged seemed to fix it. Thanks for catching that!

wesbillman (approved):

PR #2614, opened 2024-09-04
fix: use icon to indicate exported

wesbillman:
Screenshot 2024-09-04 at 12 03 47 PM
Screenshot 2024-09-04 at 12 03 58 PM


deniseli (approved):
Gorgeous <3

Issue #2616, opened 2024-09-04
Add pages for each decl type


deniseli:
Linked from the module tree view

PR #2617, opened 2024-09-04
feat: new console pages for config, data, database, enum, secret, and typealias

deniseli:
Part 1 of https://github.com/TBD54566975/ftl/issues/2616

Adds (or significantly adds to) pages for the following decl types:
* config
* data
* database
* enum
* secret
* typealias

Not yet added: FSM, topic, subscription

Adds components for:
* `DeclLink`: takes a ref and links to the page for that decl
* `TypeEl`: renders an appropriate string for a given Type, with links if appropriate
* `PanelHeader` renders the standard decl panel page header (e.g. comments, export badge)

### DIFF
--- a/frontend/console/src/features/modules/decls/ConfigPanel.tsx
+++ b/frontend/console/src/features/modules/decls/ConfigPanel.tsx
@@ -6,9 +6,7 @@ export const ConfigPanel = ({ value, moduleName, declName }: { value: Config; mo
   return (
     <div className='py-2 px-4'>
       <PanelHeader exported={false} comments={value.comments}>
-        <p>
-          Config: {moduleName}.{declName}
-        </p>
+        Config: {moduleName}.{declName}
       </PanelHeader>
       <div className='text-sm my-4'>
         Type: <TypeEl t={value.type} />
--- a/frontend/console/src/features/modules/decls/ConfigPanel.tsx
+++ b/frontend/console/src/features/modules/decls/ConfigPanel.tsx
@@ -1,0 +1,18 @@
+import type { Config } from '../../../protos/xyz/block/ftl/v1/schema/schema_pb'
+import { PanelHeader } from './PanelHeader'
+import { TypeEl } from './TypeEl'
+
+export const ConfigPanel = ({ value, moduleName, declName }: { value: Config; moduleName: string; declName: string }) => {
+  return (
+    <div className='py-2 px-4'>
+      <PanelHeader exported={false} comments={value.comments}>
+        <p>
+          Config: {moduleName}.{declName}
+        </p>
+      </PanelHeader>
+      <div className='text-sm my-4'>
+        Type: <TypeEl t={value.type} />
+      </div>
+    </div>
+  )
+}
--- a/frontend/console/src/features/modules/decls/DataPanel.tsx
+++ b/frontend/console/src/features/modules/decls/DataPanel.tsx
@@ -1,21 +1,25 @@
-import { Badge } from '../../../components/Badge'
 import type { Data } from '../../../protos/xyz/block/ftl/v1/schema/schema_pb'
+import { PanelHeader } from './PanelHeader'
+import { TypeEl } from './TypeEl'
 
 export const DataPanel = ({ value, moduleName, declName }: { value: Data; moduleName: string; declName: string }) => {
+  const maybeTypeParams = value.typeParameters.length === 0 ? '' : `<${value.typeParameters.map((p) => p.name).join(', ')}>`
   return (
-    <div className='flex-1 py-2 px-4'>
-      {value.export ? (
-        <div>
-          <Badge name='Exported' />
-        </div>
-      ) : (
-        []
-      )}
-      <div className='inline-block mr-3 align-middle'>
+    <div className='py-2 px-4'>
+      <PanelHeader exported={value.export} comments={value.comments}>
         <p>
           data: {moduleName}.{declName}
+          {maybeTypeParams}
         </p>
-        {value.comments.length > 0 ? <p className='text-xs my-1'>{value.comments}</p> : []}
+      </PanelHeader>
+      {value.fields.length === 0 || <div className='mt-8 mb-3'>Fields</div>}
+      <div className='text-xs font-mono inline-grid grid-cols-2 gap-x-4 gap-y-2' style={{ gridTemplateColumns: 'auto auto' }}>
+        {value.fields.map((f, i) => [
+          <span key={`field-name-${i}`}>{f.name}</span>,
+          <span key={`field-type-${i}`}>
+            <TypeEl t={f.type} />
+          </span>,
+        ])}
       </div>
     </div>
   )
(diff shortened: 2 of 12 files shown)


frontend/console/src/features/modules/decls/ConfigPanel.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { PanelHeader } from './PanelHeader'
+import { TypeEl } from './TypeEl'
+
+export const ConfigPanel = ({ value, moduleName, declName }: { value: Config; moduleName: string; declName: string }) => {
+  return (
+    <div className='py-2 px-4'>
+      <PanelHeader exported={false} comments={value.comments}>
+        <p>
+          Config: {moduleName}.{declName}
+        </p>
```
wesbillman, line 11: It seems like all the `PanelHeader` have this `<p>` tag in them, should we add that to the header itself, or keep separate? Or should we remove here and just show whatever text?

frontend/console/src/features/modules/decls/DeclLink.tsx
```diff
@@ -0,0 +1,26 @@
...
+  const decl = useMemo(() => {
+    const modules = (schema?.data || []) as PullSchemaResponse[]
+    const module = modules.find((m: PullSchemaResponse) => m.moduleName === moduleName)
+    if (!module?.schema) {
+      return
+    }
+    return module.schema.decls.find((d) => d.value.value?.name === declName)
+  }, [moduleName, declName, schema?.data])
+
+  const str = moduleName ? `${moduleName}.${declName}` : declName
```
wesbillman, line 17: I've been personally finding that it's easier to make "dumb" components and feed the data into them, vs. having to have `useSchema` type hooks in the smaller sub components. This also makes them much easier to write `storybook` stories for since they don't have any real dependencies. Whatcha think?

frontend/console/src/features/modules/decls/TypeEl.tsx
```diff
@@ -0,0 +1,66 @@
+import type { Optional, Ref, Array as SchArray, Map as SchMap, Type } from '../../../protos/xyz/block/ftl/v1/schema/schema_pb'
+import { DeclLink } from './DeclLink'
+
+// TypeParams ironically is built to work with the `Type` message, not the
+// `TypeParameter` message, which just has a simple string param name without
+// any higher level type information.
+const TypeParams = ({ types }: { types?: (Type | undefined)[] }) => {
+  if (!types) {
+    return []
```
wesbillman, line 9: I feel like these should return `null` or `undefined` instead of `[]` I keep getting hung up on the fact that things return either an empty array or a react component. Maybe we can pick a pattern and just make sure we use that everywhere?

wesbillman: Or even better, we should just `return` with nothing, which should be equivalent and more readable. 
```ts
if (!types) {
  return
}
```
or 
```ts
if (!types) return
```

frontend/console/src/features/modules/decls/TypeEl.tsx
```diff
@@ -0,0 +1,66 @@
...
+  }
+  const definedTypes = types.filter((t) => t !== undefined)
+  if (definedTypes.length === 0) {
+    return []
+  }
+  return (
+    <span>
+      <span>{'<'}</span>
+      {definedTypes.map((t, i) => [<TypeEl key='t' t={t} />, i === types.length - 1 ? '' : ', '])}
```
wesbillman, line 18: should this be using `definedTypes.length` here? instead of `types.length`?

Also, I think this can be refactored to take advantage of optionals.

```ts
const TypeParams = ({ types }: { types?: (Type | undefined)[] }) => {
  const definedTypes = types?.filter((t) => t !== undefined) || []
  if (definedTypes.length === 0) return
    
  return (
    <span>
      <span>{'<'}</span>
      {definedTypes.map((t, i) => [<TypeEl key='t' t={t} />, i === definedTypes.length - 1 ? '' : ', '])}
      <span>{'>'}</span>
    </span>
  )
}
```

frontend/console/src/features/modules/decls/DataPanel.tsx
```diff
@@ -1,21 +1,25 @@
...
         </p>
-        {value.comments.length > 0 ? <p className='text-xs my-1'>{value.comments}</p> : []}
+      </PanelHeader>
+      {value.fields.length === 0 || <div className='mt-8 mb-3'>Fields</div>}
+      <div className='text-xs font-mono inline-grid grid-cols-2 gap-x-4 gap-y-2' style={{ gridTemplateColumns: 'auto auto' }}>
+        {value.fields.map((f, i) => [
+          <span key={`field-name-${i}`}>{f.name}</span>,
+          <span key={`field-type-${i}`}>
+            <TypeEl t={f.type} />
+          </span>,
```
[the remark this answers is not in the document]
deniseli: nice catch! fixed

frontend/console/src/features/modules/decls/DeclLink.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+  return !decl ? (
+    str
+  ) : (
+    <Link className='rounded-md cursor-pointer hover:bg-gray-100 hover:dark:bg-gray-700 p-1 -m-1' to={`/modules/${moduleName}/${decl.value.case}/${declName}`}>
+      {str}
+    </Link>
+  )
+}
```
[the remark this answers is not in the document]
deniseli: good idea, done

frontend/console/src/features/modules/decls/ConfigPanel.tsx
deniseli: I'll just pull out all the <p> tags since we they don't actually have an effect. I think we needed that in there at some point early on and now it's just not necessary.

frontend/console/src/features/modules/decls/TypeEl.tsx
deniseli: In the past, I've noticed both null and undefined ended up being treated as real children in the React component tree when you returned anything other than `[]`. I'll see if that's still an issue.

deniseli: They finally fixed it!! Yay! Cool I'll switch to returning nothing/undefined

frontend/console/src/features/modules/decls/DeclLink.tsx
deniseli: Normally I follow that pattern too, but in this case, `DeclLink` is used by `TypeEl`, which in turn is used by most of the decl-type panels. We'd end up having to thread it through all the way, which would make the props for _all_ of those components more complicated just for _only_ `DeclLink` to actually use the contents. It gets worse from there, because `DeclLink` is probably going to end up all over the console once we're through, at which point we're basically passing `schema` down through nearly every single component in the tree. That would almost completely defeats the purpose of the hook pattern to begin with. :/

wesbillman: Awesome! Yeah, definitely seems easier to keep this here then. Thanks for the clarification!

wesbillman (approved):
Niiiiice! This is awesome

PR #2620, opened 2024-09-05
chore: remove Quarkus banner

stuartwdouglas:


jvm-runtime/ftl-runtime/common/runtime/src/main/resources/ftl-banner.txt
tomdaffurn, line 1: laaaaiiiiiid baaaack

tomdaffurn:
As cool as this looks, I think I'd rather not see this for every JVM module, every build

stuartwdouglas:
> As cool as this looks, I think I'd rather not see this for every JVM module, every build

Boo, but fair enough

tomdaffurn (approved):

Issue #2610, opened 2024-09-04
Treeview layout bug


wesbillman:
The highlighted item background doesn't fill all the way to the right when an exported badge is present. See below

We should probably truncate or find a way to make this layout look a bit more polished. 

![Screenshot 2024-09-04 at 8 53 10 AM](https://github.com/user-attachments/assets/fba7ad45-18f7-4f75-a414-66041085ba5a)

PR #2627, opened 2024-09-05
fix: polish module tree scroll and resize actions

deniseli:
Fixes https://github.com/TBD54566975/ftl/issues/2610

Issues fixed:
* The `export` icons:
  * Used to compress to a smaller size when you resized the tree panel smaller than the width of the decl's row. Now, the size is fixed.
  * Added on-hover text so users can find out what it means more easily.
* Panel resizing handle: 
  * Got rid of its background color so it's invisible until used
  * Set the width back to its original 3px, which is easier to use and not a problem now that it's invisible. I originally adjusted it down to 2px to be less visually intrusive.
  * Fixed the on-hover color in dark mode so it's always easy to latch onto it now.
* Highlight color of each decl row would only extend as far as the right-side border minus padding+margin. However, text would overflow beyond that container limit. This PR makes the decl items full-width and removes the rounded corners, so you can't see that the text is overflowing.
* Removed several layers of elements in the tree wrapping that just weren't necessary

### DIFF
--- a/frontend/console/src/features/modules/ModulesTree.tsx
+++ b/frontend/console/src/features/modules/ModulesTree.tsx
@@ -23,7 +23,11 @@ import { classNames } from '../../utils'
 import type { ModuleTreeItem } from './module.utils'
 import { addModuleToLocalStorageIfMissing, listExpandedModulesFromLocalStorage, toggleModuleExpansionInLocalStorage } from './module.utils'
 
-const ExportedIcon = () => <FileExportIcon className='size-4 text-indigo-500 -ml-1' />
+const ExportedIcon = () => (
+  <span className='w-4' title='Exported'>
+    <FileExportIcon className='size-4 text-indigo-500 -ml-1' />
+  </span>
+)
 
 type IconMap = Record<string, React.FC<Omit<HugeiconsProps, 'ref'> & React.RefAttributes<SVGSVGElement>>>
 const icons: IconMap = {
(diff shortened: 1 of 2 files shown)


frontend/console/src/features/modules/ModulesTree.tsx
```diff
@@ -23,7 +23,11 @@ import { classNames } from '../../utils'
 import type { ModuleTreeItem } from './module.utils'
 import { addModuleToLocalStorageIfMissing, listExpandedModulesFromLocalStorage, toggleModuleExpansionInLocalStorage } from './module.utils'
 
-const ExportedIcon = () => <FileExportIcon className='size-4 text-indigo-500 -ml-1' />
+const ExportedIcon = () => (
+  <span className='w-4' title='Exported'>
+    <FileExportIcon className='size-4 text-indigo-500 -ml-1' />
+  </span>
```
wesbillman, line 29: nit: and mostly curious, if we have a `span` with `w-4` and the icon in it is `size-4`, do we need both? Like can we just put the `size` on the wrapper span?

Also, thanks for adding `title` here! I was thinking the same thing, but didn't get to it yet :)

wesbillman (approved):

frontend/console/src/features/modules/ModulesTree.tsx
deniseli: For some reason, the `svg` treats the size as a max size instead of an exact size, so I basically added the span just to force it into a fixed-size container where it can't complain about losing space. My guess is there's something deep inside the implementation of the icon that tries to be "smart" and adjust size automatically, but that's just totally incompatible with complex containers like this. :/

PR #2662, opened 2024-09-12
feat: add decl pages for topics and subscriptions

deniseli:
Issue: https://github.com/TBD54566975/ftl/issues/2616

All the remains now is FSM.


wesbillman (approved):
Nice!

PR #2712, opened 2024-09-17
chore: remove Page.Header console component

deniseli:
Part 1 of https://github.com/TBD54566975/ftl/issues/2490

Deletes the `Page.Header` component, which is an artifact of a bygone era.

Since the timeline page also had a time range filter embedded in the header, I moved that over to the left-hand-side filter panel. It just need a tiny bit of resizing to fit.

Screenshot 2024-09-17 at 4 29 45 PM
Screenshot 2024-09-17 at 4 36 51 PM

### DIFF
--- a/frontend/console/src/features/console/ConsolePage.tsx
+++ b/frontend/console/src/features/console/ConsolePage.tsx
@@ -25,7 +25,6 @@ export const ConsolePage = () => {
 
   return (
     <Page>
-      <Page.Header icon={<CubeIcon />} title='Console' />
       <Page.Body className='flex h-full'>
         <ResizablePanels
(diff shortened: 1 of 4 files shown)


wesbillman (approved):
Legend! Thanks for cleaning this up! 

I'm guessing we should nuke `Page` at some point too maybe? Or at least make sure it's what we actually want.

deniseli:
> Legend! Thanks for cleaning this up!
> 
> I'm guessing we should nuke `Page` at some point too maybe? Or at least make sure it's what we actually want.

Yeah I was looking around at the usage and one of the main ones was the deployment page, which I'm about to rewrite anyway, so I'm going to just nuke `Page` after that to avoid churning CSS fixes.

PR #2722, opened 2024-09-18
fix: restore deployed log

stuartwdouglas:


jvmakine (approved):

wesbillman (approved):

backend/controller/controller.go
```diff
@@ -1767,6 +1767,7 @@ func (s *Service) syncRoutes(ctx context.Context) (ret time.Duration, err error)
 					continue
 				}
 				logger.Debugf("Adding %s to route table", v.Key.String())
+				s.getDeploymentLogger(ctx, v.Key).Infof("Deployed %s", v.Key.String())
```
wesbillman, line 1770: We can update the logger here to use `deploymentLogger` for all these if you wanna too. I'm not sure exactly what we want to log here, but if we have a deployment and the logs are related to that deployment, it might be a good idea to just make that the `logger` instance.

PR #2723, opened 2024-09-19
feat: new console module page

deniseli:
Screencast: https://sq-tbd.slack.com/archives/C04PEQERFM0/p1726706223996289


frontend/console/e2e/module.spec.ts
```diff
@@ -10,6 +10,5 @@ ftlTest('shows verbs for deployment', async ({ page }) => {
 
   await expect(page).toHaveURL(/\/modules\/echo/)
 
-  await expect(page.getByText('Deployment', { exact: true })).toBeVisible()
-  await expect(page.getByText('Created Deployment dpl-echo')).toBeVisible()
+  await expect(page.getByText('module', { exact: true })).toBeVisible()
```
wesbillman, line 13: Should we look for like `echo` or something since this should be the echo module? :)

frontend/console/src/features/modules/decls/DeclLink.tsx
```diff
@@ -3,12 +3,13 @@ import { useNavigate } from 'react-router-dom'
 import { useSchema } from '../../../api/schema/use-schema'
 import type { PullSchemaResponse } from '../../../protos/xyz/block/ftl/v1/ftl_pb.ts'
 import type { Decl } from '../../../protos/xyz/block/ftl/v1/schema/schema_pb'
+import { classNames } from '../../../utils'
 import { DeclSnippet } from './DeclSnippet'
 
 const SnippetContainer = ({ decl }: { decl: Decl }) => {
```
wesbillman, line 9: Should we use a headless ui popup for this? It might help with ensuring this is always visible even if it's at the bottom of the page (or far right of the page) so it doesn't get cut off.

Screenshot 2024-09-18 at 7 29 39 PM

wesbillman (approved):
So much regex! 😂

frontend/console/src/features/modules/decls/DeclLink.tsx
wesbillman: This might become more of an issue when we're update the modules page to show more than just the schema. For example, we'll probably want to show traces/events etc. on the modules page as well, which would mean this might only be show on left or right half of the page. So having a popup that can adapt will be very useful.

frontend/console/e2e/module.spec.ts
deniseli: thanks for catching that! fixed

PR #2742, opened 2024-09-20
chore: add ftlDebug script

stuartwdouglas:
This makes it easy to run FTL under dlv, and dlv will wait for the IDE to attach. This is useful if you want to debug commands other than ftl dev


jvmakine (approved):
Nice! the camel case for the script name feels a bit strange, but I can not come up with anything better.

stuartwdouglas:
I took the same naming approach as mvnDebug, which is a script maven provides to allow you to connect a debugger to the build process.

PR #2744, opened 2024-09-20
feat: add replay command

stuartwdouglas:


frontend/cli/cmd_replay.go
```diff
@@ -0,0 +1,103 @@
...
+			},
+		},
+	}))
+	if err != nil {
+		return fmt.Errorf("failed to get events: %w", err)
+	}
+	if len(events.Msg.GetEvents()) == 0 {
+		return fmt.Errorf("no events found for %v", c.Verb)
+	}
+	println(events.Msg.GetEvents()[0].String())
```
worstell, line 98: is this intentional?

worstell (approved):

frontend/cli/cmd_replay.go
stuartwdouglas: oops

PR #2745, opened 2024-09-20
chore: refactor go-runtime build

worstell:
this file was becoming difficult to reason with, particularly in regards to generating imports in the scaffolded files

fixes redundancies and unifies the logic for generating imports across the various sources that may require importing. also ensures everything goes through the logic @matt2e added to avoid import conflicts


alecthomas (approved):

go-runtime/compile/build.go
```diff
@@ -404,6 +480,237 @@ func SyncGeneratedStubReferences(ctx context.Context, projectRootDir string, stu
...
+		}
+
+		resolved, ok := b.sch.Resolve(n).Get()
+		if !ok {
+			return optional.None[goType](), b.visitingMainModule(), nil
+		}
+
+		// update the current module for the visit to the resolved node
+		currentModule := b.currentModule
+		b.currentModule = n.Module
```
deniseli, line 592: Would it make sense to refactor this slightly so that the caller of `getGoType` doesn't need to maintain the `currentModule` state, and instead, `getGoType` does that itself?

go-runtime/compile/build-template/.ftl.tmpl/go/main/main.go.tmpl
```diff
@@ -1,57 +1,47 @@
+{{- $verbs := .Verbs -}}
+{{- $name := .Name -}}
+{{- with .MainCtx -}}
+
```
deniseli, line 4: nit: does this newline result in the generated code having an extra newline before the package comment? I'm guessing it's fine since `types.ftl.go.tmpl` has it too, but just wanted to make sure

go-runtime/compile/build.go
```diff
@@ -48,45 +48,155 @@ type ExternalModuleContext struct {
...
+	return out
+}
+
+func (c mainModuleContext) generateTypesImports() []string {
+	imports := sets.NewSet[string]()
+	if len(c.TypesCtx.SumTypes) > 0 || len(c.TypesCtx.ExternalTypes) > 0 {
+		imports.Add(`"github.com/TBD54566975/ftl/go-runtime/ftl/reflection"`)
+	}
+	for _, st := range c.TypesCtx.SumTypes {
+		// subpackage
```
deniseli, line 105: nit: would you mind making these two comments something like `// if import path is more than 2 dirs deep, it's a subpackage`? It feels obvious now that I've grokked it, but it took me a minute to figure out initially. Already super helpful having the existing comment there anyway so thanks :D

deniseli (approved):
I've wanted this for so long hehe this is awesome!! Thanks :D

go-runtime/compile/build-template/.ftl.tmpl/go/main/main.go.tmpl
worstell: just separated to make the template more readable; the dash at the end of `{{- with .MainCtx -}}`  trims any whitespace/newlines after the block, so it doesn't affect the generated code

go-runtime/compile/build.go
worstell: you mean passing the current module name through `getGoType` rather than keeping it in the `mainModuleContextBuilder`? i was going back and forth on which approach would read less confusing, but i think you're right! i'll update

PR #2778, opened 2024-09-23
fix: include raw response body

stuartwdouglas:
Currently by the time the response body is added into the timeline it has been deserialized, so may not longer be valid JSON.


stuartwdouglas:
@wesbillman without this change I was seeing errors for requests that return a string

wesbillman (approved):
Nice! Thank you :)

PR #2792, opened 2024-09-23
feat: instrument app DBs

mistermoe:
This PR instruments app DBs to report db related metrics for application dbs. I verified that it's working locally

image

> [!WARNING]
> I haven't been able to verify whether the appropriate service name is being set. [This](https://github.com/tbd54566975/ftl/blob/main/go-runtime/server/server.go#L50) made me feel like it would but i'm not 100% certain yet. I need to figure out whether grafana treats `service.name` as a label that can be queried. If that is the case then `service.name` isn't being set in which case we'll need to provide a custom metrics provider to `otelsql`. We can hold off on merging this PR until i get that figured out or we can do it in a subsequent PR.
>
> FWIW, i am setting the `db_name` attribute to the appropriate db name, so there technically is a way to distinguish between ftl db metrics and app db metrics right now, though it's definitely unideal

### DIFF
--- a/go-runtime/ftl/database.go
+++ b/go-runtime/ftl/database.go
@@ -42,7 +42,6 @@ func PostgresDatabase(name string) Database {
 				semconv.DBSystemPostgreSQL,
 				semconv.DBNameKey.String(name),
 				attribute.Bool("ftl.is_user_service", true),
-				attribute.String("ftl.sql.module.name", provider.module),
 			)
 			err = otelsql.RegisterDBStatsMetrics(db, metricAttrs)
 			if err != nil {
--- a/go-runtime/ftl/database.go
+++ b/go-runtime/ftl/database.go
@@ -8,6 +8,7 @@ import (
 	"github.com/XSAM/otelsql"
 	"github.com/alecthomas/types/once"
 	_ "github.com/jackc/pgx/v5/stdlib" // Register Postgres driver
+	"go.opentelemetry.io/otel/attribute"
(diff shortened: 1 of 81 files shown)


go-runtime/ftl/database.go
```diff
@@ -29,10 +31,17 @@ func PostgresDatabase(name string) Database {
...
 				return nil, fmt.Errorf("failed to get database %q: %w", name, err)
 			}
-			db, err := sql.Open("pgx", dsn)
+			db, err := otelsql.Open("pgx", dsn)
 			if err != nil {
 				return nil, fmt.Errorf("failed to open database %q: %w", name, err)
 			}
+
+			// sets db.system and db.name attributes
+			metricAttrs := otelsql.WithAttributes(semconv.DBSystemPostgreSQL, semconv.DBNameKey.String(name))
```
safeer, line 40: Let's also add `ftl.is_user_service` and `ftl.sql.module.name` as attributes?
```suggestion
			metricAttrs := otelsql.WithAttributes(
				semconv.DBSystemPostgreSQL,
				semconv.DBNameKey.String(name),
				attribute.Bool("ftl.is_user_service", true),
				attribute.String("ftl.sql.module.name", provider.module),
			)
```

safeer (approved):
Nice! Also guessing this also works without an active otel collector?
(cc @jonathanj-square metrics DRI)

AlexSzlavik (approved):

alecthomas (approved):

PR #2800, opened 2024-09-24
chore(refactor): add utils for timeline and requests

wesbillman:
- Add utils for sharing colors for events etc. to make them easily changeable and consistent
- Add utils for shared computations on event bars
- Add child components to make the component sizes more manageable
- Minor adjustments to the bar colors to be more subtle

Screenshot 2024-09-24 at 9 26 29 AM
Screenshot 2024-09-24 at 9 26 48 AM
Screenshot 2024-09-24 at 9 27 01 AM
Screenshot 2024-09-24 at 9 27 21 AM

### DIFF
--- a/frontend/console/src/features/timeline/TimelineIcon.tsx
+++ b/frontend/console/src/features/timeline/TimelineIcon.tsx
@@ -1,23 +1,24 @@
 import { Call02Icon, CallIncoming04Icon, Menu01Icon, PackageReceiveIcon, Rocket01Icon } from 'hugeicons-react'
 import type { Event } from '../../protos/xyz/block/ftl/v1/console/console_pb'
 import { LogLevelBadgeSmall } from '../logs/LogLevelBadgeSmall'
+import { eventTextColor } from './timeline.utils'
 
 export const TimelineIcon = ({ event }: { event: Event }) => {
   const icon = (event: Event) => {
-    const style = 'h4 w-4 text-indigo-500'
+    const style = 'h4 w-4'
+    const textColor = eventTextColor(event)
+
     switch (event.entry.case) {
       case 'call': {
-        const textColor = event.entry.value.error ? 'text-red-600' : 'text-indigo-500'
         return event.entry.value.sourceVerbRef ? <Call02Icon className={`${style} ${textColor}`} /> : <CallIncoming04Icon className={`${style} ${textColor}`} />
       }
       case 'deploymentCreated':
-        return <Rocket01Icon className='h4 w-4 text-green-500' />
+        return <Rocket01Icon className={`${style} ${textColor}`} />
       case 'deploymentUpdated':
-        return <Rocket01Icon className='h4 w-4 text-indigo-500' />
+        return <Rocket01Icon className={`${style} ${textColor}`} />
       case 'log':
         return <LogLevelBadgeSmall logLevel={event.entry.value.logLevel} />
       case 'ingress': {
-        const textColor = event.entry.value.error ? 'text-red-600' : 'text-indigo-500'
         return <PackageReceiveIcon className={`${style} ${textColor}`} />
       }
       default:
--- a/frontend/console/src/features/timeline/filters/TimelineFilterPanel.tsx
+++ b/frontend/console/src/features/timeline/filters/TimelineFilterPanel.tsx
@@ -21,12 +21,12 @@ const EVENT_TYPES: Record<string, EventFilter> = {
   deploymentCreated: {
     label: 'Deployment Created',
     type: EventType.DEPLOYMENT_CREATED,
-    icon: <Rocket01Icon className='w-4 h-4 text-green-500 ml-1' />,
+    icon: <Rocket01Icon className='w-4 h-4 text-green-500 dark:text-green-300 ml-1' />,
   },
   deploymentUpdated: {
     label: 'Deployment Updated',
     type: EventType.DEPLOYMENT_UPDATED,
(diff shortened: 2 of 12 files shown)


frontend/console/src/features/timeline/TimelineIcon.tsx
```diff
@@ -1,23 +1,24 @@
...
+    const textColor = eventTextColor(event)
+
     switch (event.entry.case) {
       case 'call': {
-        const textColor = event.entry.value.error ? 'text-red-600' : 'text-indigo-500'
         return event.entry.value.sourceVerbRef ? <Call02Icon className={`${style} ${textColor}`} /> : <CallIncoming04Icon className={`${style} ${textColor}`} />
       }
       case 'deploymentCreated':
-        return <Rocket01Icon className='h4 w-4 text-green-500' />
+        return <Rocket01Icon className={`${style} ${textColor}`} />
```
deniseli, line 16: Do you wanna `classNames(...)` this?

frontend/console/src/features/timeline/filters/TimelineFilterPanel.tsx
```diff
@@ -28,7 +28,7 @@ const EVENT_TYPES: Record<string, EventFilter> = {
     type: EventType.DEPLOYMENT_UPDATED,
     icon: <Rocket01Icon className='w-4 h-4 text-indigo-500 ml-1' />,
   },
-  ingress: { label: 'Ingress', type: EventType.INGRESS, icon: <PackageReceiveIcon className='w-4 h-4 text-indigo-500 ml-1' /> },
+  ingress: { label: 'Ingress', type: EventType.INGRESS, icon: <PackageReceiveIcon className='w-4 h-4 text-sky-400 ml-1' /> },
```
deniseli, line 31: niiiice this color is perfect

frontend/console/src/features/timeline/timeline.utils.ts
```diff
@@ -0,0 +1,23 @@
+import type { Event } from '../../protos/xyz/block/ftl/v1/console/console_pb'
+
+const eventBackgroundColorMap: Record<string, string> = {
+  log: 'bg-gray-500',
+  call: 'bg-indigo-500',
```
deniseli, line 5: What do you think about making this one different from the deployment event colors? Possibly even just a different shade of indigo?

frontend/console/src/features/traces/TraceDetailItem.tsx
```diff
@@ -0,0 +1,74 @@
...
+    eventName = `${traceEvent.destinationVerbRef?.module}.${traceEvent.destinationVerbRef?.name}`
+  } else if (traceEvent instanceof IngressEvent) {
+    action = `HTTP ${traceEvent.method}`
+    eventName = `${traceEvent.path}`
+  }
+
+  let barColor = eventBackgroundColor(event)
+  if (event.id === selectedEventId) {
+    barColor = 'bg-green-500'
+  }
```
deniseli, line 47: It'd be nice to get rid of this `let` with:

```
const barColor = event.id === selectedEventId ? 'bg-green-500' : eventBackgroundColor(event)
```

frontend/console/src/features/traces/TraceDetailItem.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+  const isSelected = event.id === selectedEventId
+  const listItemClass = isSelected
+    ? 'flex items-center justify-between p-2 bg-indigo-100/50 dark:bg-indigo-700 rounded cursor-pointer'
+    : 'flex items-center justify-between p-2 hover:bg-indigo-500/10 rounded cursor-pointer'
```
deniseli, line 52: Since there's so much overlap between these two branch cases, what do you think about something like:

```
const listItemClass = classNames(isSelected ? '...' : '...', 'shared class names')
```

As is, it took me a minute to parse out the exact differences between the two, since there are some shared classes at both the head and tail ends

frontend/console/src/features/traces/traces.utils.ts
```diff
@@ -0,0 +1,14 @@
+import type { Timestamp } from '@bufbuild/protobuf'
+import type { Event } from '../../protos/xyz/block/ftl/v1/console/console_pb'
+
+export const eventBarLeftOffsetPercentage = (event: Event, requestStartTime: Timestamp | undefined, requestDurationMs: number) => {
+  if (!requestStartTime) {
+    return 0
+  }
+
+  const callTime = event.timeStamp?.toDate() ?? new Date()
+  const initialTime = requestStartTime?.toDate() ?? new Date()
```
deniseli, line 10: Is defaulting to the current datetime the behavior we really want here? Should we just return early instead when the input timestamps aren't provided?

deniseli (approved):

frontend/console/src/features/timeline/TimelineIcon.tsx
deniseli: retracting this comment lol

frontend/console/src/features/timeline/timeline.utils.ts
wesbillman: Honestly, I'm not sure :) I like the idea of each thing having its own color, but it also starts to get very colorful really quick! I'll give it a shot to see what we think. As I start to think about all the other things like `pubsub`, `db`, etc. we might end up with like "theme" colors so they are somehow connected to each other.

wesbillman: I'll go with this light green for now and we can adjust if we wanna.

Screenshot 2024-09-24 at 11 54 32 AM

PR #2801, opened 2024-09-24
fix: super tiny fix to tab count indicators

wesbillman:
BEFORE:
Screenshot 2024-09-24 at 9 56 13 AM

AFTER:
Screenshot 2024-09-24 at 9 57 19 AM


deniseli:
nice!

Issue #2759, opened 2024-09-20
console snippet improvements


deniseli:
* [x] Make the snippets handle page borders (not overflow down or right)
* [x] Snippet text should come from module schema
* [x] In the backend we have some [custom grouping](https://github.com/TBD54566975/ftl/blob/05eee6aa6be61ba3e910c15ac7ffb892b9b9897b/backend/schema/module.go) rules to reduce vertical spacing, which would be nice to reproduce in the frontend. They're pretty small and keeping them in sync isn't critical, so the rules can probably just be copy pastad.
* [x] Drop triangle and add border
* [x] Syntax highlighting

PR #2804, opened 2024-09-24
feat: facelift to console schema snippets

deniseli:
Fixes https://github.com/TBD54566975/ftl/issues/2759

* Snippet text comes from module schema
* In the backend we have some [custom grouping](https://github.com/TBD54566975/ftl/blob/05eee6aa6be61ba3e910c15ac7ffb892b9b9897b/backend/schema/module.go) rules to reduce vertical spacing, which would be nice to reproduce in the frontend. They're pretty small and keeping them in sync isn't critical, so the rules can probably just be copy pastad.
* Drop triangle and add border
* Syntax highlighting
* Make the snippets handle page borders - position to the top / left when there is not enough space on the screen


frontend/console/src/features/modules/schema/schema.utils.ts
```diff
@@ -28,3 +71,56 @@ export function isFirstLineOfBlock(ll: string[], i: number): boolean {
...
+  })
+
+  if (foundIdx === -1) {
+    return
+  }
+
+  const line = lines[foundIdx]
+  let out = line
+  let subLineIdx = foundIdx + 1
+  while (subLineIdx < lines.length && lines[subLineIdx].startsWith('    ')) {
```
safeer, line 105: nit: does it make sense to do `startsWith(' (\\w+)')` here as below instead of 4 spaces?

frontend/console/src/features/modules/schema/schema.utils.ts
```diff
@@ -28,3 +71,56 @@ export function isFirstLineOfBlock(ll: string[], i: number): boolean {
...
+    subLineIdx++
+  }
+  // Check for closing parens
+  if (subLineIdx < lines.length && line.endsWith('{') && lines[subLineIdx] === '  }') {
+    out += '\n  }'
+  }
+
+  // Scan backwards for comments
+  subLineIdx = foundIdx - 1
+  while (subLineIdx >= 0 && lines[subLineIdx].startsWith(commentPrefix)) {
```
safeer, line 116: nice

safeer (approved):
LGTM. I just skimmed the React stuff though 😅

frontend/console/src/features/modules/schema/LinkTokens.tsx
```diff
@@ -2,24 +2,24 @@ import { useParams } from 'react-router-dom'
...
 
-export const LinkVerbNameToken = ({ token }: { token: string }) => {
+export const LinkVerbNameToken = ({ token, containerRect }: { token: string; containerRect?: DOMRect }) => {
   const splitToken = token.split('(')
   return (
     <span>
-      <LinkToken token={splitToken[0]} />
-      (<UnderlyingType token={splitToken[1]} />
+      <LinkToken token={splitToken[0]} containerRect={containerRect} />
+      (<UnderlyingType token={splitToken[1]} containerRect={containerRect} />
```
wesbillman, line 22: do we need to check for `splitToken.length` before these?

frontend/console/src/features/modules/schema/schema.utils.ts
```diff
@@ -1,12 +1,55 @@
...
+const skipNewLineDeclTypes = ['config', 'secret', 'database', 'topic', 'subscription']
+const skipGapAfterTypes: { [key: string]: string[] } = {
+  secret: ['config'],
+  subscription: ['topic'],
+}
+
 export const specialChars = ['{', '}', '=']
 
-export function isFirstLineOfBlock(ll: string[], i: number): boolean {
+export function shouldAddLeadingSpace(ll: string[], i: number): boolean {
```
wesbillman, line 19: What are `ll`s in these?

frontend/console/src/features/modules/schema/schema.utils.ts
deniseli: Great question! So the behavior here is slightly different - basically, the `(\\w+)` extracts the decl type from the line but ignores the start of the line, whereas this line is just checking the nesting depth from the start of the line.

frontend/console/src/features/modules/schema/schema.utils.ts
```diff
@@ -28,3 +71,56 @@ export function isFirstLineOfBlock(ll: string[], i: number): boolean {
...
+  }
+  const module = modules.data.modules.find((module) => module.name === moduleName)
+  if (!module?.schema) {
+    return
+  }
+  return declFromModuleSchemaString(declName, module.schema)
+}
+
+export function declFromModuleSchemaString(declName: string, schema: string) {
+  const lines = schema.split('\n')
```
wesbillman, line 92: A lot of this function and file seems to be scanning the schema string to pull it back apart into decls, but we can get the raw schema as well and not have to deal with regexing on the string if that's easier.

deniseli: posting update from our slack discussion: we'll do that with the new endpoint: https://github.com/TBD54566975/ftl/issues/2805

frontend/console/src/features/modules/schema/LinkTokens.tsx
deniseli: Added. It _should_ never happen, but if we add a new caller that doesn't check for it, then yeah the guard would definitely come in handy

frontend/console/src/features/modules/schema/schema.utils.ts
deniseli: renamed to `lines`

Issue #2807, opened 2024-09-24
Killing the `ftl` interactive command causes panic


wesbillman:
Start `ftl dev --recreate`. Wait for echo and time to deploy, then `CTRL+C` to kill it.

```sh
ftl dev --recreate

info:controller0: Deployed dpl-echo-43mar6x7owpoqa01
info: All modules deployed in 13.974872167s, watching for changes...
echo:  ✅  time:  ✅
panic: channel not subscribed

goroutine 69 [running]:
github.com/alecthomas/types/pubsub.(*Topic[...]).Unsubscribe(0x104b437a0, 0x14000666e00)
	/Users/wesbillman/go/pkg/mod/github.com/alecthomas/types@v0.16.0/pubsub/pubsub.go:139 +0xb4
github.com/TBD54566975/ftl/internal/buildengine.(*Engine).watchForModuleChanges.func2()
	/Users/wesbillman/dev/ftl/internal/buildengine/engine.go:369 +0x34
github.com/TBD54566975/ftl/internal/buildengine.(*Engine).watchForModuleChanges(0x140006eb2c0, {0x104b0f0d0, 0x14000a1a690}, 0x1dcd6500)
	/Users/wesbillman/dev/ftl/internal/buildengine/engine.go:411 +0xec0
github.com/TBD54566975/ftl/internal/buildengine.(*Engine).Dev(...)
	/Users/wesbillman/dev/ftl/internal/buildengine/engine.go:337
main.(*devCmd).Run.func2()
	/Users/wesbillman/dev/ftl/frontend/cli/cmd_dev.go:102 +0x5cc
golang.org/x/sync/errgroup.(*Group).Go.func1()
	/Users/wesbillman/go/pkg/mod/golang.org/x/sync@v0.8.0/errgroup/errgroup.go:78 +0x54
created by golang.org/x/sync/errgroup.(*Group).Go in goroutine 1
	/Users/wesbillman/go/pkg/mod/golang.org/x/sync@v0.8.0/errgroup/errgroup.go:75 +0x98
```

PR #2810, opened 2024-09-24
fix: panic on shutdown

stuartwdouglas:
Also some cleanup around shutdown handling

fixes: #2807

### DIFF
--- a/internal/terminal/noop.go
+++ b/internal/terminal/noop.go
@@ -1,4 +1,4 @@
-package console
+package terminal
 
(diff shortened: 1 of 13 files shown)


internal/terminal/noop.go
```diff
@@ -1,4 +1,4 @@
-package console
+package terminal
```
wesbillman, line 1: Good idea!

alecthomas (approved):

PR #2820, opened 2024-09-25
feat: update terminal status indicators

wesbillman:
Another option for the progress/complete indicators on the interactive terminal.

No worries if the other is preferred, I was mostly playing around to see what we could do here.

https://github.com/user-attachments/assets/d2fc028a-e9b2-4087-b306-28ca0ab82817

### DIFF
--- a/internal/terminal/status.go
+++ b/internal/terminal/status.go
@@ -41,7 +41,7 @@ var _ StatusLine = &terminalStatusLine{}
 var buildColors map[BuildState]string
 var buildStateIcon map[BuildState]func(int) string
 
-var spinner = []string{"▁", "▃", "▄", "▅", "▆", "▇", "█", "▇", "▆", "▅", "▄", "▃"}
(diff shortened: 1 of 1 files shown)


internal/terminal/status.go
```diff
@@ -41,7 +41,7 @@ var _ StatusLine = &terminalStatusLine{}
 var buildColors map[BuildState]string
 var buildStateIcon map[BuildState]func(int) string
 
-var spinner = []string{"▁", "▃", "▄", "▅", "▆", "▇", "█", "▇", "▆", "▅", "▄", "▃"}
+var spinner = []string{"◜", "◝", "◞", "◟"}
```
wesbillman, line 44: Another option might be `var spinner = []string{"|", "/", "-", "\\"}`

stuartwdouglas:
Brail block spinners are another option, as they should render consistently

Screenshot 2024-09-26 at 05 45 25

stuartwdouglas:
I think I like this one the best

PR #2825, opened 2024-09-25
feat: send ingress requests via http

### LINKED ISSUE #2720: Make console VerbForm send http requests for ingress endpoints
We're currently bypassing ingress and calling the verb directly which is confusing and doesn't log the correct events/errors.

### LINKED ISSUE #2753: Add "Copy as FTL command" button to Console requests
A button on the Console verb call page that takes the current request and body, and copies `ftl call <module>.<verb> <body>` to the clipboard

wesbillman:
Fixes #2720
Fixes #2753 (will add similar functionality to request/events page in the future)

With this change we can now see `ingress` events are part of the request since we're using fetch.

Screenshot 2024-09-25 at 1 06 11 PM

Copy `curl`
Screenshot 2024-09-25 at 1 06 36 PM

Copy `ftl.call`
Screenshot 2024-09-25 at 1 06 42 PM

### DIFF
--- a/frontend/console/src/providers/app-providers.tsx
+++ b/frontend/console/src/providers/app-providers.tsx
@@ -1,3 +1,4 @@
+import { Notification } from '../layout/Notification'
 import { NotificationsProvider } from './notifications-provider'
 import { ReactQueryProvider } from './react-query-provider'
 import { RoutingProvider } from './routing-provider'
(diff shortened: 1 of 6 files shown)


frontend/console/src/providers/app-providers.tsx
```diff
@@ -9,6 +10,7 @@ export const AppProvider = () => {
       <UserPreferencesProvider>
         <NotificationsProvider>
           <RoutingProvider />
+          <Notification />
```
wesbillman, line 13: This fixes notifications :)

AlexSzlavik (approved):

PR #2860, opened 2024-09-27
feat: timeline events for scheduled cron jobs

safeer:
Adds an event to the timeline when a cron job is scheduled. No other cron events are needed since the rest will be covered by async events.

Screenshot 2024-10-02 at 10 51 45 AM


backend/controller/timeline/events_cron.go
```diff
@@ -0,0 +1,70 @@
...
+func (e *CronScheduledEvent) GetID() int64 { return e.ID }
+func (e *CronScheduledEvent) event()       {}
+
+type eventCronScheduledJSON struct {
+	ScheduledAt time.Time               `json:"scheduled_at"`
+	Schedule    string                  `json:"schedule"`
+	Error       optional.Option[string] `json:"error,omitempty"`
+}
+
+func (s *Service) InsertCronScheduledEvent(ctx context.Context, event *CronScheduledEvent) {
```
wesbillman, line 38: It's a bit of duplication, but we'll probably want a different type for inserting than returning since the inserted type doesn't need an `ID` an easy way around this it so use 2 types like this:

```go

type CronScheduledEvent struct {
	ID                  int64
	CronScheduled
}

type CronScheduled struct {
	DeploymentKey model.DeploymentKey
	Verb          schema.Ref

	Time        time.Time
	ScheduledAt time.Time
	Schedule    string
	Error       optional.Option[string]
}

File: backend/controller/sql/schema/20240926231955_timeline_async_call_event_types.sql [hunk 2]
```diff
@@ -0,0 +1,6 @@
+-- migrate:up
+
+ALTER TYPE event_type ADD VALUE IF NOT EXISTS 'cron_scheduled';
```
wesbillman, line 3: let's chat about what these events will have. One option would be to have a `cron` event type and use the custom fields and payload for the `scheduled` part. This works assuming the payloads of the different cron types would be the same or similar. If each `cron` event has significantly different payloads, then `cron_scheduled` makes sense.

backend/controller/cronjobs/cronjobs.go
```diff
@@ -206,5 +230,12 @@ func (s *Service) scheduleCronJob(ctx context.Context, tx *dal.DAL, job model.Cr
 	if err != nil {
 		return fmt.Errorf("failed to update cron job %q: %w", job.Key, err)
 	}
```
wesbillman, line 232: Do we need to have an event in this error case as well?

backend/controller/cronjobs/cronjobs.go
```diff
@@ -115,8 +120,16 @@ func (s *Service) scheduleCronJobs(ctx context.Context) (err error) {
...
+		err = s.scheduleCronJob(ctx, tx, job, now)
 		if err != nil {
+			s.timelineService.EnqueueEvent(ctx, &timeline.CronScheduled{
+				DeploymentKey: job.DeploymentKey,
+				Verb:          job.Verb,
+				StartTime:     now,
+				ScheduledAt:   job.NextExecution,
+				Schedule:      job.Schedule,
+				Error:         optional.Some(err.Error()),
+			})
```
wesbillman, line 132: Would it make sense to have these `EnqueueEvent` calls inside of the `scheduleCronJob` func?

backend/controller/timeline/events_cron.go
```diff
@@ -0,0 +1,82 @@
...
+type CronScheduledEvent struct {
+	ID            int64
+	DeploymentKey model.DeploymentKey
+	Verb          schema.Ref
+
+	Time        time.Time
+	Duration    time.Duration
+	ScheduledAt time.Time
+	Schedule    string
+	Error       optional.Option[string]
```
wesbillman, line 27: I think you can embed `CronScheduled` here so you don't have to duplicate these fields.

```go
type CronScheduledEvent struct {
	ID            int64
	CronScheduled
}
```

Assuming they are the same set of fields.

frontend/console/src/features/timeline/TimelineCronScheduled.tsx
```diff
@@ -0,0 +1,17 @@
+import type { CronScheduledEvent } from '../../protos/xyz/block/ftl/v1/console/console_pb'
+import { formatTimestampShort } from '../../utils/date.utils.ts'
+import { verbRefString } from '../verbs/verb.utils'
+
+export const TimelineCronScheduled = ({ cron }: { cron: CronScheduledEvent }) => {
+  const title = `Cron (${cron.schedule}) scheduled at ${cron.scheduledAt}`
+  return (
+    <span title={title}>
+      {'Cron '}
+      <span className='text-indigo-500 dark:text-indigo-300'>{cron.schedule} </span>
```
deniseli, line 10: Since you have the same className defined on each of these 3 child spans, you could bubble that up to the wrapper span instead. i.e. `<span title={title}  className='text-indigo-500 dark:text-indigo-300'>` and no `className` attr below. Helps reduce duplication

frontend/console/src/features/timeline/TimelineCronScheduled.tsx
```diff
@@ -0,0 +1,17 @@
+import type { CronScheduledEvent } from '../../protos/xyz/block/ftl/v1/console/console_pb'
+import { formatTimestampShort } from '../../utils/date.utils.ts'
+import { verbRefString } from '../verbs/verb.utils'
+
+export const TimelineCronScheduled = ({ cron }: { cron: CronScheduledEvent }) => {
+  const title = `Cron (${cron.schedule}) scheduled at ${cron.scheduledAt}`
+  return (
+    <span title={title}>
```
deniseli, line 8: uber-nit: we can do 
```<span title={`Cron (${cron.schedule}) scheduled at ${cron.scheduledAt}`}>``` since it's short anyway and we aren't doing any funky string manipulation that would be worth pulling into a separate line

File: frontend/console/src/features/timeline/TimelineIcon.tsx [hunk 8]
```diff
@@ -18,9 +18,10 @@ export const TimelineIcon = ({ event }: { event: Event }) => {
         return <Rocket01Icon className={`${style} ${textColor}`} />
       case 'log':
         return <LogLevelBadgeSmall logLevel={event.entry.value.logLevel} />
-      case 'ingress': {
+      case 'ingress':
         return <PackageReceiveIcon className={`${style} ${textColor}`} />
-      }
+      case 'cronScheduled':
+        return <AlarmClockIcon className={`${style} ${textColor}`} />
```
wesbillman, line 24: I think I like the alarm clock icon, but also maybe a clock icon would be cool?

https://hugeicons.com/icon/clock-01-stroke-rounded
or maybe
https://hugeicons.com/icon/clock-05-stroke-rounded

wesbillman: we should also update `eventTextColor(event)` and set a color for `cron` events

frontend/console/src/features/timeline/TimelineCronScheduled.tsx
```diff
@@ -0,0 +1,17 @@
...
+      {'Cron '}
+      <span className='text-indigo-500 dark:text-indigo-300'>{cron.schedule} </span>
+      {'verb '}
+      {cron.verbRef?.module && <span className='text-indigo-500 dark:text-indigo-300'>{verbRefString(cron.verbRef)} </span>}
```
wesbillman, line 12: there seems to be whitespace before the ` </span>` on these lines. I guess it's something the linter does't catch/fix for us 🤔

frontend/console/src/features/timeline/TimelineCronScheduled.tsx
```diff
@@ -0,0 +1,17 @@
+import type { CronScheduledEvent } from '../../protos/xyz/block/ftl/v1/console/console_pb'
+import { formatTimestampShort } from '../../utils/date.utils.ts'
+import { verbRefString } from '../verbs/verb.utils'
+
+export const TimelineCronScheduled = ({ cron }: { cron: CronScheduledEvent }) => {
```
wesbillman, line 5: We can do it in a follow on PR, but we'll also want and need a `TimelineCronDetails.tsx` component for the right side bar.

And we'll need to update the verb page event list to show these correctly (seems like it has the `call` icon).
Screenshot 2024-10-01 at 3 36 15 PM

wesbillman (approved):

frontend/console/src/features/timeline/TimelineCronScheduled.tsx
safeer: Thanks @deniseli. In this case, the schedule, verb ref, and time are colored differently from "Cron {} verb {} scheduled for", and the span would format all the text the same. I put a screenshot in the pr description.

backend/controller/timeline/events_cron.go
safeer: This embeds the `CronScheduled` struct under the `CronScheduled` key though, right?

backend/controller/cronjobs/cronjobs.go
safeer: Yeah guess I could. It'll just be 8 places that call `EnqueueEvent` instead of 2

wesbillman: ah ok! this is cool if it's best 👍 I wasn't quite sure on that one :)

frontend/console/src/features/timeline/TimelineCronScheduled.tsx
deniseli: Ahhh gotcha! The screenshot is awesome, thanks for adding that :)

backend/controller/timeline/events_cron.go
safeer: Nvm, ignore me; that's true for initialization but not access

wesbillman (approved):

Issue #2875, opened 2024-09-27
Create new decl specific endpoints and wire them to the console decl pages


wesbillman:
We currently build the entire module tree and pages from either `PullSchema` or `GetModules` we should create dedicated types and requests for our console needs as part of our console BFF.

We already have a few types like this, so we'll likely just need to add the additional decl types we show on the console.

```proto
message Module {
  string name = 1;
  string deployment_key = 2;
  string language = 3;
  string schema = 4;
  repeated Verb verbs = 5;
  repeated Data data = 6;
  repeated Secret secrets = 7;
  repeated Config configs = 8;
}

message Verb {
  schema.Verb verb = 1;
  string schema = 2;
  string json_request_schema = 3;
}

message Data {
  schema.Data data = 1;
  string schema = 2;
}

message Secret {
  schema.Secret secret = 1;
}

message Config {
  schema.Config config = 1;
}

// Add other decls
```

We'll also want dedicated requests for each of these:

```proto
service ConsoleService {
  // some existing stuff...

  rpc GetModule(GetModuleRequest) returns (Module);
  rpc GetVerb(GetVerbRequest) returns (Verb);
  rpc GetData(GetDataRequest) returns (Data);
  rpc GetSecret(GetSecretRequest) returns (Secret);
  rpc GetConfig(GetConfigRequest) returns (Config);

// Add other decls
}
```

Then we can create dedicated `use<Delc>()` hooks in our `api` package to be used by the `<Decl>Page.tsx` components.

deniseli:
reminder: test loading time / impact with `GetVerb`

deniseli:
not completed, just change a plans :)

PR #2882, opened 2024-09-27
feat: add ConsoleService endpoints to get data for each decl type

deniseli:
Part 1 of https://github.com/TBD54566975/ftl/issues/2875

This adds:
* Endpoints for each decl type
* `use<decltype>` helpers for each type
* use `useVerb` on `VerbPage` as proof of concept

This does not add:
* calling any other `use<decltype>` helpers from decl panels
* `GetModule`


alecthomas (commented):
Don't you already have the entire schema in memory from GetSchema()? Why would you need a separate RPC call for each type when you have that?

PR #2905, opened 2024-09-30
wip: node identity part 2

gak:
continuation from https://github.com/TBD54566975/ftl/pull/2771 for #2595


internal/identity/identity.go
```diff
@@ -0,0 +1,199 @@
...
+	return fmt.Sprintf("%s:%s", r.Key, r.Module)
+}
+
+func Parse(s string) (Identity, error) {
+	if s == "" {
+		return nil, fmt.Errorf("empty identity")
+	}
+	parts := strings.Split(s, ":")
+
+	if parts[0] == "ca" {
```
alecthomas, line 57: Might be good to have:
```suggestion
	if parts[0] == "ca" && len(parts) == 1 {
```

internal/identity/identity.go
```diff
@@ -0,0 +1,199 @@
...
+		return Controller{}, nil
+	}
+
+	key, err := model.ParseRunnerKey(parts[0])
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse key: %w", err)
+	}
+
+	return NewRunner(key, parts[1]), nil
```
alecthomas, line 66: You'll need to check that `len(parts) > 1` before this.

internal/identity/identity.go
```diff
@@ -0,0 +1,199 @@
...
+
+	"github.com/alecthomas/types/optional"
+	"google.golang.org/protobuf/proto"
+
+	v1 "github.com/TBD54566975/ftl/backend/protos/xyz/block/ftl/v1"
+	"github.com/TBD54566975/ftl/internal/model"
+)
+
+type Identity interface {
+	String() string
```
alecthomas, line 16: Include a private method here to differentiate Identity from any other type that might happen to implement String().

internal/identity/identity.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	return &Store{
+		Identity: identity,
+		KeyPair:  pair,
+		Signer:   signer,
+	}, nil
+}
+
+func (s *Store) NewGetCertificateRequest() (v1.GetCertificationRequest, error) {
```
alecthomas, line 96: Why Certificate vs. Certification here?

internal/identity/sign.go
```diff
@@ -0,0 +1,132 @@
+package identity
```
alecthomas, line 1: Tests?

alecthomas (approved):

internal/identity/identity.go
gak: Trying to be fancy, I will make it more consistent!

internal/identity/sign.go
gak: Next PR!

internal/identity/identity.go
```diff
@@ -0,0 +1,199 @@
...
+// Runner identity
+// TODO: Maybe use KeyType[T any, TP keyPayloadConstraint[T]]?
+type Runner struct {
+	Key    model.RunnerKey
+	Module string
+}
+
+func NewRunner(key model.RunnerKey, module string) Runner {
+	return Runner{
+		Key:    key,
```
AlexSzlavik, line 42: This is the FTL "key", as in the unique workload identifier at from a platform perspective. I think this constitutes of the host, port and a nonce.

internal/identity/identity.go
```diff
@@ -0,0 +1,199 @@
...
+	key, err := model.ParseRunnerKey(parts[0])
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse key: %w", err)
+	}
+
+	return NewRunner(key, parts[1]), nil
+}
+
+// Store is held by a node and contains the node's identity, key pair, signer, and certificate.
+type Store struct {
```
AlexSzlavik, line 70: We could also call this an identity wallet (for terms borrowed from our web5 colleagues)

internal/identity/identity.go
```diff
@@ -0,0 +1,199 @@
...
+	if s == "" {
+		return nil, fmt.Errorf("empty identity")
+	}
+	parts := strings.Split(s, ":")
+
+	if parts[0] == "ca" && len(parts) == 1 {
+		return Controller{}, nil
+	}
+
+	key, err := model.ParseRunnerKey(parts[0])
```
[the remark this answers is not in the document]
alecthomas: model.RunnerKey is the identity used internally by FTL to identify runners. This is just propagating that existing identity.

internal/identity/api.go
```diff
@@ -1,21 +1,51 @@
...
+	_, err := verifier.Verify(signedData)
+	if err != nil {
+		return SignedData{}, fmt.Errorf("failed to verify data: %w", err)
+	}
+
+	return signedData, nil
+}
+
+type PublicKey struct {
+	Bytes []byte
```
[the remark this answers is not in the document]
gak: That was left over when I abstracted these types. It will probably go away!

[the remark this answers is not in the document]
gak: The base64 encoding is only for the controller public key to be put in the env so the runner can know what to trust.

PR #2907, opened 2024-09-30
fix: remove module from status line

### LINKED ISSUE #2900: Removed modules do not get removed from CLI status section
Steps to repro:
- `ftl dev --recreate`
- Wait for time and echo to get deployed
- Remove echo (by renaming `examples/go/ftl.toml`)

Expected behaviour: 
- echo would be removed from the bottom status section in the CLI, or marked in some way.

Actual behaviour:
- echo remains with a green tick

stuartwdouglas:
fixes #2900

### DIFF
--- a/internal/terminal/status.go
+++ b/internal/terminal/status.go
@@ -30,6 +30,7 @@ const BuildStateBuilt BuildState = "Built"
 const BuildStateDeploying BuildState = "Deploying"
 const BuildStateDeployed BuildState = "Deployed"
 const BuildStateFailed BuildState = "Failed"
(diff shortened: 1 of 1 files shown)


wesbillman (approved):

matt2e (commented):
@stuartwdouglas are we missing the code passing the new state in?

stuartwdouglas:
Ugh, I had this locally, looks like the second `git push` didn't actually work because of a transient failure in the pre-push hook, and I didn't notice. Proper fix is:  https://github.com/TBD54566975/ftl/pull/2916

PR #2912, opened 2024-09-30
fix: don't cache SNAPSHOT versions

stuartwdouglas:
I am not 100% sure this will work, but the docs seem to imply it will.

### DIFF
--- a/.github/actions/build-cache/action.yml
+++ b/.github/actions/build-cache/action.yml
@@ -26,7 +26,9 @@ runs:
       id: cache-maven
       uses: actions/cache/restore@v4
       with:
-        path: ~/.m2/repository
+        path: |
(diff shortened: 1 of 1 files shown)


jvmakine (approved):
No idea what is happening here, but how bad can a 3 line change be!

stuartwdouglas:
So when you build FTL JVM `mvn install` will install those artifacts in the `.m2/repository`, which means they get written back into the cache. If you don't explicitly build Java in your test you can end up using the Java artifacts from another PR.

stuartwdouglas:
This also means that if you forget to build Java your tests might still pass using the Java artifacts from another PR, but then when the SNAPSHOT artifacts expire the next day things will start failing, which is what happened with `TestSmokeTest`

PR #2921, opened 2024-10-01
feat: dump kube pod logs on test failure

stuartwdouglas:


internal/integration/wrapper.go
```diff
@@ -10,28 +12,35 @@ type TestingError string
 // T wraps testing.TB, trapping calls to Fatalf et al and panicking. The panics are caught by
 type T struct {
 	testing.TB
+	tc TestContext
 }
 
 func (t T) Fatal(args ...interface{}) {
-	panic(TestingError(fmt.Sprint(args...)))
+	logs := t.tc.maybeDumpKubePods()
```
stuartwdouglas, line 19: I don't really like this, but I have not been able to find any other way of getting the dump to actually show up in the failure logs.

PR #2930, opened 2024-10-01
feat: add cron go+java examples

safeer:


examples/go/echo/go.mod
```diff
@@ -1,6 +1,6 @@
 module ftl/echo
 
-go 1.23.0
+go 1.23.1
```
stuartwdouglas, line 3: Why this change?

stuartwdouglas: It seems like almost all our tests and examples are on 1.23.0, if we are going to upgrade maybe we should do it all in one hit, although I don't know if there is any real reason to.

stuartwdouglas (approved):

examples/go/echo/go.mod
safeer: This was just to address a build error, but happy to leave as is:
```
info: Modules waiting to build: echo
error: initial deploy failed: failed to build module: failed to build module "echo": err: exit status 1: stderr: go: module ../../../.ftl/go/modules/cron listed in go.work file requires go >= 1.23.1, but go.work lists go 1.23.0; to update it:
	go work use
```

Issue #2805, opened 2024-09-24
Set up a new console endpoint to eventually serve both `useModules` and `useSchema` console use cases


deniseli:
Currently, we have two separate endpoints, one on the console service and one on the controller service, that both provide _most_ of the data we need. We can't standardize to just one of those two because they're each missing data that only the other contains. We should just build a new endpoint that actually includes everything, formatted correctly, so we don't need to do this going forward.

Also, the new one needs to include the `builtin` module.

wesbillman:
We'll probably want a dedicated `StreamModules` rpc in console service so we can live update the tree and the rest of the app as needed.

That might look something like this, and would replace `PullSchema` (which lives on the `controller` service now).
```proto
rpc StreamModules(StreamModulesRequest) returns (stream StreamModulesResponse);
```

The response here would be lighter weight than the `PullSchema` since we only really need enough data to build the module tree and provide data to dedicated decl pages. This will also let us control what the `Module` looks like to the frontend.

PR #2931, opened 2024-10-01
feat: add StreamModules to ConsoleService

deniseli:
Part 1 of https://github.com/TBD54566975/ftl/issues/2805

Next:
1. Client side integration for this PR's version of StreamModules, which behaves in a way that's very similar to the existing `get` endpoint
2. Address the TODO in StreamModules: detect + push deployment updates to the stream
3. Handle streaming updates on the client side


frontend/console/src/api/modules/use-stream-modules.ts
```diff
@@ -0,0 +1,46 @@
+import { Code, ConnectError } from '@connectrpc/connect'
+import { useQuery, useQueryClient } from '@tanstack/react-query'
+import { useClient } from '../../hooks/use-client'
+import { useVisibility } from '../../hooks/use-visibility'
+import { ConsoleService } from '../../protos/xyz/block/ftl/v1/console/console_connect'
+import type { Module } from '../../protos/xyz/block/ftl/v1/console/console_pb'
+
+const streamModulesKey = 'streamModules'
+const updateIntervalMs = 60 * 1000
```
wesbillman, line 9: Should we pass this in and default it? Also, is this the correct number of nano seconds (I might have had this wrong before too 😂)?

Math, mostly for me...
1 nano = 1,000,000 milliseconds
```
60 * 1000 * 1000 = 60,000,000
```
So I think this is updating at `60ms` which might be faster than we want.

frontend/console/src/api/modules/use-stream-modules.ts
```diff
@@ -0,0 +1,46 @@
...
+          const newModuleNames = response.modules.map((m) => m.name)
+          queryClient.setQueryData<Module[]>(queryKey, (prev = []) => {
+            return [...response.modules, ...prev.filter((m) => !newModuleNames.includes(m.name))]
+          })
+        }
+      }
+    } catch (error) {
+      if (error instanceof ConnectError) {
+        if (error.code !== Code.Canceled) {
+          console.error('Console service - streamEvents - Connect error:', error)
```
wesbillman, line 33: `streamEvents` -> `streamModules`

wesbillman (approved):

frontend/console/src/api/modules/use-stream-modules.ts
deniseli: whoopsie!

backend/controller/console/console.go
```diff
@@ -188,6 +153,229 @@ func (c *ConsoleService) GetModules(ctx context.Context, req *connect.Request[pb
...
+		Verbs:         verbs,
+	}, nil
+}
+
+func configFromDecl(decl *schema.Config) *pbconsole.Config {
+	//nolint:forcetypeassert
+	config := decl.ToProto().(*schemapb.Config)
+	return &pbconsole.Config{
+		Config: config,
+	}
```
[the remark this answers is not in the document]
deniseli: great idea - done! and the linter didn't even complain :)

frontend/console/src/api/modules/use-stream-modules.ts
deniseli: Oof thanks for catching that nano issue!

In retrospect, I'm going to just remove this and leave it totally controlled by the server side. It seems like something we'll be able to optimize a lot more cleanly without having to _also_ conform to potentially several distinct client-side refresh thresholds...

backend/controller/console/console.go
```diff
@@ -188,6 +153,229 @@ func (c *ConsoleService) GetModules(ctx context.Context, req *connect.Request[pb
...
+
+	module.Name = deployment.Module
+	module.DeploymentKey = deployment.Key.String()
+	module.Language = deployment.Language
+	module.Schema = deployment.Schema.String()
+
+	return module, nil
+}
+
+func moduleFromDecls(decls []schema.Decl, sch *schema.Schema) (*pbconsole.Module, error) {
```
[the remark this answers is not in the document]
deniseli: Done! I only added an integration test for now and didn't test for the contents of the various `<blahblah>FromDecl` methods since we're going to change so much of that anyway.

PR #2932, opened 2024-10-01
refactor: remove ftl.Call

worstell:
- removes `call.go` from `go-runtime` library
- removes old generated verb stubs
- updates all usages of `ftl.Call` to the new pattern
- adds build error when direct verb invocations are detected


go-runtime/schema/common/fact.go
```diff
@@ -106,6 +106,8 @@ func (*ExternalType) schemaFactValue() {}
 type FunctionCall struct {
 	// The function being called.
 	Callee types.Object
+	// Position where the call takes place.
+	Position schema.Position
```
deniseli, line 110: Should we call this `callerPos`? It's sweet that we'll be able to make stack traces with this now!

deniseli (approved):

go-runtime/schema/common/fact.go
worstell: this is only used for erroring at the moment, and the `Position` in this case refers to the position of the function call itself- it's currently used to surface a build error at the position of a direct verb invocation (which is disallowed in favor of using the generated clients)

deniseli: Ah, gotcha

PR #2945, opened 2024-10-02
refactor: move build errors out of schema

matt2e:
This is in preparation for a full `language.proto` spec for language plugins.
Changes:
- Adds `language.proto` with only error specs
- Adds `builderrors` package to separate build errors out of `schema`
- Use `builderrors.Error` values directly, rather than pointers
- `builderrors.Position` has `StartColumn` and `EndColumn`, rather than `builderrors.Error` having `schema.Pos` and `EndColumn`.
    - Convert to an error position by using `schema.Pos.ToErrorPos()` or `schema.Pos.ToErrorPosWithEnd(column)`
- `Error.Pos.String()` skips the end column if its the same as the start column. `25:7-7` -> `25:7`


jvmakine (approved):

stuartwdouglas (approved):

stuartwdouglas:
I don't think the `skip-proto-breaking` label takes effect until there is a change to the PR.

Issue #2418, opened 2024-08-18
Capture comments in Kotlin for the schema


stuartwdouglas:
This will require hooking into the Kotlin compilation process as they are not preserved in the bytecode.

PR #2946, opened 2024-10-02
feat: capture Kotlin comments

stuartwdouglas:
fixes: #2418


jvm-runtime/ftl-runtime/kotlin/build-parent/pom.xml
```diff
@@ -28,6 +31,26 @@
...
+                            <sourceDirs>
+                                <sourceDir>src/main/kotlin</sourceDir>
+                                <sourceDir>src/main/java</sourceDir>
+                            </sourceDirs>
+                            <annotationProcessorPaths>
+                                <!-- Specify your annotation processors here. -->
+                                <annotationProcessorPath>
+                                    <groupId>xyz.block.ftl</groupId>
+                                    <artifactId>ftl-java-runtime</artifactId>
+                                    <version>${project.version}</version>
```
tomdaffurn, line 49: Is this running AnnotationProcessor in the Kotlin build?

tomdaffurn (approved):

jvm-runtime/ftl-runtime/kotlin/build-parent/pom.xml
stuartwdouglas: Yes, it turns out that this is supported. Kotlin 2.0 is bringing in a new approach but it seems a bit immature at the moment. 

Thinking about it I guess I should move the processor to common.

PR #2974, opened 2024-10-03
fix: don't retry indefinitly with local scaling

stuartwdouglas:
If the runner keeps crashing dont keep restarting it

### DIFF
--- a/backend/controller/scaling/localscaling/local_scaling.go
+++ b/backend/controller/scaling/localscaling/local_scaling.go
@@ -25,6 +25,8 @@ import (
 
 var _ scaling.RunnerScaling = &localScaling{}
 
+const maxExits = 10
+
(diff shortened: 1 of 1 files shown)


docs/content/docs/reference/secretsconfig.md
```diff
@@ -17,6 +17,11 @@ top = false
 
 Configuration values are named, typed values. They are managed by the `ftl config` command-line.
 
+{% code_selector() %}
```
jvmakine, line 20: Is this intentional?

jvmakine (approved):

docs/content/docs/reference/secretsconfig.md
stuartwdouglas: oops

PR #2986, opened 2024-10-03
fix: verb for resetting after submit

### LINKED ISSUE #2975: Console forgets verb body after send
When a verb is invoked from the console, the body text is immediately wiped and replaced with the previous version from local storage.

1. Load echo.echo. Body is '{}'
2. Enter `{"name":"Tom"}`
3. Click 'Send'
-> Body is wiped and reset to '{}'

You can also enter some body text, reload the page (which stores it in local storage). Then that version is what replaces anything you enter.

Screenshot 2024-10-03 at 3 23 17 PM

wesbillman:
- Fixes #2975 
- Adds back in the example request body instead of empty `{}`
- Adds a `reset` button to get reset to the default body

Screenshot 2024-10-03 at 3 38 12 PM
Screenshot 2024-10-03 at 3 38 19 PM


deniseli (approved):
Niiiiiiice! Very impressed you got the Reset button working on that component :D

tomdaffurn:
wooooop! Thanks Wes

PR #2995, opened 2024-10-04
refactor: move verb export check to before req body validation

mistermoe:
# Summary
This smol PR includes a micro optimization by ensuring that a verb is callable (a.k.a exported OR a call within the same module) _before_ validating the request body. 

# Rationale
This change is not necessary by any means. Just something that crossed my mindwhile looking at controller logic with @wesbillman. 

# Changes
* moves `!verb.IsExported` check to before validating request body
* use the current caller to see if the call is intra-module instead of looping over all callers. 

> [!NOTE]
> there may have been a reason why all callers are being looped over instead of checking the current caller, but i couldn't think of a good one. if someone knows definitely let me know!

### DIFF
--- a/backend/controller/controller.go
+++ b/backend/controller/controller.go
@@ -1018,12 +1018,13 @@ func (s *Service) callWithRequest(
 		return nil, err
 	}
 
-	var currentCaller *schema.Ref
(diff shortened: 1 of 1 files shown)


wesbillman (approved):
Nice @mistermoe!

Thanks for cleaning this up :)

alecthomas (approved):

PR #3001, opened 2024-10-04
fix: enum syntax highlighting

### LINKED ISSUE #2997: Console syntax highlighting missing last enum character
The last character of an enum name is not highlighted properly in Console schema page:

image

wesbillman:
Fixes #2997

Also fixes refreshing of modules when schema changes

Screenshot 2024-10-04 at 9 07 40 AM
Screenshot 2024-10-04 at 10 04 30 AM

### DIFF
--- a/frontend/console/src/api/modules/use-modules.ts
+++ b/frontend/console/src/api/modules/use-modules.ts
@@ -1,6 +1,5 @@
 import { Code, ConnectError } from '@connectrpc/connect'
-import { useQuery, useQueryClient } from '@tanstack/react-query'
-import { useEffect } from 'react'
+import { useQuery } from '@tanstack/react-query'
 import { useClient } from '../../hooks/use-client'
 import { ConsoleService } from '../../protos/xyz/block/ftl/v1/console/console_connect'
 import { useSchema } from '../schema/use-schema'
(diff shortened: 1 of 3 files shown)


frontend/console/src/api/modules/use-modules.ts
```diff
@@ -38,7 +28,7 @@ export const useModules = () => {
   }
 
   return useQuery({
-    queryKey: [useModulesKey],
+    queryKey: [useModulesKey, schemaUpdatedAt],
```
wesbillman, line 31: without this, the schema would change (delete old modules) and never trigger the modules to reload. This should be better once we have a dedicated streaming endpoint :)

Also note that with the current deployment/schema stuff, it's possible to have 2 active deployments for the same module for a few seconds while the old module gets cleaned up. We'll want to add handling to our streaming endpoint to return the most recently activated deployment vs. returning both the old and new.

In the future, this will need more attention since we'll likely want to show multiple deployed versions, but until we have versioning worked out we should probably just show the most recent active deployment.

jonathanj-square (approved):
looks good

frontend/console/src/api/modules/use-modules.ts
deniseli: Nice, that makes sense. Thanks for catching that! The streaming endpoint does already filter for unique module names and chooses the latest when there is a dup. I haven't added update handling to the backend yet, though, so it's not yet tested.

PR #3016, opened 2024-10-07
feat: separate out language logic from moduleconfig

matt2e:
### Why
Language support was previously tightly integrated with how a `ModuleConfig` is created. Without this defaulting, FTL does not know key info about a module like deploy directory or where to find the schema.

Another complication is that module defaults for jvm modules is done by looking for certain files to automatically choose defaults for gradle or maven, so we can't get the defaults once per language and use them for all modules of that language.

### Creating a ModuleConfig
To create a `ModuleConfig`, the steps are:
```go
// all error checking is skipped for clarity

// first load the toml file to get a UnvalidatedModuleConfig,
// which holds only the values that were in the toml file.
partialConfig, err := moduleconfig.LoadConfig(dir)

// from this we know the language, and can create a language plugin
plugin, err := languageplugin.New(ctx, uncheckedConfig.Language)

// we then ask the language plugin for the custom defaults
customDefaults, err := plugin.ModuleConfigDefaults(ctx, uncheckedConfig.Dir)

// we can now fill in the defaults and validate the config
config, err := uncheckedConfig.FillDefaultsAndValidate(customDefaults)
```

### Other changes
- Removed `ModuleGoConfig `, `ModuleKotlinConfig`, `ModuleJavaConfig`, replace with `LanguageConfig: map[string]any`
    - To do this, we unmarshal the toml into a `map[string]any` and then parse that into an `UnvalidatedModuleConfig` using `mapstructure` and then pull out the language config from the correct key (`go`, `java`, `kotlin`, etc)
- `buildengine` caches the defaults per module so that each time it reads the toml it does not to ask for CustomDefaults from the plugin.


backend/controller/admin/local_client.go
```diff
@@ -45,19 +47,52 @@ func (s *diskSchemaRetriever) GetActiveSchema(ctx context.Context) (*schema.Sche
...
+
 	for _, m := range modules {
-		config := m.Abs()
-		schemaPath := config.Schema()
-		if r, ok := s.deployRoot.Get(); ok {
-			schemaPath = filepath.Join(r, m.Module, m.DeployDir, m.Schema())
-		}
+		go func() {
+			// Loading a plugin can be expensive. Is there a better way?
+			plugin, err := languageplugin.New(ctx, m.Language)
```
matt2e, line 56: This change slows down this logic a fair bit because now we need to create a language plugin per module just so we can figure out where the schema is saved.
I'm not sure of a good solution to this, other than having a central place FTL saves schemas (similar to the central `/.ftl/` external modules) so that having a ModuleConfig is not readed to read the saved schema for a module.

internal/buildengine/deps.go
```diff
@@ -1,31 +0,0 @@
...
-
-import (
-	"context"
-
-	"github.com/TBD54566975/ftl/internal/log"
-)
-
-// UpdateDependencies finds the dependencies for a module and returns a
-// Module with those dependencies populated.
-func UpdateDependencies(ctx context.Context, module Module, plugin LanguagePlugin) (Module, error) {
```
matt2e, line 11: this was unused after previous language plugin changes

internal/buildengine/deps_test.go
```diff
@@ -1,28 +0,0 @@
...
-
-import (
-	"context"
-	"testing"
-
-	"github.com/TBD54566975/ftl/internal/moduleconfig"
-	"github.com/alecthomas/assert/v2"
-)
-
-func TestExtractModuleDepsGo(t *testing.T) {
```
matt2e, line 11: these have been moved to unit tests within `languageplugin`

internal/buildengine/languageplugin/go_plugin_helpers.go
```diff
@@ -0,0 +1,123 @@
...
+	"go/token"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/TBD54566975/ftl/internal/errors"
+	"golang.org/x/mod/modfile"
+)
+
+func replacementWatches(moduleDir, deployDir string) ([]string, error) {
```
matt2e, line 15: all this code was previously in `moduleconfig` but is all specific to the go plugin

internal/buildengine/languageplugin/java_plugin.go
```diff
@@ -29,19 +30,71 @@ import (
...
 		internalPlugin: internal,
 	}
 }
 
+func (p *javaPlugin) ModuleConfigDefaults(ctx context.Context, dir string) (moduleconfig.CustomDefaults, error) {
+	defaults := moduleconfig.CustomDefaults{
+		GeneratedSchemaDir: "src/main/ftl-module-schema",
+		Deploy:             []string{"launch", "quarkus-app"},
+		// Watch defaults to files related to maven and gradle
+		Watch: []string{"pom.xml", "src/**", "build/generated", "target/generated-sources"},
```
matt2e, line 68: Previously `watch` was different based on maven/gradle, but I think it's fine to just use the same watch patterns for both

internal/watch/discover.go
```diff
@@ -13,8 +13,8 @@ import (
 
 // DiscoverModules recursively loads all modules under the given directories
 // (or if none provided, the current working directory is used).
-func DiscoverModules(ctx context.Context, moduleDirs []string) ([]moduleconfig.ModuleConfig, error) {
-	out := []moduleconfig.ModuleConfig{}
+func DiscoverModules(ctx context.Context, moduleDirs []string) ([]moduleconfig.UnvalidatedModuleConfig, error) {
```
matt2e, line 16: Discover deals with `UnvalidatedModuleConfig`s now

internal/watch/discover_test.go
```diff
@@ -14,162 +14,31 @@ func TestDiscoverModules(t *testing.T) {
 	ctx := log.ContextWithNewDefaultLogger(context.Background())
 	modules, err := DiscoverModules(ctx, []string{"testdata"})
 	assert.NoError(t, err)
-	expected := []moduleconfig.ModuleConfig{
+	expected := []moduleconfig.UnvalidatedModuleConfig{
```
matt2e, line 17: Previously this test was testing discovery and defaulting.
Now it just tests discovery. Generic defaulting tests can now be found in `moduleconfig` tests, while language specific default testing can be found in `languageplugin`

Issue #2982, opened 2024-10-03
console modules tree cannot re-open group containing selected node


mistermoe:
# Repro Steps
1. go to http://localhost:8892/modules
2. open pubsub group
3. select the first pubsub decl
4. collapse the pubsub group
5. try to open it again

https://github.com/user-attachments/assets/bfc4dadb-fcdb-42f9-800c-f464cd279424

PR #3022, opened 2024-10-07
fix: module tree expansion state when closing the module of the selected decl

deniseli:
Fixes: https://github.com/TBD54566975/ftl/issues/2982

This fix delves into the nuance of component re-rendering lifecycle a bit. Basically, before, we would add the selected module to local storage AND update the component state on every single re-render of the component (at `const [expandedModules, setExpandedModules] = useState(...)`. The problem with that was that when we toggled the selected module to close it, the re-render would cause the selected module to still be added back to local storage, but the component state would be out of sync. Then when you went to expand that module again, because the states got unsynced, it would stay stuck in the visibly-closed state. Now, we only add the selected module to local storage in the effect, which doesn't get triggered on toggle a module's expansion state. As a result, the state doesn't de-sync.

So now, you can do the following:
1. Select a decl
2. Close the module of the selected decl in the module tree
3. Re-expand that same module in the module tree. It works.

While the following still works correctly:
1. Close all the modules (i.e. to-be-selected module will not be cached as expanded)
2. Go to a decl page (e.g. paste a link into the browser)
3. The selected module should be expanded in the tree.

### DIFF
--- a/frontend/console/src/features/modules/ModulesTree.tsx
+++ b/frontend/console/src/features/modules/ModulesTree.tsx
@@ -127,13 +127,6 @@ const ModuleSection = ({
   )
 }
 
-function getInitialExpandedModules(moduleName?: string, declName?: string): string[] {
-  if (moduleName && declName) {
-    addModuleToLocalStorageIfMissing(moduleName)
-  }
-  return listExpandedModulesFromLocalStorage()
-}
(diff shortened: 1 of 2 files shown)


wesbillman (approved):
Awesome thanks!!

PR #3028, opened 2024-10-08
feat: Improve local subscriptions

stuartwdouglas:


stuartwdouglas:
@alecthomas I think this should address your type safety concerns from last week. Unfortunately it does not help with same module verb invocation though, I am still not sure how to be make that type safe.

tomdaffurn (approved):

Issue #3037, opened 2024-10-08
Update mangleQueryParameters to convert to underlying schema type instead of always strings


wesbillman:
Using a non-string type (like `int`) in a query param struct will cause an encoding failure and internal server error. For example, it would be nice if we could support an http ingress like this:

```go
type GetQueryParams struct {
	Age ftl.Option[int] `json:"age"`
}

type GetPathParams struct {
	Name string `json:"name"`
}

type GetResponse struct {
	Name string          `json:"name"`
	Age  ftl.Option[int] `json:"age"`
}

// Example usage of path and query params
// curl http://localhost:8891/get/wicket?age=10
//
//ftl:ingress http GET /get/{name}
func Get(ctx context.Context, req builtin.HttpRequest[ftl.Unit, GetPathParams, GetQueryParams]) (builtin.HttpResponse[GetResponse, ApiError], error) {
	return builtin.HttpResponse[GetResponse, ApiError]{
		Status: 200,
		Body: ftl.Some(GetResponse{
			Name: req.PathParameters.Name,
			Age:  req.Query.Age,
		}),
	}, nil
}
```

This will currently fail with:

```sh
error:http:runner0: Call to deployments dpl-http-e12hoaj2xyu13r5 failed: invalid request to verb http.get: json: cannot unmarshal string into Go value of type int
```

PR #3049, opened 2024-10-09
fix: query/path params only allow string

stuartwdouglas:
fixes: #3037


tomdaffurn (approved):

wesbillman (approved):
Awesome! Thanks @stuartwdouglas

PR #3051, opened 2024-10-09
feat: observability support for JVM

stuartwdouglas:
fixes: #3041

### DIFF
--- a/examples/java/time/src/main/java/ftl/time/Time.java
+++ b/examples/java/time/src/main/java/ftl/time/Time.java
@@ -2,14 +2,25 @@ package ftl.time;
 
 import java.time.OffsetDateTime;
 
+import io.opentelemetry.api.metrics.LongCounter;
+import io.opentelemetry.api.metrics.Meter;
 import xyz.block.ftl.Export;
 import xyz.block.ftl.Verb;
 
 public class Time {
+    final LongCounter counter;
+
+    public Time(Meter meter) {
+        counter = meter.counterBuilder("time.invocations")
+                .setDescription("The number of time invocations")
+                .setUnit("invocations")
+                .build();
+    }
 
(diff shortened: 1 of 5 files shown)


tomdaffurn (approved):

jvm-runtime/ftl-runtime/common/runtime/src/main/java/xyz/block/ftl/runtime/VerbHandler.java
```diff
@@ -11,14 +15,21 @@
 public class VerbHandler extends VerbServiceGrpc.VerbServiceImplBase {
 
     final VerbRegistry registry;
+    final LongCounter counter;
 
-    public VerbHandler(VerbRegistry registry) {
+    public VerbHandler(VerbRegistry registry, Meter meter) {
         this.registry = registry;
+        counter = meter.counterBuilder("ftl-jvm-runtime/verb_invocations")
```
jonathanj-square, line 22: this appears to mirror: https://github.com/TBD54566975/ftl/blob/8453735f03dae4795a34ab0019201ddc292aa76b/backend/controller/observability/calls.go#L44

can you use the same naming scheme

jonathanj-square (approved):
looks good overall, minor naming request added

jvm-runtime/ftl-runtime/common/runtime/src/main/java/xyz/block/ftl/runtime/VerbHandler.java
stuartwdouglas: I actually wonder if we need this at all, given that we already track request numbers in the controller, it was more added just to make sure the metrics were working.

I might replace it with some user level metrics in the examples.

PR #3057, opened 2024-10-09
feat: replace all useModules and useStream callers in ModulesPage with useStreamModules

deniseli:
This leaves the legacy stuff alone and just refactors the newer pieces, with a long list of minor bug fixes built in.

https://github.com/user-attachments/assets/07280879-ed11-40ec-ab0d-89a3edd7471b


frontend/console/src/features/modules/module.utils.ts
```diff
@@ -96,6 +126,44 @@ export const declFromSchema = (moduleName: string, declName: string, schema: Pul
...
+
+export const declFromModules = (moduleName: string, declCase: string, declName: string, modules?: Module[]) => {
+  if (!modules) {
+    return undefined
+  }
+  const module = modules.find((m) => m.name === moduleName)
+  if (!module) {
+    return undefined
+  }
+  switch (declCase) {
```
jonathanj-square, line 143: might be nice to reduce the repetition here; would something like this be suitable?

const declMap: Record<string, keyof typeof Module> = {
  config: 'configs',
  data: 'data',
  database: 'databases',
  ...
};

const key = declMap[declCase];
if (key) {
  return module[key].find((id) => d[key]?.name === declName)?.[key];
}

frontend/console/src/api/modules/use-stream-modules.ts
```diff
@@ -24,11 +24,11 @@ export const useStreamModules = () => {
           hasModules = true
           const newModuleNames = response.modules.map((m) => m.name)
           queryClient.setQueryData<Module[]>(queryKey, (prev = []) => {
-            return [...response.modules, ...prev.filter((m) => !newModuleNames.includes(m.name))]
+            return [...response.modules, ...prev.filter((m) => !newModuleNames.includes(m.name))].sort((a, b) => a.name.localeCompare(b.name))
           })
         }
       }
-      return hasModules ? queryClient.getQueryData(queryKey) : []
+      return hasModules ? (queryClient.getQueryData(queryKey) as Module[]) : []
```
wesbillman, line 31: Because of the way the way deployments work now, it's possible to change to a new deployment while the other deployment is being shut down which will result in 2 deployments for the same module being active for a few seconds.

You can see this in the UI here:
Screenshot 2024-10-10 at 9 23 02 AM

That's what this little bit of code in the `use-schema` is for: https://github.com/TBD54566975/ftl/blob/main/frontend/console/src/api/schema/use-schema.ts#L24-L43

frontend/console/src/features/modules/decls/DeclLink.tsx
```diff
@@ -1,33 +1,48 @@
...
+    }
+  }, [containerRect])
 
-const SnippetContainer = ({ decl, linkRect, containerRect }: { decl: DeclSchema; linkRect?: DOMRect; containerRect?: DOMRect }) => {
   const ref = useRef<HTMLDivElement>(null)
   const snipRect = ref?.current?.getBoundingClientRect()
 
   const hasRects = !!snipRect && !!linkRect
-  const toTop = hasRects && window.innerHeight - linkRect.top - linkRect.height < snipRect.height + linkRect.height
+  const fitsAbove = hasRects && linkRect.top - 64 > snipRect.height
```
wesbillman, line 27: What is the `64` here?

frontend/console/src/features/modules/decls/TopicPanel.tsx
```diff
@@ -1,15 +1,18 @@
...
   return (
     <div className='py-2 px-4'>
       <PanelHeader exported={value.export} comments={value.comments}>
         Topic: {moduleName}.{declName}
       </PanelHeader>
-      <div className='text-sm my-4'>
-        Event Type: <TypeEl t={value.event} />
+      <div className='-mx-3.5'>
+        <Schema schema={schema} />
       </div>
```
wesbillman, line 16: Maybe we should remove this `px-4` in the outer `div`s for these and just add padding to the header so we don't need negative margins on the decl pages?

frontend/console/src/features/modules/schema/Schema.tsx
```diff
@@ -6,18 +6,18 @@ import { LinkToken, LinkVerbNameToken } from './LinkTokens'
...
-function maybeRenderDeclName(token: string, declType: string, tokens: string[], i: number, containerRect?: DOMRect) {
+function maybeRenderDeclName(token: string, declType: string, tokens: string[], i: number, moduleName: string, containerRect?: DOMRect) {
   const offset = declType === 'database' ? 4 : 2
   if (i - offset < 0 || declType !== tokens[i - offset]) {
     return
   }
-  if (declType === 'enum') {
-    return [<LinkToken key='l' token={token} containerRect={containerRect} />]
+  if (declType === 'enum' && token.endsWith(':')) {
+    return [<LinkToken key='l' moduleName={moduleName} token={token.slice(0, token.length - 1)} containerRect={containerRect} />, ':']
```
wesbillman, line 15: :)

wesbillman (approved):

frontend/console/src/features/modules/decls/TopicPanel.tsx
deniseli: The negative margin here is to remove the additional tab width that each of the decls has as a part of its schema string, since we're parsing these out of the module schema string. I thought about it a bit and realized the annoying thing is basically, once we add more stuff to these pages, every one of those new things would also need the px-4 to be set in order to match the header, and we'd have that duplication through every element in this page except for Schema. So this makes future development cleaner, even though having the negative margin is gross :/

frontend/console/src/features/modules/decls/DeclLink.tsx
deniseli: Oooo whoops thanks for catching that! This is the height of the topbar. I'll move it to a const. Completely slipped my mind

frontend/console/src/api/modules/use-stream-modules.ts
deniseli: Ahhhh thanks for catching that! I'll do it in this PR so it can all go in cleanly.

deniseli: Done! I confirmed the module schema updates nicely upon redeploy too.

PR #3070, opened 2024-10-10
feat: initial helm charts

stuartwdouglas:
There is still a tiny bit of kustomize around istio, but it is almost there. Once this is working I will work on getting a release process for the charts to make them easy to consume.

The charts themselves were mostly written by @tlongwell-block


alec-brooks (approved):
Looks good.

Dockerfile.initdb
```diff
@@ -0,0 +1,50 @@
...
+
+# Build
+COPY . /src/
+RUN just errtrace
+# Reset timestamps so that the build state is reset
+RUN git ls-files -z | xargs -0 touch -r go.mod
+RUN just build ftl-controller
+
+# Finally create the runtime image.
+FROM scratch
```
AlexSzlavik, line 33: nice

Dockerfile.initdb
```diff
@@ -0,0 +1,50 @@
...
+COPY frontend/vscode/package.json ./frontend/vscode/
+COPY pnpm-workspace.yaml pnpm-lock.yaml ./
+RUN pnpm install --frozen-lockfile
+
+# Build
+COPY . /src/
+RUN just errtrace
+# Reset timestamps so that the build state is reset
+RUN git ls-files -z | xargs -0 touch -r go.mod
+RUN just build ftl-controller
```
AlexSzlavik, line 30: should this `just build ftl-initdb`?

Dockerfile.initdb
```diff
@@ -0,0 +1,50 @@
...
+FROM scratch
+
+COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
+
+WORKDIR /service/
+
+COPY --from=builder /src/build/release/ftl-controller .
+COPY --from=builder /src/build/release/ftl-initdb . #TODO: Remove this line, we need a different migration image, this will get us moving for now though
+
+EXPOSE 8891
```
AlexSzlavik, line 42: Should this image be a one time run, not a persistent image?

Dockerfile.initdb
AlexSzlavik, line 1: This file is suffixed initdb, but I don't see any DB initing here? This looks like a controller docker file.

charts/values.yaml
```diff
@@ -0,0 +1,162 @@
...
+  replicas: 2
+  revisionHistoryLimit: 0
+
+  image:
+    repository: "ftl0/ftl-controller"
+    tag: "v0.361.2"
+
+  envFrom: null
+  dbConnectionString: "postgres://$(endpoint):$(port)/tbd?sslmode=disable&user=$(username)&password=$(password)"
+  kmsUri: "fake-kms://CKbvh_ILElQKSAowdHlwZS5nb29nbGVhcGlzLmNvbS9nb29nbGUuY3J5cHRvLnRpbmsuQWVzR2NtS2V5EhIaEE6tD2yE5AWYOirhmkY-r3sYARABGKbvh_ILIAE"
```
AlexSzlavik, line 27: This will result in the same fake KMS, tink key being used everywhere. This is probably fine, but something we definitely want to document and ensure users are aware of changing (and using a real KMS, key injection abstraction)

Dockerfile.initdb
stuartwdouglas: Ugh, I didn't end up actually needthing this, and I forgot to delete the file. The runner image contains ftl so I run ftl migrate instead of initdb.

PR #3073, opened 2024-10-10
fix: ftltest.Call always allows direct initial verb

### LINKED ISSUE #2984: make `WithCallsAllowedWithinModule()` the default and remove
now that we have injected resources on verbs, users should always be using `ftltest.Call` to invoke verbs in tests rather than invoking the verb function directly. currently, users must also set `WithCallsAllowedWithinModule()` in the test context in order to use `ftltest.Call` on a verb that is not mocked (and get its inline behavior).

it seems like we should just execute inline behavior as the default for `ftltest.Call` on same-module verbs, and overridde with any mocks when present. only external verbs would require mocks

this came up when updating customer code to the new verb client injection patterns + corresponding unit tests. for every unit test, `WithCallsAllowedWithinModule()` had to be set on the context

worstell:
when using ftltest.Call, you don't need to mock the verb directly invoked. only downstream verbs must be mocked (or overridden by setting WithCallsAllowedWithinModule globally)

https://sq-tbd.slack.com/archives/C04PEQERFM0/p1727995042643089

fixes #2984

### DIFF
--- a/go-runtime/ftl/ftltest/ftltest.go
+++ b/go-runtime/ftl/ftltest/ftltest.go
@@ -523,8 +523,13 @@ func CallEmpty[VerbClient any](ctx context.Context) error {
 
 func call[VerbClient, Req, Resp any](ctx context.Context, req Req) (resp Resp, err error) {
 	ref := reflection.ClientRef[VerbClient]()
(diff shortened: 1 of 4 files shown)


matt2e (approved):
Awesome

PR #3074, opened 2024-10-10
fix: JVM support on Kube

stuartwdouglas:


Dockerfile.runner
```diff
@@ -7,10 +7,7 @@ COPY ./bin /src/bin
 ENV PATH="/src/bin:$PATH"
 ENV HERMIT_STATE_DIR=/hermit
 RUN hermit uninstall jbr
-RUN hermit install openjre-18.0.2.1_1
-# openjre and jbr conflict, but we want the JRE in the cache
-RUN hermit uninstall openjre
-RUN hermit install jbr
+RUN hermit install openjdk-17.0.8_7
```
deniseli, line 10: Just curious - are there any good ways to templatize the version number in docker files so you don't have to hardcode the full version number 4 times?

deniseli (approved):

Dockerfile.runner
stuartwdouglas: I could probably do it with a build arg.

safeer (approved):
Nice

Dockerfile.runner
safeer: Can we land as is and follow up on the build arg? I'd love to use this for the smoketest 😁

deniseli: @safeer sgtm :D

PR #3078, opened 2024-10-10
chore: release helm charts

stuartwdouglas:


jvmakine (approved):

.github/workflows/release.yml
```diff
@@ -101,6 +101,10 @@ jobs:
         with:
           username: ftl0
           password: ${{ secrets.FTL_DOCKER_PUSH_TOKEN }}
+      - name: Push Helm Chart
+        run: |
+          version="$(git describe --tags --abbrev=0 | sed s/v//)"
+          just chart release "$version" oci://ftl0/ftl-charts
```
AlexSzlavik, line 107: Very cool

AlexSzlavik: Does this work yet?

stuartwdouglas: I does, the new cluster is already running on these charts.

That said I am going to change where they are hosted, the oci approach has not been ideal.

PR #3088, opened 2024-10-11
chore: move renovate config

stuartwdouglas:
We have a lot of files in the root of the repo, this makes things slightly cleaner


wesbillman (approved):
Oh this is nice! I didn't know you could put this in the `.github` folder! 🥳

Issue #3105, opened 2024-10-14
Logs should include span/trace IDs in JSON


stuartwdouglas:
If we include this info in the correct format then we should be able to correlate log messages to traces in the DD console

PR #3107, opened 2024-10-14
feat: add DD trace id to the logs

stuartwdouglas:
Only works for the go runtime at the moment, as the JVM logging needs to be cleaned up to match the go format.

fixes: #3105

### DIFF
--- a/internal/observability/logger.go
+++ b/internal/observability/logger.go
@@ -1,12 +1,37 @@
 package observability
 
 import (
+	"context"
 	"fmt"
+	"strconv"
 
 	"github.com/go-logr/logr"
 
+	"go.opentelemetry.io/otel/trace"
+
 	"github.com/TBD54566975/ftl/internal/log"
 )
+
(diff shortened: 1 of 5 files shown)


jvmakine (approved):

internal/observability/logger.go
```diff
@@ -1,13 +1,37 @@
...
+}
+
+func AddSpanContextToLogger(ctx context.Context) context.Context {
+	sc := trace.SpanContextFromContext(ctx)
+	return wrapLogger(ctx, sc)
+}
+func wrapLogger(ctx context.Context, sc trace.SpanContext) context.Context {
+	logger := log.FromContext(ctx)
+	attributes := map[string]string{}
+	if sc.HasSpanID() {
```
stuartwdouglas, line 26: I guess an open question here is if we should be hard coding DD specific attributes. I don't think it matters for now, but we may need to revisit in future.

PR #3122, opened 2024-10-15
fix: remove cron jobs on deactivation

### LINKED ISSUE #3109: Cron methods called multiple times for long running deployments
It appears old cron methods are not being deactivated correctly, so on redeploy we end up with multiple calls.

stuartwdouglas:
fixes #3109


stuartwdouglas:
Turns out this needs more thought, it leads to import cycles in other tests.

jvmakine (approved):
Nice!

PR #3137, opened 2024-10-16
fix: Dockerfile.provisioner

jvmakine:


stuartwdouglas:
This conflicts with https://github.com/TBD54566975/ftl/pull/3135

It had some weird istio RBAC errors I am still trying to figure out, as it works fine for me locally

jvmakine:
Ah right, maybe I will wait for that to come through.

PR #3140, opened 2024-10-16
fix: wait for provisioner

stuartwdouglas:


backend/provisioner/service.go
```diff
@@ -27,7 +27,7 @@ type CommonProvisionerConfig struct {
 }
 
 type Config struct {
-	Bind               *url.URL `help:"Socket to bind to." default:"http://127.0.0.1:8893" env:"FTL_PROVISIONER_BIND"`
+	Bind               *url.URL `help:"Socket to bind to." default:"http://127.0.0.1:8894" env:"FTL_PROVISIONER_BIND"`
```
matt2e, line 30: I wonder if we want to use one less than the default controller port...
8893 and can legitimately be taken up by other controllers, runners, language plugins etc.
As in running `ftl build` could prevent the provisioner being able to start which is surprising.

matt2e: (The language plugin stuff is only half in main atm)

matt2e: I've been doing hacky step in a few places to try and not use the controller port in other commands: https://github.com/TBD54566975/ftl/blob/main/frontend/cli/cmd_schema_diff.go#L51

PR #3144, opened 2024-10-17
feat: Move VerbClient annotation to method

tomdaffurn:
- Move @VerbClient annotation from class to method 
- Rename client method from "call" to \<verbName\>

### DIFF
--- a/jvm-runtime/testdata/java/javaclient/src/main/java/xyz/block/ftl/test/TestInvokeGoFromJava.java
+++ b/jvm-runtime/testdata/java/javaclient/src/main/java/xyz/block/ftl/test/TestInvokeGoFromJava.java
@@ -59,106 +59,106 @@ public class TestInvokeGoFromJava {
      */
     @Export
     @Verb
-    public void emptyVerb(EmptyVerbClient emptyVerbClient) {
-        emptyVerbClient.call();
+    public void emptyVerb(EmptyVerbClient client) {
+        client.emptyVerb();
     }
 
     @Export
     @Verb
-    public void sinkVerb(String input, SinkVerbClient sinkVerbClient) {
-        sinkVerbClient.call(input);
+    public void sinkVerb(String input, SinkVerbClient client) {
+        client.sinkVerb(input);
     }
 
     @Export
     @Verb
-    public String sourceVerb(SourceVerbClient sourceVerbClient) {
-        return sourceVerbClient.call();
+    public String sourceVerb(SourceVerbClient client) {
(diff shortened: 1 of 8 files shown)


jvm-runtime/testdata/java/javaclient/src/main/java/xyz/block/ftl/test/TestInvokeGoFromJava.java
```diff
@@ -59,176 +59,176 @@ public class TestInvokeGoFromJava {
...
             OptionalStringMapVerbClient client) {
-        return client.call(val);
+        return client.optionalStringMapVerb(val);
     }
 
     @Export
     @Verb
     public @Nullable ZonedDateTime optionalTimeVerb(@Nullable ZonedDateTime instant, OptionalTimeVerbClient client) {
-        return client.call(instant);
+        return client.optionalTimeVerb(instant);
```
stuartwdouglas, line 212: I wonder if we should still stick with 'call' as the default name. You already have the name in the interface, so I don't think we need to duplicate the name.

stuartwdouglas (approved):

jvm-runtime/testdata/java/javaclient/src/main/java/xyz/block/ftl/test/TestInvokeGoFromJava.java
tomdaffurn: I was actually wondering if we should combine all verbs into a single interface/class, then we get an API like:
```
    @Export
    @Verb
    public @Nullable ZonedDateTime optionalTimeVerb(@Nullable ZonedDateTime instant, GoModule goModule) {
        return goModule.optionalTimeVerb(instant);
    }
```

stuartwdouglas: The issue is then you can't detect usages. I guess if you have a few that are commonly used you could allow a pattern where you have `CombinedInterface extends Client1, Client2`.

stuartwdouglas: Let's leave the names as-is for now, and see how it goes.

tomdaffurn: Yeah, I hadn't considered the usage detection. Worth discussing all this next week

PR #3147, opened 2024-10-17
fix: typealias case issues in icons and urls

wesbillman:
- Fixes issues with `CommandPallete` erroring when `typeAlias` is used since we had `typealias` in the icons list
- Fixes issues with urls that have `module/typealias/declName` in them vs. `module/typeAlias/declName`
- Performance improvements on `CommandPallete` searching/filtering.

See also #3146

### DIFF
--- a/frontend/console/src/features/command-pallete/CommandPalette.tsx
+++ b/frontend/console/src/features/command-pallete/CommandPalette.tsx
@@ -25,20 +25,23 @@ export const CommandPalette: React.FC<CommandPaletteProps> = ({ isOpen, onClose 
     }
   }, [schemaData])
 
-  const fuse = useMemo(() => {
-    return new Fuse(items, {
-      keys: ['title', { name: 'subtitle', weight: 1 }],
-      minMatchCharLength: 2,
-      fieldNormWeight: 2,
-    })
-  }, [items])
+  const fuseOptions = {
+    keys: ['title', { name: 'subtitle', weight: 1 }],
+    minMatchCharLength: 2,
+    fieldNormWeight: 2,
+  }
+
+  const fuseIndex = useMemo(() => Fuse.createIndex(fuseOptions.keys, items), [items])
+  const fuse = useMemo(() => new Fuse(items, fuseOptions, fuseIndex), [items, fuseIndex])
 
   useEffect(() => {
     if (query === '') {
       setFilteredItems([])
     } else {
       const results = fuse.search(query).map((result) => result.item)
-      setFilteredItems(results)
+
+      // only show first 25 results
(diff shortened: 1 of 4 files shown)


frontend/console/src/features/command-pallete/CommandPalette.tsx
```diff
@@ -25,20 +25,23 @@ export const CommandPalette: React.FC<CommandPaletteProps> = ({ isOpen, onClose
...
-      keys: ['title', { name: 'subtitle', weight: 1 }],
-      minMatchCharLength: 2,
-      fieldNormWeight: 2,
-    })
-  }, [items])
+  const fuseOptions = {
+    keys: ['title', { name: 'subtitle', weight: 1 }],
+    minMatchCharLength: 2,
+    fieldNormWeight: 2,
+  }
```
deniseli, line 32: Really nitty optimization but we could pull this out of the function/component now that it's just a constant to avoid it getting reinstantiated every call

frontend/console/src/features/command-pallete/CommandPalette.tsx
```diff
@@ -25,20 +25,23 @@ export const CommandPalette: React.FC<CommandPaletteProps> = ({ isOpen, onClose
...
+  const fuse = useMemo(() => new Fuse(items, fuseOptions, fuseIndex), [items, fuseIndex])
 
   useEffect(() => {
     if (query === '') {
       setFilteredItems([])
     } else {
       const results = fuse.search(query).map((result) => result.item)
-      setFilteredItems(results)
+
+      // only show first 25 results
```
deniseli, line 43: niiiiiiice!

frontend/console/src/features/modules/module.utils.ts
```diff
@@ -167,21 +168,30 @@ export const addModuleToLocalStorageIfMissing = (moduleName?: string) => {
...
+    topic: BubbleChatIcon,
+    typealias: AnonymousIcon,
+    secret: SquareLock02Icon,
+    subscription: MessageIncoming02Icon,
+    verb: FunctionIcon,
+  }
+
+  const normalizedDeclCase = declCase?.toLowerCase()
+  if (!normalizedDeclCase || !declIcons[normalizedDeclCase]) {
+    console.warn(`No icon for decl case: ${declCase}`)
```
deniseli, line 187: Ah this is way better! Thanks!

frontend/console/src/features/command-pallete/CommandPalette.tsx
wesbillman: Yeah we probably have a few places like this 😬

PR #3150, opened 2024-10-17
fix: reduce spammy logs for pings

matt2e:
We call Ping over grpc at a frequent rate when waiting for services to be ready, with each grpc failure being automatically logged (debug).
Also, soon language plugins will be using grpc and so the amount of ping logs at start up will roughly double.
This PR tries to reduce the noise.

### Changes:
- Ping logs:
    - Constant ping call failures are silenced (noop log sync passed into the call)
- Refactor: moved more logic into `rpc.Ping()` so that each caller doesn't need to write extra code, including
   - deadline handling
   - returning the relevant error (the latest ping error if deadline reached, otherwise the context error)

### DIFF
--- a/common/plugin/spawn.go
+++ b/common/plugin/spawn.go
@@ -12,12 +12,12 @@ import (
 	"time"
 
 	"connectrpc.com/connect"
+	"github.com/jpillora/backoff"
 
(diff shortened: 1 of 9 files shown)


matt2e:
People might prefer the noise... will wait for review on this one.

stuartwdouglas (approved):

PR #3161, opened 2024-10-18
fix: reporting call and ingress event errors

### LINKED ISSUE #3149: Console swallows invalid request errors
Sending a bad request in Console shows nothing on screen. e.g.
`echo.echo`:
```
{
  "name": 1
}
```

There's a 500 response to /xyz.block.ftl.v1.VerbService/Call that isn't being handled

wesbillman:
Fixes #3149

Also fixes:
- ingress request keys were not being added to the database
- ingress events not adding to the db because of request body reading twice resulting in `error:timeline: Failed to insert 1 events, most recent error: failed to read request body: http: invalid Read on closed Body`
- call event errors were not being inserted as events (just logged)
- call not reusing the ingress request key correctly
- TimelineIcon not respecting error fields in responses

Screenshot 2024-10-18 at 9 55 13 AM
Screenshot 2024-10-18 at 9 55 35 AM

### DIFF
--- a/backend/controller/controller.go
+++ b/backend/controller/controller.go
@@ -352,7 +352,17 @@ func New(ctx context.Context, conn *sql.DB, config Config, devel bool, runnerSca
 
 func (s *Service) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	start := time.Now()
-	requestKey := model.NewRequestKey(model.OriginIngress, fmt.Sprintf("%s %s", r.Method, r.URL.Path))
+	method := strings.ToLower(r.Method)
+	requestKey := model.NewRequestKey(model.OriginIngress, fmt.Sprintf("%s %s", method, r.URL.Path))
+
+	sourceAddress := r.RemoteAddr
+	err := s.dal.CreateRequest(r.Context(), requestKey, sourceAddress)
+	if err != nil {
+		observability.Ingress.Request(r.Context(), r.Method, r.URL.Path, optional.None[*schemapb.Ref](), start, optional.Some("failed to create request"))
+		http.Error(w, "Internal Server Error", http.StatusInternalServerError)
+		return
+	}
+
 	routes := s.schemaState.Load().httpRoutes[r.Method]
 	if len(routes) == 0 {
 		http.NotFound(w, r)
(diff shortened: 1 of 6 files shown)


backend/controller/controller.go
```diff
@@ -1074,6 +1072,34 @@ func (s *Service) callWithRequest(
...
+		ParentRequestKey: parentKey,
+		StartTime:        start,
+		DestVerb:         verbRef,
+		Callers:          callers,
+		Request:          req.Msg,
+	}
+
+	if currentCaller != nil && currentCaller.Module != module && !verb.IsExported() {
+		observability.Calls.Request(ctx, req.Msg.Verb, start, optional.Some("invalid request: verb not exported"))
+		err = connect.NewError(connect.CodePermissionDenied, fmt.Errorf("verb %q is not exported", verbRef))
```
deniseli, line 1087: Since this is the same as the error you return, should this just be:
```
err = connect.NewError(connect.CodePermissionDenied, fmt.Errorf("verb %q is not exported", verbRef))
callEvent.Response = either.RightOf[*ftlv1.CallResponse](err)
s.timeline.EnqueueEvent(ctx, callEvent)
return nil, err
```

frontend/console/src/features/timeline/timeline.utils.ts
```diff
@@ -22,4 +27,19 @@ const eventTextColorMap: Record<string, string> = {
...
+  return eventTextColorMap[event.entry.case || '']
+}
+
+const isError = (event: Event) => {
+  if (event.entry.case === 'call' && event.entry.value.error) {
+    return true
+  }
+  if (event.entry.case === 'ingress' && event.entry.value.error) {
+    return true
+  }
```
deniseli, line 43: `if (event.entry.value.error && ['call', 'ingress'].includes(event.entry.case))`? Or is there some nullability handling this doesn't work for hehe

deniseli (approved):
Niiiice

frontend/console/src/features/timeline/timeline.utils.ts
wesbillman: Yeah it's the typical "make sure this is a call, then you can read the value.error" stuff :)

backend/controller/controller.go
wesbillman: Nice, yeah good call. Updated.

PR #3214, opened 2024-10-28
chore: attempt to improve integration test output

stuartwdouglas:
There have been lots of failures recently where tests and timing out on the JVM build.

Because the go test runtime panics we don't get any output, and the tests on CI are configured to fully buffer output.

This change will start streaming output after a set time, so if there is a hang we should be able to get more information.

### DIFF
--- a/internal/exec/circularbuffer.go
+++ b/internal/exec/circularbuffer.go
@@ -14,10 +14,11 @@ import (
 )
 
 type CircularBuffer struct {
-	r    *ring.Ring
-	size int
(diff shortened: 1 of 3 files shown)


stuartwdouglas:
I thought about that, but if we end up with multiple timeouts then things can end up being confusing as to which timeout is actually taking effect. 

Also will context cancellation actually guarantee that we get the output? Doesn't it happen in an async manner, so if the initial timeout results in a fast exit the goroutine handing the cancellation may not have time to write the output?

PR #3230, opened 2024-10-29
fix: `ftl diff` no longer needs language plugins for schemas

matt2e:
This was slow and required a bind allocator.
Now that we have `$(projectroot)/.ftl/schemas`, `ftl diff` can use that to find module schemas.


frontend/cli/cmd_schema_diff.go
```diff
@@ -111,23 +101,7 @@ func localSchema(ctx context.Context, projectConfig projectconfig.Config, bindAl
 
 	for _, m := range modules {
 		go func() {
-			// Loading a plugin can be expensive. Is there a better way?
```
deniseli, line 114: very satisfying comment to remove :D

PR #3237, opened 2024-10-29
fix: improve delay before subscriber receives event

matt2e:
Before the change, the new integration test was showing delays ranging from 0.2s - 1.2s from when an event is published to when the async call is created.
After the change we are consistently under 0.2s.

### DIFF
--- a/backend/controller/pubsub/service.go
+++ b/backend/controller/pubsub/service.go
@@ -82,7 +82,9 @@ func (s *Service) poll(ctx context.Context) {
 			publishedAt = optional.None[time.Time]()
 
 		case <-time.After(poll):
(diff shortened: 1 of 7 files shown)


backend/controller/pubsub/service.go
```diff
@@ -22,7 +23,7 @@ const (
 	// Events can be added simultaneously, which can cause events with out of order create_at values
 	// By adding a delay, we ensure that by the time we read the events, no new events will be added
 	// with earlier created_at values.
-	eventConsumptionDelay = 200 * time.Millisecond
+	eventConsumptionDelay = 100 * time.Millisecond
```
matt2e, line 26: Reduced this delay as well.

PR #3251, opened 2024-10-30
chore: don't build/start console for tests unless explicitly required

stuartwdouglas:


Justfile
```diff
@@ -48,7 +50,15 @@ build-generate:
 
 # Build command-line tools
 build +tools: build-protos build-zips build-frontend
+  @just build-without-frontend $@
```
safeer, line 53: Do we need to include the `pnpm-install` dependency here?

stuartwdouglas: I don't think so, because build-protos is often a no-op, and if it actually does something it calls 'build-protos-unconditionally', which already has the pnpn-install dependency.

This optimization saves a bit of time, as there is no pnpn involved in most of the tests.

safeer (approved):

Justfile
```diff
@@ -48,7 +50,15 @@ build-generate:
...
+
+# Build command-line tools
+# This does not have a dependency on the frontent
```
wesbillman, line 56: `frontent` -> `frontend`

@mistermoe `frontent` lives on!

wesbillman (approved):

PR #3257, opened 2024-10-30
chore: split up proto files

stuartwdouglas:
This allows the JVM subsystem to only generate Java classes for services it cares about, speeding up build times.


deniseli (approved):
lgtm for dx

alecthomas (approved):

PR #3266, opened 2024-10-31
fix: clear events when streaming starts

wesbillman:
See #3260

We can get some console connection errors in our deployed environment which causes streaming to get restarted. However, it seems like we don't clear the cache when this happens so duplicate events get returned from our streaming hooks. This change should ensure the events are cleared anytime we restart the stream.


deniseli (approved):
nice!

PR #3276, opened 2024-10-31
fix: smoke test the tagged release before upgrading

safeer:
Run the tagged release's smoke test on its released image, which sets up the deployments to upgrade, before running the upgrade tests.

### DIFF
--- a/.github/workflows/smoketest.yml
+++ b/.github/workflows/smoketest.yml
@@ -29,31 +29,32 @@ jobs:
           ref: v${{ env.LATEST_VERSION }}
           fetch-depth: 1
           fetch-tags: true
-      # The following grabs the current smoke test and deployment Justfile from the HEAD. This
-      # is to solve the chicken-and-egg. Once it's in the tagged release, skip this.
-      - name: Check out the current smoke test
-        uses: actions/checkout@v4
-        with:
-          clean: false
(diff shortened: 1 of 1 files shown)


.github/workflows/smoketest.yml
```diff
@@ -29,31 +29,32 @@ jobs:
...
-      - name: Replace the tagged release smoketest with the current smoketest
-        run: |
-          set -euo pipefail
-          echo "Replacing tagged release smoketest with current smoketest"
-          rm -r ./smoketest && mv ./tmpsmoketest/smoketest ./smoketest
-          echo "Replacing deployment Justfile with current Justfile"
-          cp ./tmpsmoketest/deployment/Justfile ./deployment/Justfile
-          echo "Removing tmpsmoketest"
-          rm -r ./tmpsmoketest
+      # # # These are disabled until the tagged release has the integration test changes.
```
jonathanj-square, line 32: is this commented out because you plan on bringing it back?

jonathanj-square (approved):

.github/workflows/smoketest.yml
safeer: Yeah it was, until we had a tagged release with the smoketest changes. Going in a different direction now though 🙃

stuartwdouglas (approved):

PR #3299, opened 2024-11-01
feat: JVM can make local verb calls via generated clients

tomdaffurn:


internal/buildengine/engine.go
```diff
@@ -905,6 +905,10 @@ func (e *Engine) buildWithCallback(ctx context.Context, callback buildCallback,
 				return err
 			}
 		}
+		err = WriteGenericSchemaFiles(newSchemas, metasMap)
```
stuartwdouglas, line 908: So this approach will write out the generated schema after the build, which then results in clients being generated?

Won't this also generate new copies of things like Data classes that will then be duplicated?

I guess with this approach you have to check the schema in, as projects won't be buildable otherwise.

tomdaffurn: This doesn't change the approach much. Previously it wrote schema files inside the generated stubs function on line 903. Stubs are only written for modules _not_ in the last topology group. ie modules that have dependant modules.

Now it does the same but outside the if statement, for every group in the topology.

PR #3301, opened 2024-11-01
feat: codegen python classes from protos for schema extraction

mistermoe:
# Summary
This PR configures `buf` to generate python classes from protos to be used for schema generation 

> [!NOTE]
> currently the codgen'd proto classes are configured to output into the `ftl` python sdk. it's likely that this will change if we separate python schema generation code from the python sdk itself.  haven't decided yet, so defaulting to a single python package atm.

> [!WARNING]
> running `buf generate` changed the `connect-es` version in some ts proto classes from `v1.6.1` -> `v1.5.0`. not 100% sure why this happened. may be because protos haven't been generated in awhile? or maybe the `v1.6.1` came from when we were using remote plugins. @wesbillman @deniseli y'all might have more insight on this!


python-runtime/ftl/src/ftl/protos/xyz/block/ftl/v1/language/language_pb2.pyi
```diff
@@ -0,0 +1,273 @@
+from google.protobuf import struct_pb2 as _struct_pb2
```
alecthomas, line 1: Mind adding these to the linguist generated file types?

alecthomas (approved):

python-runtime/ftl/src/ftl/protos/xyz/block/ftl/v1/language/language_pb2.pyi
mistermoe: yee @alecthomas. done https://github.com/TBD54566975/ftl/pull/3301/commits/7e77dfc9979beedd5e12fbb4fc404924a8889e33. that look right to you?

mistermoe: @alecthomas  so, i was curious about whether the line i added to `.gitattributes` actually worked. ran `github-linguist` in a docker container. turns out linguist automatically ignores generated protos!

```
find python-runtime/ftl/src/ftl/protos -type f -name "*.py" | head -n 1 | xargs github-linguist
python-runtime/ftl/src/ftl/protos/xyz/block/ftl/v1/ftl_pb2.py: 43 lines (37 sloc)
  type:      Text
  mime type: application/x-python
  language:  Python
  appears to be generated source code
```

there's several rules baked right into linguist which you can see [here](https://github.com/github-linguist/linguist/blob/main/lib/linguist/generated.rb#L91-L93)

image

alecthomas: Yeah I was specifically referring to the pyi files

mistermoe: aye, automatically identified as well:

```
python-runtime/ftl/src/ftl/protos/xyz/block/ftl/v1/controller_pb2.pyi: 348 lines (313 sloc)
  type:      Text
  mime type: text/plain
  language:  Python
  appears to be generated source code
```

PR #3311, opened 2024-11-04
chore: add istio access logging to tests

stuartwdouglas:


internal/rpc/rpc.go
```diff
@@ -21,6 +21,40 @@ import (
 	"github.com/TBD54566975/ftl/internal/log"
 )
 
+// rbacWotkaroundTransport is a RoundTripper that retries requests when they fail due to RBAC errors.
```
matt2e, line 24: spelling mistake `Wotk` => `Work` in comment and func name

PR #3325, opened 2024-11-04
fix: improve startup time

stuartwdouglas:
1s is a very long poll time, reducing this will improve startup time


frontend/cli/cmd_serve.go
```diff
@@ -390,7 +390,7 @@ func waitForControllerOnline(ctx context.Context, startupTimeout time.Duration,
 	ctx, cancel := context.WithTimeout(ctx, startupTimeout)
 	defer cancel()
 
-	ticker := time.NewTicker(time.Second)
+	ticker := time.NewTicker(time.Millisecond + 50)
```
wesbillman, line 393: Do you want `50ms` here? Should it be `*` instead of `+`?

stuartwdouglas: oops

PR #3382, opened 2024-11-13
feat: add online-boutique example

stuartwdouglas:
Moves the online-boutique example out of the examples repo and get it compiling. I have not looked at the web or mobile directories, they have just been copied as-is.


alecthomas (approved):
Awesome.

It'd be nice to move this to real databases/KV stores once provisioning + migrations are done.

PR #3392, opened 2024-11-15
feat: hot reload

stuartwdouglas:
JVM languages can now used how reload for instant feedback


jvm-runtime/plugin/common/jvmcommon.go
```diff
@@ -167,6 +172,9 @@ func (s *Service) SyncStubReferences(ctx context.Context, req *connect.Request[l
 // rebuild must include the latest build context id provided by the request or subsequent BuildContextUpdated
 // calls.
 func (s *Service) Build(ctx context.Context, req *connect.Request[langpb.BuildRequest], stream *connect.ServerStream[langpb.BuildEvent]) error {
+	if req.Msg.RebuildAutomatically {
+		return s.handleDevModeRequest(ctx, req, stream)
+	}
 	events := make(chan updateEvent, 32)
```
matt2e, line 178: the rest of this function can be gutted I think as it has a lot of logic for receiving new build contexts and watching for file changes, which only happens if `req.Msg.RebuildAutomatically` is true.

Oh wait I'm actually not sure how you are reacting to build context changes?

stuartwdouglas: I am not handling that properly I think, I need to move some of this logic into the hot reload path.

stuartwdouglas: I think I am handling this properly now

stuartwdouglas:
I need to add some tests around this, but I think they need a different approach to the existing language tests, as most of the logic is handled by Quarkus itself.

PR #3394, opened 2024-11-15
feat: inject topics/subscriptions in go

worstell:
go:
`type MyTopic = ftl.TopicHandle[MyEvent]`
schema:
`topic myTopic foo.MyEvent`

go:
`type MySubscription = ftl.SubscriptionHandle[MyTopic, MySinkClient, MyEvent]`
schema:
`subscription mySubscription foo.MyEvent`


go-runtime/schema/schema_integration_test.go
```diff
@@ -565,7 +567,8 @@ func testErrorReporting(t *testing.T) {
 		`146:6-45: enum discriminator "TypeEnum3" cannot contain exported methods`,
 		`149:6-35: enum discriminator "NoMethodsTypeEnum" must define at least one method`,
 		`161:3-14: unexpected token "d"`,
-		`168:2-62: can not publish directly to topics in other modules`,
+		// todo: do we want this? why do we support exporting topics if so?
+		//`168:2-62: can not publish directly to topics in other modules`,
```
worstell, line 571: should this check be removed?

go-runtime/ftl/ftltest/ftltest.go
```diff
@@ -438,67 +438,6 @@ func getDSNFromSecret(ftl internal.FTL, module, name string) (string, error) {
 	return dsn, nil
 }
 
-// WithSubscriber adds a subscriber during a test
```
worstell, line 441: will be reimplementing with a different strategy separately

go-runtime/schema/schema_integration_test.go
matt2e: Ah we still want this limitation. We export topics so that subscriptions in other modules can subscribe to them. But we don't want to let anyone publish to topics directly.

worstell: ah ok cool, i'll restrict that via the generated external module so there's no `Publish` method at all

worstell: actually that won't work. will keep the static check

PR #3429, opened 2024-11-19
feat: deterministic ordering for verb metadata

worstell:
inject in the correct position, but display metadata in a deterministic order in the schema. if param orderings change, schema will remain consistent


stuartwdouglas:
I wonder if this should actually go into some kinda of schema normalization handler on the server? This would mean that all runtimes get the same normalization rules and it does not have to be implemented for each language.

worstell:
> I wonder if this should actually go into some kinda of schema normalization handler on the server? This would mean that all runtimes get the same normalization rules and it does not have to be implemented for each language.

this will actually apply across runtimes already, `schema.ValidateSchema` gets called when we deserialize schema from proto

stuartwdouglas (approved):

PR #3434, opened 2024-11-19
feat: always --recreate in dev mode

stuartwdouglas:


worstell (approved):

matt2e (commented):
Does `ftl dev --help` still declare `--recreate` as an option? You'd probably need to shift around the cmd structs a little to remove it

alecthomas (approved):

PR #3437, opened 2024-11-20
chore: spelling fix

matt2e:


jvmakine (approved):
This is why I should write all comments with AI!

PR #3458, opened 2024-11-22
feat: Resolve DSN from runtime in PG Proxy

jvmakine:
High level changes:
- Stop sending PostgreSQL DSNs with the `ModuleContext` to the runners.
- Instead, when `GetDatabase` in the runner is called, and there is no DSN for the DB, it gets translated to pgproxy DSN. 
- We start the pgproxy before the deployment in the runner, and pass the local address in an environment variable to the runner.
- pgproxy gets the active schema, and redirects connections to the DSN defined in the runtime of that schema.

Follow-ups
- Integrate MySQL proxy
- Remove DSNs from `ModuleContext` completely

### DIFF
--- a/backend/runner/runner.go
+++ b/backend/runner/runner.go
@@ -24,6 +24,7 @@ import (
 	"github.com/alecthomas/types/optional"
 	"github.com/jpillora/backoff"
 	"github.com/otiai10/copy"
+	"golang.org/x/sync/errgroup"
 	"google.golang.org/protobuf/types/known/structpb"
 	"google.golang.org/protobuf/types/known/timestamppb"
 
@@ -39,6 +40,7 @@ import (
 	"github.com/TBD54566975/ftl/internal/log"
 	"github.com/TBD54566975/ftl/internal/model"
 	ftlobservability "github.com/TBD54566975/ftl/internal/observability"
+	"github.com/TBD54566975/ftl/internal/pgproxy"
 	"github.com/TBD54566975/ftl/internal/rpc"
 	"github.com/TBD54566975/ftl/internal/schema"
 	"github.com/TBD54566975/ftl/internal/slices"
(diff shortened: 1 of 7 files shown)


backend/runner/runner.go
```diff
@@ -568,3 +595,27 @@ func (s *Service) healthCheck(writer http.ResponseWriter, request *http.Request)
...
+			databases[db.Name] = db
+		}
+	}
+
+	if err := pgproxy.New(":0", func(ctx context.Context, params map[string]string) (string, error) {
+		db, ok := databases[params["database"]]
+		if !ok {
+			return "", fmt.Errorf("database %s not found", params["database"])
+		}
+		logger.Debugf("Resolved DSN (%s): %s", params["database"], db.Runtime.DSN)
```
jvmakine, line 614: Here is where the logic to use a dsn resolver based on provisioned runtime (like `type: aws-iam-auth`) will be added

backend/runner/runner.go
```diff
@@ -568,3 +603,27 @@ func (s *Service) healthCheck(writer http.ResponseWriter, request *http.Request)
 	}
 	writer.WriteHeader(http.StatusServiceUnavailable)
 }
+
+func (s *Service) startPgProxy(ctx context.Context, module *schema.Module, started chan<- pgproxy.Started) error {
+	logger := log.FromContext(ctx)
+
+	databases := map[string]*schema.Database{}
+	for _, decl := range module.Decls {
```
stuartwdouglas, line 611: Shouldn't this check the database type? Also if there is no databases should it just return?

jvmakine: Yes, it should filter by type! Will fix.

It will need to fail if there is no DB, though. That is the error returned to the runner when connecting to the DB

jvmakine: Ah, you mean should we not start the proxy at all if there are no PG databases? Yes, good idea. Will change that.

jvmakine: Done!

stuartwdouglas (approved):

PR #3499, opened 2024-11-25
chore: convert database runtime to a oneof to support more connection methods in the future

jvmakine:
The idea is to have the different ways to do DB connections in the DatabaseRuntime, and then to use that in DB proxies to establish the connection.


backend/protos/xyz/block/ftl/v1/schema/schema.proto
```diff
@@ -36,6 +36,11 @@ message Config {
   Type type = 4;
 }
 
+message DSNDatabaseRuntime {
```
jvmakine, line 39: This generated code is the crux of this change

stuartwdouglas (approved):

PR #3528, opened 2024-11-26
feat: replace subscriptions in schema with verb subscribe metadata

matt2e:
Subscriptions are now defined on a verb. Schema no longer lists subscriptions.

go:
```go
//ftl:subscribe <topic> from=(beginning|latest) (deadletter)?
```

java and kotlin:
```java
// Separating options from @Subscription allows us to use generated subscription types for external topics
@Subscription(topic="verbName")
@SubscriptionOptions(from = FromOffset.LATEST, deadLetter = false) // deadLetter can be left out
```

Other changes:
- FromOffset and DeadLetter options have no effect yet
- Only one subscription per verb
- Updated docs


frontend/console/src/features/graph/graph-utils.ts
```diff
@@ -73,17 +73,17 @@ const createModuleChildren = (module: Module, nodePositions: Record<string, { x:
...
-        module.name,
-        nodeId(module.name, subscription.subscription?.name),
-        subscription.subscription?.name || '',
-        'subscription',
-        nodePositions,
-        subscription,
-        isDarkMode,
-      ),
-    ),
+    // ...(module.subscriptions || []).map((subscription: Subscription) =>
```
matt2e, line 76: @wesbillman This PR removes Subscriptions as a standalone concept in the schema. And I'm not sure how we want that to affect the console... I've just commented out the broken parts for now

stuartwdouglas (approved):

PR #3529, opened 2024-11-26
feat: move runner scaling to the provisioner

### LINKED ISSUE #3483: Refactor k8s provisioning to be a provisioner

stuartwdouglas:
fixes #3483


backend/provisioner/service.go
```diff
@@ -149,6 +150,8 @@ func replaceOutputs(to []*provproto.Resource, from []*provproto.Resource) error
 				}
 				r.Subscription.Output = subscriptionFrom.Subscription.Output
 			}
+		case *provproto.Resource_Runner:
+			// Ignore
```
jvmakine, line 154: Why ignoring here? Should we not just set the outpus?

backend/provisioner/registry.go
```diff
@@ -273,9 +276,18 @@ func ExtractResources(msg *ftlv1.CreateDeploymentRequest) (*ResourceGraph, error
 			to:   dep.ResourceId,
 		})
 	}
-
+	runnerResource := &provisioner.Resource{
+		ResourceId: module.GetName() + "-runner",
```
jvmakine, line 280: Should we use deployment id here instead? I think that would fit better to the given model, given that each deployment has a separate runner as a resource.

jvmakine (approved):

backend/provisioner/registry.go
stuartwdouglas: Good point

stuartwdouglas: Actually we don't really seem to have the deployment name here?

jvmakine: Hmm, yeah. I think it currently created after the provisioning, which is annoying. I suppose that is fine for now, though it breaks the model a bit.

PR #3538, opened 2024-11-26
chore: remove default from provisioner config

jvmakine:
The default is confusing regarding the ordering and grouping of provisioner tasks. Let's get rid of it!


stuartwdouglas (changes requested):
You missed the config map in the helm chart

jvmakine:
@stuartwdouglas Updated the config map as well!

stuartwdouglas (approved):

PR #3546, opened 2024-11-27
feat: add ⌥+⏎ as a shortcut for the verb form

bradleydwyer:


frontend/console/src/features/modules/decls/verb/VerbFormInput.tsx
```diff
@@ -18,13 +19,29 @@ export const VerbFormInput = ({
...
+        event.preventDefault()
+        formRef.current?.dispatchEvent(new Event('submit', { cancelable: true, bubbles: true }))
+      }
+    }
+
+    document.addEventListener('keydown', handleGlobalKeyDown)
+    return () => {
+      document.removeEventListener('keydown', handleGlobalKeyDown)
+    }
+  }, [path, readOnly, onSubmit])
```
wesbillman, line 41: Not super critical here, but we have a similar example of this in the searchInput for command pallete: https://github.com/TBD54566975/ftl/blob/main/frontend/console/src/layout/navigation/SearchInput.tsx#L10-L25

Most of this is basically the same, but you could add support for other `meta` keys as well like this:

```ts
  const shortcutText = window.navigator.userAgent.includes('Mac') ? '⌘ + K' : 'Ctrl + K'
```

frontend/console/e2e/cron.spec.ts
```diff
@@ -24,3 +24,33 @@ test('send cron request', async ({ page }) => {
...
+  const responseText = await responseEditor.textContent()
+  const responseJson = JSON.parse(responseText?.trim() || '{}')
+
+  expect(responseJson).toEqual({})
+})
+
+test('submit cron form using ⌥+⏎ shortcut without focusing first', async ({ page }) => {
+  await navigateToDecl(page, 'cron', 'thirtySeconds')
+
+  await page.keyboard.press('Alt+Enter')
```
wesbillman, line 47: I'm surprised we have to use the `alt` key here, but maybe that's just how this works :)

wesbillman (approved):
Woot!!

LGTM, but you might want to align this logic with the command palette so we have a consistent approach to handling these shortcuts.

frontend/console/e2e/cron.spec.ts
bradleydwyer: I think this is because of the difference between using document and window. Switching to window will allow the command key. Will update.

frontend/console/src/features/modules/decls/verb/VerbFormInput.tsx
bradleydwyer: Yeah. considered the component and just avoided until a second case case popped up - but I had forgotten about the search input. What I'll do is align to the search, and come back and figure out a more generic component later.

wesbillman: for sure! That's mostly what I meant too. No need to make something generic for just these 2 cases, but the command palette might be a good place to copy pasta from.

frontend/console/e2e/cron.spec.ts
```diff
@@ -24,3 +24,46 @@ test('send cron request', async ({ page }) => {
...
+
+      const responseText = await responseEditor.textContent()
+      const responseJson = JSON.parse(responseText?.trim() || '{}')
+
+      expect(responseJson).toEqual({})
+      break
+    } catch (error) {
+      if (attempt === 2) throw error
+    }
+  }
```
wesbillman, line 48: @bradleydwyer I added this helper to fix the `CommandPalette` tests, which might help here too: https://github.com/TBD54566975/ftl/blob/main/frontend/console/e2e/helpers.ts#L27-L35.

You can see how it's used here: https://github.com/TBD54566975/ftl/blob/main/frontend/console/e2e/command-palette.spec.ts#L16-L28

PR #3549, opened 2024-11-27
feat: change JVM API for pubsub

stuartwdouglas:
The main changes are:

- No generated annotations, instead `ConsumeableTopic` classes will be generated
- Subscriptions for local and remote topics look the same
- This results in a more consistent and user friendly API


matt2e (approved):
Sweet

PR #3557, opened 2024-11-28
fix: support annotations in registry

jvmakine:


charts/ftl/templates/registry.yaml
```diff
@@ -25,6 +25,10 @@ apiVersion: v1
 kind: Service
 metadata:
   name: {{ include "ftl.fullname" . }}-registry
+  {{- if .Values.registry.service.annotations }}
```
stuartwdouglas, line 28: This fails if the annotations are not defined, as registry.service is null in the values file

PR #3598, opened 2024-12-03
feat: jvm publishes to kafka and adds partition mapper

matt2e:
Removes jvm from pubsub integration tests as it is no longer consuming from where it is publishing.

### DIFF
--- a/backend/controller/pubsub/integration_test.go
+++ b/backend/controller/pubsub/integration_test.go
@@ -19,7 +19,7 @@ func TestPubSub(t *testing.T) {
 	calls := 20
 	events := calls * 10
 	in.Run(t,
-		in.WithLanguages("java", "go", "kotlin"),
+		in.WithLanguages("go"),
 		in.CopyModule("publisher"),
 		in.CopyModule("subscriber"),
 		in.Deploy("publisher"),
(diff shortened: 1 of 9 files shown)


examples/java/ExampleInterface.java
```diff
@@ -0,0 +1,8 @@
+
+public interface ExampleInterface {
+    void instanceMethod();
```
stuartwdouglas, line 3: What is this?

backend/controller/pubsub/integration_test.go
```diff
@@ -19,7 +19,7 @@ func TestPubSub(t *testing.T) {
 	calls := 20
 	events := calls * 10
 	in.Run(t,
-		in.WithLanguages("java", "go", "kotlin"),
```
stuartwdouglas, line 22: How far away is consumer support? Should we hold this until it is ready?

stuartwdouglas (approved):

examples/java/ExampleInterface.java
matt2e: 🤦‍♂️

backend/controller/pubsub/integration_test.go
stuartwdouglas: What do you mean by:

> runner sends to kafka/redpanda, but also to the runner.

PR #3609, opened 2024-12-03
fix: publish to old pubsub impl and kafka

matt2e:
This leaves pubsub in a stable state while we put kafka work on pause


matt2e:
Blocked on this: https://github.com/TBD54566975/ftl/issues/3617

PR #3618, opened 2024-12-04
feat: timeline service keeps events in memory

### LINKED ISSUE #3626: In-memory storage

matt2e:
closes #3626
Very simple implementation to be begin with.
Allows:
- creating events
- getting events with super simple filters
- clearing out old events


backend/protos/xyz/block/ftl/timeline/v1/event.proto
```diff
@@ -139,15 +139,19 @@ message Event {
   google.protobuf.Timestamp time_stamp = 1;
   // Unique ID for event.
   int64 id = 2;
+  EventEntry entry = 3;
+}
+
+message EventEntry {
```
matt2e, line 145: EventEntry was created was created just so we can reuse the sumtype within `Event` and `CreateEvent`.
The extra level of wrapping is ugly everywhere else though...

backend/controller/console/console.go
```diff
@@ -640,21 +640,23 @@ func eventDALToProto(event timeline.Event) *pbtimeline.Event {
...
-				Call: &pbtimeline.CallEvent{
-					RequestKey:    requestKey,
-					DeploymentKey: event.DeploymentKey.String(),
-					TimeStamp:     timestamppb.New(event.Time),
-					SourceVerbRef: sourceVerbRef,
-					DestinationVerbRef: &schemapb.Ref{
-						Module: event.DestVerb.Module,
-						Name:   event.DestVerb.Name,
+			Entry: &pbtimeline.EventEntry{
+				Entry: &pbtimeline.EventEntry_Call{
```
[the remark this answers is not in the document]
matt2e: [See comment above](https://github.com/TBD54566975/ftl/pull/3618#discussion_r1868769503)

backend/protos/xyz/block/ftl/timeline/v1/event.proto
matt2e: I suppose I can just repeat the `oneof` in both places and it'd make it a lot nicer to deal with events everywhere (the 99% case)

matt2e: (removed now)

PR #3641, opened 2024-12-05
chore: remove DB based leases

stuartwdouglas:
This means async calls no longer have linked leases, but as they are going away anyway that should not be a problem.


stuartwdouglas:
This seems to work now the hermit stuff is fixed, the helm charts need to be updated though, and possibly some kube tests.

charts/ftl/templates/lease-services.yaml
```diff
@@ -0,0 +1,21 @@
...
+  labels:
+    {{- include "ftl.labels" . | nindent 4 }}
+  name: {{ include "ftl.fullname" . }}-lease
+  {{- if .Values.lease.service.annotations }}
+  annotations:
+    {{- toYaml .Values.service.annotations | nindent 4 }}
+  {{- end }}
+spec:
+  ports:
+    {{- range .Values.lease.ports }}
```
worstell, line 13: should this be `{{- range .Values.lease.service.ports }}`? and should it be `{{- toYaml .Values.lease.service.annotations | nindent 4 }}` above?

charts/ftl/templates/lease-services.yaml
```diff
@@ -0,0 +1,21 @@
...
+    - name: {{ .name }}
+      port: {{ .port }}
+      protocol: {{ .protocol | default "TCP" }}
+      targetPort: {{ .targetPort }}
+    {{- end }}
+  selector:
+    {{- include "ftl-lease.selectorLabels" . | nindent 4 }}
+  type: {{ .Values.controller.leaseService.type | default "ClusterIP" }}
```
worstell, line 21:
```suggestion
  type: {{ .Values.lease.service.type | default "ClusterIP" }}
```

worstell (approved):

charts/ftl/templates/lease-services.yaml
stuartwdouglas: yes, it should

PR #3668, opened 2024-12-08
chore: remove types.ftl.go

stuartwdouglas:
I don't really think these should be checked in, a lot of the ones in the repo are out of date anyway.


stuartwdouglas:
This seems to have unexpected consequences for go mod tidy, as it seems like the types.ftl.go files in some cases are the only actual FTL dependencies in the source.

PR #3689, opened 2024-12-09
chore: move pubsub out of controller

stuartwdouglas:
Move some of the legacy pubsub stuff out of the controller. This code won't be around for that much longer, so this will let us remove DB dependencies from the controller.


backend/runner/proxy/proxy.go
```diff
@@ -96,8 +100,10 @@ func (r *Service) AcquireLease(ctx context.Context, c *connect.BidiStream[ftllea
 
 }
 
-func (r *Service) PublishEvent(ctx context.Context, c *connect.Request[ftldeployment.PublishEventRequest]) (*connect.Response[ftldeployment.PublishEventResponse], error) {
-	event, err := r.controllerDeploymentService.PublishEvent(ctx, connect.NewRequest(c.Msg))
+func (r *Service) PublishEvent(ctx context.Context, c *connect.Request[ftlpubsubv1.PublishEventRequest]) (*connect.Response[ftlpubsubv1.PublishEventResponse], error) {
+	logger := log.FromContext(ctx)
+	logger.Warnf("PUBLISH MESSAGE -------")
```
matt2e, line 105: Looks like this is just some debug logs?

stuartwdouglas: oops, fixed

matt2e (approved):

PR #3693, opened 2024-12-10
fix: wait for module deployment

stuartwdouglas:
Wait for module deployment before building next group


wesbillman (approved):
Nice!!!

matt2e (approved):

matt2e:
Ahh so this may have been a older issue, but we sped up build times and slowed down deploy times...

PR #3711, opened 2024-12-10
chore: pull out admin service

worstell:


backend/admin/service.go
```diff
@@ -2,22 +2,52 @@ package admin
...
+	DSN                  string   `help:"DAL DSN." default:"${dsn}" env:"FTL_DSN"`
+	MaxOpenDBConnections int      `help:"Maximum number of database connections." default:"20" env:"FTL_MAX_OPEN_DB_CONNECTIONS"`
+	MaxIdleDBConnections int      `help:"Maximum number of idle database connections." default:"20" env:"FTL_MAX_IDLE_DB_CONNECTIONS"`
+}
+
+func (c *Config) SetDefaults() {
+	if err := kong.ApplyDefaults(c, kong.Vars{"dsn": dsn.PostgresDSN("ftl")}); err != nil {
+		panic(err)
+	}
+}
```
wesbillman, line 39: Do we want/need `kong` stuff in here? I wonder if we could do this `kong` work in the `cli/cmd` code. Then, we might be able to remove this from the `service` layer.

backend/timeline/service.go
```diff
@@ -40,6 +40,8 @@ type service struct {
 	events []*timelinepb.Event
 }
 
+var _ timelineconnect.TimelineServiceHandler = (*service)(nil)
+
```
wesbillman, line 44: nice!

wesbillman (approved):

backend/admin/service.go
worstell: i don't feel super strongly - i think we went with calling it in `start` to centralize the logic across invocations from CLI, wdyt? since the config fields are defined with `default` field tags it doesn't feel completely out of place to me in the service layer

PR #3717, opened 2024-12-11
feat: consume pubsub

matt2e:
- Runners now consume from redpanda
- Controller no longer sets up a pub sub service
- DB deletion and old pubsub deletion will be in a follow up PR


internal/integration/harness.go
```diff
@@ -395,6 +408,35 @@ func run(t *testing.T, actionsOrOptions ...ActionOrOption) {
...
+				assert.NoError(t, err)
+				defer client.Close()
+
+				clusterAdmin, err := sarama.NewClusterAdminFromClient(client)
+				assert.NoError(t, err)
+				defer clusterAdmin.Close()
+
+				// Try to reset up to 3 times
+				// There can be a slight delay for consumer groups to be empty after a previous run of FTL ends
+				for i := range 15 {
```
matt2e, line 427: This PR does not close kafka consumers cleanly so it can take about 10s for pubsub for consumer groups to be empty. Resetting pubsub has to wait for this to happen first.

stuartwdouglas: Maybe removed the `volumes` section from docker compose so the data does not persist over restarts?

stuartwdouglas (approved):
Awesome, lets get this in.

PR #3722, opened 2024-12-11
fix: inject timeline client instead of using ctx

### LINKED ISSUE #3704: Pass `timeline` client into cron at start rather than grabbing it from context
Then we can get rid of grabbing it from context here: https://github.com/TBD54566975/ftl/blob/main/backend/cron/service.go#L135

There might be other services that need a similar change.

### LINKED ISSUE #3708: Timeline client should be passed around rather then attached to context

wesbillman:
Fixes #3708
Fixes #3704

### DIFF
--- a/backend/timeline/client.go
+++ b/backend/timeline/client.go
@@ -20,8 +20,6 @@ const (
 	maxBatchDelay = 100 * time.Millisecond
 )
 
-type timelineContextKey struct{}
-
 type Client struct {
 	timelinev1connect.TimelineServiceClient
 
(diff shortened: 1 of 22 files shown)


backend/timeline/client.go
```diff
@@ -40,18 +40,6 @@ func NewClient(ctx context.Context, endpoint *url.URL) *Client {
 	return client
 }
 
-func ContextWithClient(ctx context.Context, client *Client) context.Context {
```
worstell, line 43: can delete line 23 as well (`type timelineContextKey struct{}`)

worstell (approved):

PR #3736, opened 2024-12-12
feat: allow easy generation of a trace file

stuartwdouglas:
If you start FTL with --trace=trace.out after execution you can look at the execution time details via:

go tool trace trace.out


wesbillman (approved):
Sweet!

PR #3791, opened 2024-12-17
fix: FTL init dir

stuartwdouglas:


wesbillman (approved):
Nice! Yeah this seems more like what a user would want to me as well.

PR #3792, opened 2024-12-17
chore: fix integration test matrix

matt2e:


matt2e:
Looking at the results:
- Some integration tests don't seem to run still? "no tests to run"
- Some integration tests fail because ftl is still running from the previous test?

PR #3795, opened 2024-12-17
chore: add a raft library

jvmakine:
This is a typed version of dragonboat APIs. It also does some simplifications, namely
- No immutable files in snapshots
- No support for cancelling snapshot save / restore
- No active sessions, which can result into duplicate events on timeouts / errors.

Also, it does not support cleanly removing members from the cluster yet.

However, it supports
- typed events
- typed queries
- snapshotting
- joining to an existing cluster as a new member.

Also, I did not enable gossip protocol that can be used to retain node identity over changing addresses, as stateful sets should have a persistent address in K8s


internal/raft/cluster.go
```diff
@@ -0,0 +1,201 @@
...
+// R is the query response type.
+type ShardHandle[E Event, Q any, R any] struct {
+	shardID uint64
+	cluster *Cluster
+	session *client.Session
+}
+
+// Propose an event to the shard.
+func (s *ShardHandle[E, Q, R]) Propose(ctx context.Context, msg E) error {
+	if s.cluster.nh == nil {
```
alecthomas, line 50: This probably needs a lock too?

internal/raft/statemachine.go
```diff
@@ -0,0 +1,108 @@
...
+	Update(msg E) error
+	// Save the state of the state machine to a snapshot.
+	Save(writer io.Writer) error
+	// Recover the state of the state machine from a snapshot.
+	Recover(reader io.Reader) error
+	// Close the state machine.
+	Close() error
+}
+
+type stateMachineShim[Q any, R any, E Event, EPtr Unmasrshallable[E]] struct {
```
alecthomas, line 41: Doc for this would be good.

internal/raft/statemachine.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/lni/dragonboat/v4/statemachine"
+)
+
+// Event to update the state machine. These are stored in the Raft log.
+type Event interface {
+	encoding.BinaryMarshaler
+}
+
+// Unmasrshallable is a type that can be unmarshalled from a binary representation.
+type Unmasrshallable[T any] interface {
```
stuartwdouglas, line 17: There is a typo in masrsable.

internal/raft/cluster.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+// AddShard adds a shard to the cluster.
+func AddShard[Q any, R any, E Event, EPtr Unmasrshallable[E]](
+	ctx context.Context,
+	to *Cluster,
+	shardID uint64,
+	sm StateMachine[Q, R, E, EPtr],
+) *ShardHandle[E, Q, R] {
+	to.shards[shardID] = newStateMachineShim[Q, R, E, EPtr](sm)
```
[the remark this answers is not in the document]
jvmakine: I was thinking this is only called at creation before the cluster member is started. Will clarify

internal/raft/cluster.go
jvmakine: Does it? It should be only set once at `Start`

internal/raft/cluster.go
```diff
@@ -0,0 +1,201 @@
...
+
+type RaftConfig struct {
+	InitialMembers []string `help:"Initial members" required:""`
+	ReplicaID      uint64   `help:"Node ID" required:""`
+	DataDir        string   `help:"Data directory" required:""`
+	RaftAddress    string   `help:"Address to advertise to other nodes" required:""`
+	ListenAddress  string   `help:"Address to listen for incoming traffic. If empty, RaftAddress will be used."`
+	// Raft configuration
+	ElectionRTT        uint64 `help:"Election RTT" default:"10"`
+	HeartbeatRTT       uint64 `help:"Heartbeat RTT" default:"1"`
```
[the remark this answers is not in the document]
jvmakine: I will clarify. These are multiples of the configured average roundtrip between members, not time units.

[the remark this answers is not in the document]
jvmakine: Which I now realised I hard coded to 200ms and forgot to add to this config 🤦

internal/raft/statemachine.go
jvmakine: I could swear this was written by an AI. Maybe it has learned my typo habits.
Will fix!

internal/raft/cluster.go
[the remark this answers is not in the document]
alecthomas: If it's only called at creation, can that be enforced with a builder pattern or similar?

PR #3811, opened 2024-12-18
fix: handle pubsub rebalance and closure

matt2e:
fixes https://github.com/block/ftl/issues/3808


internal/log/logger.go
```diff
@@ -47,6 +47,17 @@ func New(level Level, sink Sink) *Logger {
 func (l Logger) Scope(scope string) *Logger {
 	return l.Attrs(map[string]string{scopeKey: scope})
 }
+
+func (l Logger) AppendScope(scope string) *Logger {
```
matt2e, line 51: Setting scope in pubsub was losing runner scope which made pubsub logs harder to follow

PR #3841, opened 2024-12-22
chore: update the JDK

stuartwdouglas:


alecthomas (approved):
Might be some mojo in the Dockerfile too?

PR #3896, opened 2025-01-03
fix: reused enum type schema extraction

### LINKED ISSUE #2857: Reused enum type not schema extracted correctly
When a type is used in 2 Go enums, the schema has that type missing from one of the enums. It's non-deterministic which enum loses. 

```go
//ftl:enum
type Animal interface{ animal() }
type Lion struct{}
type Dog struct{}

func (Lion) animal() {}
func (Dog) animal() {}

//ftl:enum
type Pet interface{ pet() }
type Cat struct{}

func (Cat) pet() {}
func (Dog) pet() {}
```

Will create either of these schemas:
```graphql
  export enum Animal {
    Lion example.Lion
    Dog example.Dog
  }
  export enum Pet {
    Cat example.Cat
  }
```
☝️  Pet.Dog is missing

or:
```graphql
  export enum Animal {
    Lion example.Lion
  }
  export enum Pet {
    Cat example.Cat
    Dog example.Dog
  }
```
☝️  Animal.Dog is missing

wesbillman:
Fixes #2857

This `go` code...
Screenshot 2025-01-03 at 10 08 02 AM

Extracts to this schema now
Screenshot 2025-01-03 at 10 07 46 AM

### DIFF
--- a/go-runtime/schema/enum/analyzer.go
+++ b/go-runtime/schema/enum/analyzer.go
@@ -111,31 +111,33 @@ func validateVariant(pass *analysis.Pass, obj types.Object, variant *schema.Enum
 }
 
 func findTypeValueVariants(pass *analysis.Pass, obj types.Object) []*schema.EnumVariant {
-	var variants []*schema.EnumVariant
+	variantMap := make(map[string]*schema.EnumVariant)
+
 	for vObj, facts := range common.GetAllFactsOfType[*common.MaybeTypeEnumVariant](pass) {
-		// there shouldn't be more than one of this type of fact on an object, but even if there are,
-		// we don't care. We just need to know if there are any.
-		if len(facts) < 1 {
-			continue
-		}
-		fact := facts[0]
+		for _, fact := range facts {
+			// Only process variants that belong to this enum
+			if fact.Parent != obj {
+				continue
+			}
 
-		if fact.Parent != obj {
-			continue
-		}
-		// extract variant type here rather than in the `typeenumvariant` extractor so that we only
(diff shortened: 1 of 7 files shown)


jvmakine (approved):

go-runtime/schema/enum/analyzer.go
```diff
@@ -111,31 +111,38 @@ func validateVariant(pass *analysis.Pass, obj types.Object, variant *schema.Enum
...
-		variants = append(variants, fact.Variant)
+	}
 
+	variants := make([]*schema.EnumVariant, 0, len(variantMap))
+	for _, variant := range variantMap {
+		variants = append(variants, variant)
 	}
+
+	// Sort variants by name for consistent output
+	slices.SortFunc(variants, func(a, b *schema.EnumVariant) int {
```
worstell, line 143: i think we should move this to [here](https://github.com/block/ftl/blob/0af296f98f65d58ea119bdac82b43e878cd9c639/common/schema/validate.go#L251) so that other runtimes all get sorted by the same mechanism as well

worstell (approved):

go-runtime/schema/enum/analyzer.go
wesbillman: Oh nice! great catch, thanks!

wesbillman: Ok, now I think we don't want sorting here at all. As a developer, I might want to control the enum ordering. For example:
```
enum Status {
  SUCCESS = 0
  OTHER_STATUS = 1
  ERROR = 100
}
```

I might want to control the order so that important items (like `SUCCESS` here) stay at the top.

PR #3917, opened 2025-01-07
fix: race condition while collecting runtime events

matt2e:
Fixes a race condition where multiple go routines would try and append to the same slice concurrently.
We would end up losing some RuntimeEvents.

The fix involves publishing to a channel, aggregating the events and then marking the step as complete.

I only started seeing the bug when pubsub provisioning began completing multiple steps in quick succession: https://github.com/block/ftl/pull/3916


jvmakine (approved):
Nice!

PR #3933, opened 2025-01-07
feat: bundle sqlc binaries in ftl release binary

worstell:
also implements the Go API for invoking generate


internal/sqlc/sqlc.go
```diff
@@ -0,0 +1,264 @@
...
+		path   string
+		sha256 string
+		tmpDir string
+		mu     sync.Mutex
+	}
+	initOnce   sync.Once
+	sqlcBinaryName = "sqlc"
+)
+
+// maintain a cache of the SQLC binary/WASM plugin per session of the FTL CLI
```
worstell, line 38: this is kind of gnarly. perhaps longer term we could touch base with the sqlc project about supporting a library

PR #3950, opened 2025-01-09
fix: deepcopy of time

jvmakine:
Seems that location fields are never mutated, so it should be safe to just copy the pointer


common/reflect/reflect.go
```diff
@@ -277,7 +278,16 @@ func copyStruct(x any, ptrs map[uintptr]any, copyConf *copyConfig) any {
 	t := reflect.TypeOf(x)
 	dc := reflect.New(t)
 	for i := 0; i < t.NumField(); i++ {
-		if copyConf.disallowCopyUnexported {
+		if t == reflect.TypeOf(time.Time{}) && v.Field(i).Type() == reflect.TypeOf(&time.Location{}) {
+			// time.Time comparest the location pointers to a hard coded global pointer when determining
```
matt2e, line 282:
```suggestion
			// time.Time compares location pointers to a hard coded global pointer when determining
```

matt2e, line 282:
```suggestion
			// time.Time compares location pointers to a hard coded global pointer when determining
```

matt2e (approved):

common/reflect/reflect.go
matt2e: Oops doubled up

PR #3951, opened 2025-01-09
chore: implement controller state as a raft compatible statemachine

jvmakine:
Implements controller state as a Raft compatible `StateMachine`. The controller events are then calculated by diffing changes in the underlying schemas.

This approach is compatible with Raft where we might not have access to all underlying events.


backend/controller/controller.go
```diff
@@ -1152,3 +1143,43 @@ func validateCallBody(body []byte, verb *schema.Verb, sch *schema.Schema) error
...
+			events = append(events, &state.DeploymentSchemaUpdatedEvent{
+				Key:    deployment.Key,
+				Schema: deployment.Schema,
+			})
+		}
+	}
+
+	currentActive := current.GetActiveDeployments()
+	currentAll := current.GetDeployments()
+	for _, deployment := range previous.GetActiveDeployments() {
```
matt2e, line 1175: I can't quite tell just by looking at this code if it's correct, but it doesnt seem to do what I'd expect.
I'd expect ModuleRemoved to be true if there are no deployments for the module in the new state, but it only seems to be checking if `deployment.Key` is there, so it would be true if a new deployment has replaced it.

internal/channels/broadcast.go
```diff
@@ -0,0 +1,64 @@
+package channels
+
+import (
+	"context"
+	"sync"
+)
+
+// Broadcaster helps to broadcast messages to multiple subscribed channels.
```
matt2e, line 8: can this be replaced with `github.com/alecthomas/types/pubsub`?

backend/controller/controller.go
jvmakine: Yes, that is a bug! Fixing!

internal/channels/broadcast.go
jvmakine: Unfortunately, I could not get that working without it deadlocking in my use case

PR #3953, opened 2025-01-09
feat: send build events over grpc and new lsp cmd

### LINKED ISSUE #3925: See if we can make LSP snippets available even if `ftl dev` isn't running (via the VSCode extension)
Ideally, we could fire up FTL in a terminal (separate from VSCode) and still be able to use features of the extension. Basic features like autocomplete would be awesome.

- [x] auto complete without running ftl dev
- [x] #3963

wesbillman:
Fixes #3925

This decoupling allows us to run the LSP independent of `ftl dev`. This means that our extensions can just start and stop the lsp whenever they want and it will pick up language server updates.

So now we can use VSCode with ftl without having it also start and stop the dev server.


worstell (approved):
so awesome

internal/buildengine/updates.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	return nil
+}
+
+func (u *updatesService) Ping(context.Context, *connect.Request[ftlv1.PingRequest]) (*connect.Response[ftlv1.PingResponse], error) {
+	return connect.NewResponse(&ftlv1.PingResponse{}), nil
+}
+
+func (u *updatesService) StreamEngineEvents(ctx context.Context, req *connect.Request[buildenginepb.StreamEngineEventsRequest], stream *connect.ServerStream[buildenginepb.StreamEngineEventsResponse]) error {
+	events := make(chan *buildenginepb.EngineEvent, 64)
```
matt2e, line 51: I wonder about this case:
- Start `ftl dev` in a terminal
- Build engine fails to build a module, has code errors
- IDE launches, connects to build engine for updates
- IDE thinks everything is all good, rather than receiving the initial state

internal/buildengine/updates.go
```diff
@@ -0,0 +1,110 @@
...
+			// If not a builderrors.Error, create a basic error
+			errs = append(errs, &langpb.Error{
+				Msg:   e.Error(),
+				Level: langpb.Error_ErrorLevel(berr.Level),
+			})
+			continue
+		}
+
+		if berr.Type == builderrors.COMPILER {
+			continue
```
matt2e, line 88: Should this be done buildengine side or LSP side? It feels LSP specific to me

internal/lsp/lsp.go
```diff
@@ -74,80 +72,67 @@ func (s *Server) Run() error {
...
-		if !errors.As(e, &ce) {
-			errUnspecified = append(errUnspecified, err)
-			continue
-		}
-		if ce.Type == builderrors.COMPILER {
+	buildErrors := langpb.ErrorsFromProto(err)
+
+	// Associate by filename.
+	for _, e := range buildErrors {
+		if e.Type == builderrors.COMPILER {
```
matt2e, line 122: ah it is filtering on LSP side anyway

matt2e (approved):
Sweet!

frontend/cli/cmd_lsp.go
```diff
@@ -0,0 +1,75 @@
...
+	client := rpc.Dial(enginepbconnect.NewBuildEngineServiceClient, l.BuildUpdatesEndpoint.String(), log.Error)
+	g.Go(func() error {
+		for {
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			default:
+				logger.Debugf("Connecting to build updates service")
+				if err := l.streamBuildEvents(ctx, client); err != nil {
+					logger.Debugf("Failed to connect to build updates service: %v", err)
```
matt2e, line 44: Similar logic to the comment about initial state in the stream, theres bad states that could arise here:
eg: errors were present in buildengine, lost connection / build engine dies, when we reconnect those errors may not be present anymore but the LSP wont be notified. Maybe we need to reset lsp state here.

internal/buildengine/updates.go
wesbillman: Yeah this will def happen. And actually did happen before as well. :)

We might want to think about the lsp itself being about to trigger a build when it activates to get the current errors. I'm guessing other extensions might do something similar to get their initial state 🤔

internal/buildengine/updates.go
wesbillman: You're totally right! I will remove this so that we can report all the errors 🤦

PR #4015, opened 2025-01-14
feat: add single partition mapper to jvm

matt2e:
closes https://github.com/block/ftl/issues/4002

### DIFF
--- a/backend/runner/pubsub/testdata/java/publisher/src/main/java/xyz/block/ftl/java/test/publisher/Publisher.java
+++ b/backend/runner/pubsub/testdata/java/publisher/src/main/java/xyz/block/ftl/java/test/publisher/Publisher.java
@@ -8,6 +8,7 @@ import xyz.block.ftl.Topic;
 import xyz.block.ftl.TopicPartitionMapper;
 import xyz.block.ftl.Verb;
 import xyz.block.ftl.WriteableTopic;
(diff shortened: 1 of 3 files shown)


jvm-runtime/ftl-runtime/common/deployment/src/main/java/xyz/block/ftl/deployment/TopicsProcessor.java
```diff
@@ -69,7 +71,10 @@ TopicsBuildItem handleTopics(CombinedIndexBuildItem index, BuildProducer<Generat
                                 + " with a concrete type parameter " + iface.name() + " does not extend this interface");
             }
             var partitionMapperClass = partitionMapperType.name();
-            beans.produce(AdditionalBeanBuildItem.unremovableOf(partitionMapperClass.toString()));
+            if (partitionMapperClass.equals(FTLDotNames.SINGLE_PARTITION_MAPPER)) {
```
stuartwdouglas, line 74: This should be `PARTITION_MAPPER` (and the DotNames class updated) assuming you want to make this the default. Otherwise this line does not do anything.

matt2e: Ahhhh. Removing, as I didn't mean to try to make it the default

PR #4059, opened 2025-01-16
fix: default file name for jvm modules

wesbillman:


matt2e:
Can you update these docs as well? https://github.com/block/ftl/blob/main/docs/content/docs/getting-started/quick-start/index.md#L136

stuartwdouglas:
TestLifecycleJVM will also need to be updated.

wesbillman:
Thanks guys! Updated :)

jvm-runtime/testdata/java/gradle/src/main/java/org/acme/Gradle.java
```diff
@@ -3,11 +3,11 @@
 import xyz.block.ftl.Export;
 import xyz.block.ftl.Verb;
 
-public class EchoVerb {
+public class Gradle {
```
matt2e, line 6: Was this a mistake?
```suggestion
public class Echo {
```

wesbillman: The `ftl.toml` here has:

```toml
module = "gradle"
language = "java"
```
so I tried to make them match up.

wesbillman: But it does look a bit odd 😂 I think it's because it's used for `TestGradle`.

matt2e: Ahhh yep yep!

worstell (approved):

PR #4061, opened 2025-01-16
fix: fix race conditions in event stream and controller state

matt2e:
fixes https://github.com/block/ftl/issues/3919
Should fix some edge cases. I experienced a crash yesterday based on this I think.

### DIFF
--- a/backend/controller/state/statemachine.go
+++ b/backend/controller/state/statemachine.go
@@ -3,6 +3,7 @@ package state
 import (
 	"context"
 	"fmt"
+	"sync"
 
 	"github.com/block/ftl/common/reflect"
 	"github.com/block/ftl/internal/channels"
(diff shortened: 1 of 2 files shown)


backend/controller/state/statemachine.go
```diff
@@ -14,15 +15,21 @@ type controllerStateMachine struct {
...
+	c.lock.Lock()
+	defer c.lock.Unlock()
 	return reflect.DeepCopy(c.state), nil
 }
 
 func (c *controllerStateMachine) Publish(msg SchemaEvent) error {
+	c.lock.Lock()
+	defer c.lock.Unlock()
 	var err error
 	c.state, err = msg.Handle(c.state)
```
matt2e, line 34: We were writing `c.state` here while Lookup() could be reading it

internal/eventstream/eventstream.go
```diff
@@ -85,7 +85,10 @@ func (i *inMemoryEventStream[T, E]) Changes(ctx context.Context) (iter.Seq[T], e
 }
 
 func (i *inMemoryEventStream[T, E]) View(ctx context.Context) (T, error) {
-	return i.view, nil
+	i.lock.Lock()
```
matt2e, line 88: `Publish()` safely writes to `i.view` with a lock while `View()` read without a lock

stuartwdouglas (approved):

PR #4072, opened 2025-01-16
fix: run sqlc as a child process

worstell:
errors use os.Exit so this prevents failures from killing the parent process


cmd/ftl-sqlc/main.go
```diff
@@ -0,0 +1,11 @@
...
+
+import (
+	"os"
+
+	sqlc "github.com/sqlc-dev/sqlc/pkg/cli"
+)
+
+func main() {
+	os.Exit(sqlc.Run(os.Args[1:]))
+}
```
wesbillman, line 11: Super nit! But we could add a comment about how this is what the `sqlc` library suggested for future devs to see :)

wesbillman (approved):
Nice!

cmd/ftl-sqlc/main.go
worstell: good call!

PR #4100, opened 2025-01-19
fix: use star imports

stuartwdouglas:
This makes IDE snippets work out of the box


matt2e (approved):

jvmakine (approved):
This will be spicy with some people! 😆

stuartwdouglas:
> This will be spicy with some people! 😆

Honestly this is just to get the demo to flow more smoothly with IDE support. We can revert it after the demo :-)

PR #4142, opened 2025-01-21
fix: use dagre layout for graph pane

wesbillman:
Screenshot 2025-01-21 at 1 06 31 PM


wesbillman:
We use ReactFlow now :)

Issue #4149, opened 2025-01-22
Paused timeline has no entries


bradleydwyer:

bradleydwyer:
Also affects time ranged (non-live/streamed) entries.

wesbillman:
Fixed with #4150

PR #4150, opened 2025-01-22
fix: Timeline events missing while paused or non-streaming.

bradleydwyer:


bradleydwyer:
Fixes #4149

wesbillman (approved):
Thanks @bradleydwyer !

PR #4164, opened 2025-01-23
fix: record metadata correctly for HTTP methods

### LINKED ISSUE #4148: JVM HTTP methods don't record metadata correctly
This is a different code path to the normal Verb code path and it is missing call and other usage metadata.

stuartwdouglas:
fixes #4148


stuartwdouglas:
All the JVM schema extraction really needs some tests, but that can come later.

PR #4232, opened 2025-01-29
fix: silence unwanted runner terminated errors

matt2e:
Fixes https://github.com/block/ftl/issues/4229
Some more info:
- The old runner is terminated properly, but there seems to be some race where the context cancellation cause is propagated places sometimes.
    - eg: `https://github.com/block/ftl/blob/main/backend/runner/runner.go#L385`
- We had checks in place to ignore the errors if they were context cancelled errors, but we have moved away from using context cancelled errors.

### DIFF
--- a/backend/runner/runner.go
+++ b/backend/runner/runner.go
@@ -79,7 +79,7 @@ type Config struct {
 
 func Start(ctx context.Context, config Config, storage *artefacts.OCIArtefactService) error {
 	ctx, doneFunc := context.WithCancelCause(ctx)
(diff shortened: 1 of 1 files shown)


matt2e:
@alecthomas is this solution going backwards though, given that we are trying to move to canceling with cause? Not sure if this is a legit use case for standard cancelation

matt2e:
Updated the code to make it a unique error still, but now it wraps `context.Canceled` so the checks still work.

PR #4243, opened 2025-01-30
fix: disallow modules to use reserved names

matt2e:
We can't compile go modules if module names use reserved go keywords.
Previously we checked for this in `ftl new` but modules can be manually renamed.


matt2e:
Turns out it is fine for node names... Because in go they are all UpperCamelCase whereas they are lowerCamelCase in schema... So it really should only apply to module names, which are lowercase in go

PR #4253, opened 2025-01-31
feat: changesets

stuartwdouglas:
This commit introduces the concept of a changeset. All modules are now deployed in a single changeset, which has removed some of the ordering invariants around deployments that we previously had with dependencies. For now this has been worked around with a 'retry until it works' approach, however significant further changes are required to get changesets to the desired end state.


frontend/console/e2e/cron.spec.ts
```diff
@@ -28,6 +29,7 @@ test('send cron request', async ({ page }) => {
 test('submit cron form using ⌘+⏎ shortcut', async ({ page }) => {
   await navigateToDecl(page, 'cron', 'thirtySeconds')
 
+  await setVerbRequestBody(page, '{}')
```
wesbillman, line 32: @stuartwdouglas I think this is why the test was failing before. This bug probably existed before your PR btw, we just never hit the failure case here. Basically it was pressing the keyboard shortcut before the page had fully loaded and the request body was available. This should fix that.

frontend/console/e2e/global-setup.ts
```diff
@@ -8,19 +8,25 @@ const globalSetup = async (config: FullConfig) => {
...
-      const loadedModules = modules.filter((module) => document.querySelector(`li#module-tree-module-${module}`) !== null)
-      console.log('Loaded modules:', loadedModules.join(', '))
-      return loadedModules.length === modules.length
+      const readyModules = modules.filter((module) => {
+        const moduleElement = document.querySelector(`li#module-tree-module-${module}`)
+        if (!moduleElement) return false
+
+        const greenDot = moduleElement.querySelector('.bg-green-400')
+        return greenDot !== null
+      })
```
wesbillman, line 21: This will make the "ready check" make sure the green dots are visible for all the modules we're starting.

internal/buildengine/updates.go
```diff
@@ -78,7 +78,12 @@ func (u *updatesService) Ping(context.Context, *connect.Request[ftlv1.PingReques
 }
 
 func (u *updatesService) StreamEngineEvents(ctx context.Context, req *connect.Request[buildenginepb.StreamEngineEventsRequest], stream *connect.ServerStream[buildenginepb.StreamEngineEventsResponse]) error {
-	// Only send cached events if replay_history is true
+	// First subscribe to new events to avoid missing any
+	events := make(chan *buildenginepb.EngineEvent, 64)
+	u.engine.EngineUpdates.Subscribe(events)
+	defer u.engine.EngineUpdates.Unsubscribe(events)
```
wesbillman, line 84: I think this is just better before we play the history. I'm not sure if this was part of the blue dots showing when the terminal is happy, but thought I'd point it out.

PR #4271, opened 2025-02-03
fix: flaky cron test

### LINKED ISSUE #4237: Look into flaky `TestCron`
https://github.com/block/ftl/actions/runs/13042754965/job/36387948927?pr=4233

```
             },
             Verb: &schemapb.Ref{
               Module: "echo",
        -      Name: "time",
        +      Name: "echo",
             },
             Body: []byte("{}"),
           },
```

wesbillman:
Fixes #4237

Hopefully. I couldn't reproduce locally.

What I suspect was happening is that the 2 crons were running at the same schedule so it could be possible for `echo` to fire twice before `time` gets a chance to run.


matt2e (approved):
Makes sense!

PR #4273, opened 2025-02-03
feat: Engine deploy routine

### LINKED ISSUE #4257: Build engine: consider single go routine for deploys
With changesets I think we need to make some changes to the buildengine because:
- We only allow a single changeset at once, so the buildengine will fail to deploy anything if it has another deploy going on
- Even when we allow multiple changesets at once, it wouldn't be too hard to hit a case where dependant modules cause changesets to be rejected

Maybe we should make to the build engine queue up all modules that need to be deployed, grouping all  pending changes into a changeset when the previous changeset completes.

wesbillman:
Fixes #4257

### DIFF
--- a/internal/buildengine/engine.go
+++ b/internal/buildengine/engine.go
@@ -110,6 +110,11 @@ type Engine struct {
 
 	devModeEndpointUpdates chan dev.LocalEndpoint
 	devMode                bool
+
+	// deployment queue and state tracking
+	deploymentQueue    chan []Module
+	activeChangesetKey optional.Option[string]
+	pendingDeployments []Module
 }
 
 type Option func(o *Engine)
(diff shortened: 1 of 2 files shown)


internal/buildengine/engine.go
```diff
@@ -1308,3 +1294,79 @@ func (e *Engine) watchForPluginEvents(originalCtx context.Context) {
...
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+
+		case modules := <-e.deploymentQueue:
+			// If we have an active changeset, add to pending deployments
+			if _, ok := e.activeChangesetKey.Get(); ok {
+				e.pendingDeployments = append(e.pendingDeployments, modules...)
```
stuartwdouglas, line 1313: What triggers the pending deployments to Deploy if no further events hit the deploymentQueue?

stuartwdouglas: Actually never mind, I see they get copied back into the queue when the changeset ends.

internal/buildengine/engine.go
```diff
@@ -1308,3 +1294,79 @@ func (e *Engine) watchForPluginEvents(originalCtx context.Context) {
...
+							Module: module.Config.Module,
+						},
+					},
+				}
+			}
+
+		case event := <-schemaChanges:
+			// Watch for changeset completion events
+			switch event := event.(type) {
+			case schemaeventsource.EventChangesetEnded:
```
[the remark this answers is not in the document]
wesbillman: yeah, this def needs some more attention. Lemme see if I can fix this up. Thanks for the review!

PR #4276, opened 2025-02-03
fix: engine use single routine for deploys

wesbillman:


internal/buildengine/engine.go
```diff
@@ -786,6 +835,58 @@ func (e *Engine) getDependentModuleNames(moduleName string) []string {
...
+				moduleList := make([]Module, len(modules))
+				for i, qm := range modules {
+					moduleList[i] = qm.module
+				}
+
+				err := Deploy(ctx, e.projectConfig, moduleList, settings.replicas, settings.waitForDeployOnline, e.deployClient, e.schemaServiceClient)
+
+				// Signal completion to all modules in this group
+				for _, qm := range modules {
+					if err != nil {
```
matt2e, line 878: This if check can be skipped, as in both cases you are sending the value of `err`

internal/buildengine/engine.go
```diff
@@ -786,6 +835,58 @@ func (e *Engine) getDependentModuleNames(moduleName string) []string {
...
+func (e *Engine) processDeployments(ctx context.Context) {
+	logger := log.FromContext(ctx)
+	ticker := time.NewTicker(100 * time.Millisecond)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
```
stuartwdouglas, line 848: I preferred the other approach, as much as possible we should be trying to avoid using a ticker vs just reading directly from the channel. 100ms isn't much, but if they start poping up everywhere they eventually really start to slow things down.

PR #4319, opened 2025-02-05
feat: use runtime elements for updates

stuartwdouglas:
This removes the use of raft events from the API and provisioner implementation, and instead moves to the use of runtime elements directly.


alecthomas (approved):
Nice!

PR #4389, opened 2025-02-12
fix: issue with rebuilds not happening on sqlc errors

### LINKED ISSUE #4371: FTL dev does not rebuild failed modules if cause was sqlc
Steps:
- Create a module
- Create invalid queries in the sqlc queries file
- Run `ftl dev`
- Fix the queries

Expected result:
- FTL dev will detect the change rebuild the module
Actual result:
- No auto rebuild will happen.

Looks like the reason for this is that the build engine does the initial sqlc generation, but automatic rebuilds happen within the language plugin. This means that if the build fails before the language plugin is told to start a build stream the module won't be looking for file changes.

wesbillman:
Fixes #4371

### DIFF
--- a/internal/buildengine/build.go
+++ b/internal/buildengine/build.go
@@ -34,10 +34,15 @@ func build(ctx context.Context, plugin *languageplugin.LanguagePlugin, projectCo
 
 	_, err = sqlc.AddQueriesToSchema(ctx, projectConfig.Root(), bctx.Config.Abs(), bctx.Schema)
 	if err != nil {
-		return nil, nil, fmt.Errorf("failed to add queries to schema: %w", err)
+		logger.Errorf(err, "Failed to add queries to schema")
+		// Continue with build even if sqlc generation fails
 	}
+
 	stubsRoot := stubsLanguageDir(projectConfig.Root(), bctx.Config.Language)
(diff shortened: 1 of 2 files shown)


internal/buildengine/build.go
```diff
@@ -34,10 +34,15 @@ func build(ctx context.Context, plugin *languageplugin.LanguagePlugin, projectCo
 
 	_, err = sqlc.AddQueriesToSchema(ctx, projectConfig.Root(), bctx.Config.Abs(), bctx.Schema)
 	if err != nil {
-		return nil, nil, fmt.Errorf("failed to add queries to schema: %w", err)
+		logger.Errorf(err, "Failed to add queries to schema")
+		// Continue with build even if sqlc generation fails
```
worstell, line 38: i think we might want to fail outright here? 

on the other hand, if you have any dependencies to these queries in your code (e.g. injected into another verb) the build will still fail. so maybe this is ok

wesbillman: I think we have to continue here otherwise it seems like the build loop stops and doesn't recover.

wesbillman: I'm still testing this a bit here too. I think it also just doesn't show failed sqlc generation errors in some cases

wesbillman:
@worstell maybe we should wait for your inflight changes and then revisit this PR. I don't think this is fixing it in a stable way because sometimes the build completes without an error even though the `queries.sql` has invalid code. I'm gonna close and we can take a stab at this one later.

PR #4390, opened 2025-02-12
feat: make services optional

stuartwdouglas:


stuartwdouglas:
closing this for now, as it is no longer needed

PR #4452, opened 2025-02-17
feat: CLI talks to admin service

stuartwdouglas:


cmd/ftl/cmd_dev.go
```diff
@@ -42,9 +41,8 @@ func (d *devCmd) Run(
 	bindContext terminal.KongContextBinder,
 	schemaEventSource *schemaeventsource.EventSource,
 	controllerClient ftlv1connect.ControllerServiceClient,
-	schemaServiceClient ftlv1connect.SchemaServiceClient,
 	timelineClient *timelineclient.Client,
-	adminClient admin.Client,
```
stuartwdouglas, line 47: This change actually speeds up startup by 500ms, as previously the admin client would try and connect before the cluster started and then fall back to a local client.

Issue #4451, opened 2025-02-17
Build failures sometimes cause valid modules to get stuck in an undeployed state.


bradleydwyer:
If you have a dependency graph of the modules A, B and C

A
B -> A
C -> A

If a build failure occurs in C, sometimes both B and A will silently fail to deploy.

e.g.

source
sinkkt depends on source
sinkgo depends on source

Type references from source in sinkgo are failing.

```
test1🐚 ➜  test1 git:(main) ✗ ftl dev
info:console: Web console available at: http://127.0.0.1:8899
info:http-ingress: HTTP ingress server listening on: http://127.0.0.1:8891
info: Controller started in 0.05s
info:source:build: Building module
info:source: Using dev mode build command 'mvn -Dquarkus.console.enabled=false -q clean quarkus:dev -Ddebug=57694 -Dftl.language.port=57695'
info:source: 
info:source: ----------------------------
info:source: --- Help improve Quarkus ---
info:source: ----------------------------
info:source: * Learn more: https://quarkus.io/usage/
info:source: * Do you agree to contribute anonymous build time data to the Quarkus community? (y/n and enter) 
info:source: [info] [Quarkus build analytics] Didn't receive a valid user's answer: `y` or `n`. The question will be asked again next time.
info:source: 
info:source: Listening for transport dt_socket at address: 57694
info:source: ANTLR Tool version 4.11.1 used for code generation does not match the current runtime version 4.13.2
info:source: ANTLR Runtime version 4.11.1 used for parser compilation does not match the current runtime version 4.13.2
info:source: ANTLR Tool version 4.11.1 used for code generation does not match the current runtime version 4.13.2
info:source: ANTLR Runtime version 4.11.1 used for parser compilation does not match the current runtime version 4.13.2
info:source: 2025-02-17 16:15:42,446 INFO  [xyz.blo.ftl.dep.HotReloadHandler] (build-62) Starting Hot Reload gRPC server on port 57695
info:source:build: Module built (12.87s)
info:sinkkt:build: Building module
info:sinkgo:build: Building module
info:sinkkt: Using dev mode build command 'mvn -Dquarkus.console.enabled=false -q clean quarkus:dev -Ddebug=57730 -Dftl.language.port=57731'
info:sinkgo:build: Module built (1.42s)
info:sinkkt: 
info:sinkkt: ----------------------------
info:sinkkt: --- Help improve Quarkus ---
info:sinkkt: ----------------------------
[...]

PR #4454, opened 2025-02-17
fix: use changeset per module in dev mode

stuartwdouglas:
fixes: #4451

This needs more work.

### DIFF
--- a/internal/buildengine/engine.go
+++ b/internal/buildengine/engine.go
@@ -80,9 +80,11 @@ type autoRebuildCompletedEvent struct {
 }
 
 type pendingDeploy struct {
-	modules  []Module
-	err      chan error
-	replicas int32
+	modules           []Module
+	err               chan error
+	replicas          int32
+	superseded        bool
+	supercededModules []*pendingDeploy
 }
 
 func (autoRebuildCompletedEvent) rebuildEvent() {}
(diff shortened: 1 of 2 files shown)


internal/buildengine/engine.go
```diff
@@ -863,14 +865,18 @@ func (e *Engine) BuildAndDeploy(ctx context.Context, replicas int32, waitForDepl
...
 			e.modulesToBuild.Store(module.Config.Module, false)
-			modulesToDeploy = append(modulesToDeploy, module)
 			e.rawEngineUpdates <- &buildenginepb.EngineEvent{
 				Event: &buildenginepb.EngineEvent_ModuleDeployWaiting{
 					ModuleDeployWaiting: &buildenginepb.ModuleDeployWaiting{
 						Module: module.Config.Module,
 					},
 				},
 			}
+			if singleChangeset {
```
matt2e, line 875: supernit: Ugly if statement we can put `return nil` in the if case and remove the `else`

internal/buildengine/engine.go
```diff
@@ -1340,84 +1349,127 @@ func (e *Engine) deploy(ctx context.Context, modules []Module, replicas int32, w
...
-					more = false
+			// Try and deploy, unless there are conflicting changesets this will happen immediately
+			if !e.tryDeployFromQueue(ctx, &deployment) {
+				// We could not deploy, add to the list of pending deployments
+				// But first check if there are older deployments that are superceded by this one
+				modules := map[string]bool{}
+				for _, module := range deployment.modules {
+					modules[module.Config.Module] = true
 				}
+				for _, existing := range toDeploy {
```
matt2e, line 1380: This would be easier if `toDeploy` was a map rather than a slice

PR #4462, opened 2025-02-18
chore: manage generated schema from build engine

worstell:


matt2e:
If the whole sql schema step is language agnostic I wonder if it'd be simpler for it to be included in the BuildContext, rather than each plugin needing to do a back and forth on each build.
- Currently build engine watches all module tomls. We could extend that to migration + queries
- Build engine sends a new build context whenever they change
I think of build context change as "please build, and he's all the extra info you need", which sql verbs fit into.

BTW it is currently optional for BuildStarted to be called by language plugins, but this would make it required (which is fine if we want that)

worstell:
> If the whole sql schema step is language agnostic I wonder if it'd be simpler for it to be included in the BuildContext, rather than each plugin needing to do a back and forth on each build.
> 
> * Currently build engine watches all module tomls. We could extend that to migration + queries
> * Build engine sends a new build context whenever they change
>   I think of build context change as "please build, and he's all the extra info you need", which sql verbs fit into.
> 
> BTW it is currently optional for BuildStarted to be called by language plugins, but this would make it required (which is fine if we want that)

that makes sense to me! i saw the build context API but wondered if we wanted that to be used exclusively for explicit builds rather than adapted to also work for automatic rebuilds- but agree that conceptually it makes sense for these to be communicated via build context updates

will update

worstell:
> > If the whole sql schema step is language agnostic I wonder if it'd be simpler for it to be included in the BuildContext, rather than each plugin needing to do a back and forth on each build.
> > 
> > * Currently build engine watches all module tomls. We could extend that to migration + queries
> > * Build engine sends a new build context whenever they change
> >   I think of build context change as "please build, and he's all the extra info you need", which sql verbs fit into.
> > 
> > BTW it is currently optional for BuildStarted to be called by language plugins, but this would make it required (which is fine if we want that)
> 
> that makes sense to me! i saw the build context API but wondered if we wanted that to be used exclusively for explicit builds rather than adapted to also work for automatic rebuilds- but agree that conceptually it makes sense for these to be communicated via build context updates
> 
> will update

oh 

> > If the whole sql schema step is language agnostic I wonder if it'd be simpler for it to be included in the BuildContext, rather than each plugin needing to do a back and forth on each build.
> > 
> > * Currently build engine watches all module tomls. We could extend that to migration + queries
> > * Build engine sends a new build context whenever they change
> >   I think of build context change as "please build, and he's all the extra info you need", which sql verbs fit into.
> > 
> > BTW it is currently optional for BuildStarted to be called by language plugins, but this would make it required (which is fine if we want that)
> 
> that makes sense to me! i saw the build context API but wondered if we wanted that to be used exclusively for explicit builds rather than adapted to also work for automatic rebuilds- but agree that conceptually it makes sense for these to be communicated via build context updates
> 
> will update

oh i misunderstood, i see what you're saying now about the toml watcher- will give that a shot

internal/watch/watch.go
```diff
@@ -293,3 +303,19 @@ func (t *modifyFilesTransaction) ModifiedFiles(paths ...string) error {
 
 	return nil
 }
+
+// excludeFromAutoRebuilds checks if a changed file is exclusively managed by the build engine and should not be propogated
+// as a change to language plugins.
+func excludeFromAutoRebuilds(path string, config moduleconfig.UnvalidatedModuleConfig) bool {
```
matt2e, line 309: This feels out of place as this is meant to be a common utility, not language plugin specific. Which made me think, plugins just shouldnt be watching `**/*.sql` right? so we could just remove it from here:
go: https://github.com/block/ftl/blob/main/go-runtime/goplugin/service.go#L180
and then same for jvm.
That way the engine would still get a list of changes of which files changed, which would eventually help us when we want the engine to merge of migration together to detect if we need to reset the db (another issue filed somewhere in github). It could make sure the changed sql is in the migrations dir before doing those checks.

matt2e (changes requested):
A lot cleaner!! Looking good @worstell

internal/watch/watch.go
worstell: ah yeah i actually originally added this in the module config, but then walked it back because i was trying to limit the number of places where you'd need to be aware of the detail that build engine handles these files and exclude them. but you're right, feels wrong to have a changed update that excludes some of the changes

matt2e (approved):
@worstell The tests were failing because the glob filter wasn't quite working right in the go plugin (I didnt check the jvm one). I then went down the path of simplifying things a bit further (I dont think we need plugins to watch for `**/*.sql` and then ignore any detected changes to sql). I'm not sure if this is missing any cases you had in mind?

I didn't want to lose the glob stuff you wrote so i pushed it under another branch so its not lost: `matt2e/old-worstell-glob`

PR #4467, opened 2025-02-19
fix: also wait for module dependencies

stuartwdouglas:
Otherwise echo might attempt to deploy before time is ready

### DIFF
--- a/internal/buildengine/engine.go
+++ b/internal/buildengine/engine.go
@@ -1370,7 +1370,12 @@ func (e *Engine) processDeploymentQueue(ctx context.Context) {
 
 		case deployment := <-e.deploymentQueue:
 			// Try and deploy, unless there are conflicting changesets this will happen immediately
-			if !e.tryDeployFromQueue(ctx, &deployment) {
+			graph, err := e.Graph()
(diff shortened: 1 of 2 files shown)


internal/buildengine/engine.go
```diff
@@ -1414,6 +1414,9 @@ func (e *Engine) tryDeployFromQueue(ctx context.Context, deployment *pendingDepl
 	modules := map[string]bool{}
 	for _, module := range deployment.modules {
 		modules[module.Config.Module] = true
+		for _, dep := range module.dependencies {
+			modules[dep] = true
```
matt2e, line 1418: Do we need to be recursive?
If C depends on B depends on A... A is in an active changeset, B is waiting, C finishes building... it should also wait

stuartwdouglas: I don't think so, although maybe we need to also check the build queue to see if direct dependencies are queued up. If the A and C changes are unrelated then this is fine, but if the sequence was A triggers B which triggers C then we want C to wait because B is in the queue.

matt2e: makes sense to me

PR #4469, opened 2025-02-19
fix: collect ftl build logs and format output

matt2e:
- Captures errors from jvm build logs (like we do already for `ftl dev` build logs)
- Format the output from jvm build to be consistent with the rest of our logs

### DIFF
--- a/jvm-runtime/plugin/common/error_detector.go
+++ b/jvm-runtime/plugin/common/error_detector.go
@@ -1,9 +1,10 @@
 package common
 
 import (
-	"fmt"
 	"io"
 	"strings"
+
+	"github.com/block/ftl/internal/log"
 
 	"github.com/alecthomas/atomic"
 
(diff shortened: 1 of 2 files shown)


jvm-runtime/plugin/common/error_detector.go
```diff
@@ -1,28 +1,46 @@
...
 }
 
 func (o *errorDetector) Write(p []byte) (n int, err error) {
 	// Forward output to stdout
-	fmt.Printf("%s", string(p))
+	for _, line := range strings.Split(string(p), "\n") {
+		if len(line) == 0 {
+			continue
+		}
+		if cleanLine, ok := strings.CutPrefix(line, "[ERROR] "); ok {
```
stuartwdouglas, line 30: Maybe this should have been a regex?

stuartwdouglas (approved):

jvm-runtime/plugin/common/error_detector.go
matt2e: We can improve it next... because sometimes I see logs with `<timestamp> [INFO] ...` and othertimes I see `<timestamp> INFO` I think? I'm not sure what is causing which logs and what we should handle

Issue #4470, opened 2025-02-19
Modules created in CWD instead of project directory


bradleydwyer:
When running `ftl dev` with a path to the project location - if you use the inline prompt to create a module, it places it in the CWD, not the project directory.

```
ftl🐚 ➜  ftl git:(main) ✗ pwd
/Users/bradleydwyer/Development/ftl/ftl

ftl🐚 ➜  ftl git:(main) ✗ ftl dev ~/Development/ftl-cwd
info:console: Building console...
info:http-ingress: HTTP ingress server listening on: http://127.0.0.1:8891
info:console: Web console available at: http://127.0.0.1:8899
info: Controller started in 0.05s
info:http-ingress: FTL startup command ⚡️
new kotlin cwdmodule
info: acquired file lock for /Users/bradleydwyer/Development/ftl/ftl/.ftl/modules.lock
Successfully created kotlin module "cwdmodule" in /Users/bradleydwyer/Development/ftl/ftl/cwdmodule
>  
```

matt2e:
Is this still an issue in the latest `matt2e/goose/progress` @bradleydwyer? Yesterday I made module creation fail if it isn't within the dev directory, with an explanation of why. And goose seems to pick it up in most cases. (though it fails if it tries to do `cd <path> & ftl new ...`)

bradleydwyer:
```
error: module directory /Users/bradleydwyer/Development/allhands/ftl/foo2 is not within the expected module directories (/Users/bradleydwyer/Development/allhands/test4). Please choose an appropriate path or force create the module outside of the expected directories by using the --force argument
```

So yeah - prevents this on that branch.

matt2e:
It should allow you to do it?
```
mkdir example
ftl dev example
...
> new go mymodule example
```
You just need to provide a path from your cwd to where you want the modules. Did goose not end up creating the module in the right place for you?

bradleydwyer:
This wasn't with Goose. Just manually. I wasn't aware of the 3rd parameter, so I didn't include it 🤦🏻

matt2e:
All good! I'll make a PR for the fix thats currently in the goose branch

PR #4471, opened 2025-02-19
fix: modules created in incorrect directory

stuartwdouglas:
fixes: #4470


matt2e:
Why set the cwd to dev's dir and not just stop resetting cwd to $gitroot? Or is that not what the underlying is?

Issue #4536, opened 2025-02-21
ftl --version exit code is 130


bradleydwyer:
`ftl --version` is returning exit code 130 instead of 0 in the latest release 0.451.2

PR #4537, opened 2025-02-21
fix: install signal handler earlier

stuartwdouglas:
fixes: #4536

### DIFF
--- a/cmd/ftl/main.go
+++ b/cmd/ftl/main.go
@@ -13,6 +13,7 @@ import (
 	"strconv"
 	"syscall"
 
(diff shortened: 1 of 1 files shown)


cmd/ftl/main.go
```diff
@@ -219,6 +221,7 @@ func addToExit(k *kong.Kong, cleanup func(code int)) {
 	k.Exit = func(code int) {
 		cleanup(code)
 		if originalExit == nil {
+			println("exit 3")
```
matt2e, line 224: was this a debug line that accidentally made it in?

matt2e (approved):

PR #4661, opened 2025-02-26
fix: validate built module against schema

matt2e:
Build plugins could claim a module was built when it was actually invalid when validated against the build schema.

### DIFF
--- a/common/schema/validate.go
+++ b/common/schema/validate.go
@@ -57,10 +57,26 @@ func (s *Schema) Validate() (*Schema, error) {
 
 // ValidateModuleInSchema clones and normalises a schema and semantically validates a single module within it.
 // If no module is provided, all modules in the schema are validated.
+// m can be a new or updated module that will be added to the schema before validation.
 //
 //nolint:maintidx
 func ValidateModuleInSchema(schema *Schema, m optional.Option[*Module]) (*Schema, error) {
 	schema = dc.DeepCopy(schema)
+
+	if m, ok := m.Get(); ok {
+		// Replace original version of module with new version in case they differ
+		var found bool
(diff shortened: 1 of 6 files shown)


common/schema/validate.go
```diff
@@ -57,10 +57,26 @@ func (s *Schema) Validate() (*Schema, error) {
 
 // ValidateModuleInSchema clones and normalises a schema and semantically validates a single module within it.
 // If no module is provided, all modules in the schema are validated.
+// m can be a new or updated module that will be added to the schema before validation.
 //
 //nolint:maintidx
 func ValidateModuleInSchema(schema *Schema, m optional.Option[*Module]) (*Schema, error) {
```
matt2e, line 63: Previously this was not actually validating `m` but rather the matching module in `schema` for its name.
If we wanted that we should probably just pass an optional name into this func.

But also to achieve what we want in this PR it is very convenient to just pass in the module we want to validate with the schema and have it all work.

internal/buildengine/stubs.go
```diff
@@ -48,10 +48,13 @@ func CleanStubs(ctx context.Context, projectRoot string) error {
 func SyncStubReferences(ctx context.Context, projectRoot string, moduleNames []string, metas map[string]moduleMeta, view *schema.Schema) error {
 	wg, wgctx := errgroup.WithContext(ctx)
 	for _, meta := range metas {
-		stubsRoot := stubsLanguageDir(projectRoot, meta.module.Config.Language)
```
matt2e, line 51: One module failing here should not prevent others from being attempted.

stuartwdouglas (approved):

PR #4669, opened 2025-02-26
fix: disable autobuild in VSCode

stuartwdouglas:


internal/projectinit/scaffolding/.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "java.autobuild.enabled":false
+}
```
wesbillman, line 3: Uber nit! but we might want to format this file so it doesn't get reformatted in users projects?

```json
{
  "java.autobuild.enabled": false
}

```

wesbillman (approved):

internal/projectinit/scaffolding/.vscode/settings.json
wesbillman, line 3: Uber nit! but we might want to format this file so it doesn't get reformatted in users projects?

```suggestion
{
  "java.autobuild.enabled": false
}
```

wesbillman (approved):

PR #4679, opened 2025-02-26
fix: handle newline better

stuartwdouglas:


internal/terminal/status.go
```diff
@@ -299,9 +299,7 @@ func (r *terminalStatusManager) consoleNewline(line string) {
 		r.underlyingWrite(ansiUpOneLine + ansiClearLine)
 	}
 	r.underlyingWrite("\r" + line + "\n")
-	if line == "" {
-		r.redrawStatus()
-	}
+	r.redrawStatus()
```
wesbillman, line 302: Nice! Should we remove this code in `build.go` as well now that we have this change?

https://github.com/block/ftl/blob/main/go-runtime/compile/build.go#L615-L617

wesbillman (approved):

internal/terminal/status.go
stuartwdouglas: This only affects when you run a command in the console. I will look into the build.go issue.

PR #4684, opened 2025-02-26
test: add integration test for go compile error parsing

matt2e:
Need to fix this up.
Currently it seems we are not getting the full logs back (cut short as program exits)


matt2e:
Closing this as I have added unit tests instead

PR #4697, opened 2025-02-27
feat: recreate DB on schema change

stuartwdouglas:


common/schema/database.go
```diff
@@ -60,8 +60,9 @@ func (d *Database) GetProvisioned() ResourceSet {
 	migration, ok := slices.FindVariant[*MetadataSQLMigration](d.Metadata)
 	if ok {
 		result = append(result, &ProvisionedResource{
-			Kind:   ResourceTypeSQLMigration,
-			Config: &Database{Type: d.Type, Metadata: []Metadata{migration}},
+			Kind:               ResourceTypeSQLMigration,
+			Config:             &Database{Type: d.Type, Metadata: []Metadata{migration}},
+			DeploymentSpecific: true,
```
stuartwdouglas, line 65: This means that provision will always be called, even when databases are the same. I'm not sure what impact this would have for the other DB provisioners.

jvmakine (approved):

PR #4704, opened 2025-02-27
fix: more commands have autocomplete

matt2e:


cmd/ftl/cmd_sql_migration_new.go
```diff
@@ -31,32 +31,22 @@ func (i migrationSQLCmd) Run(ctx context.Context) error {
 		return fmt.Errorf("could not discover modules: %w", err)
 	}
 	var module *moduleconfig.UnvalidatedModuleConfig
-	parts := strings.Split(i.Datasource, ".")
```
matt2e, line 34: I don't think we want to have optimisations for if we're in the special case of 1 module + 1 db do we?

PR #4707, opened 2025-02-27
feat: open cmd

matt2e:
`ftl open <ref> <editor>`

IDEs:
- `vscode`
- `intellij`

Editor is autodetected within vscode and intellij terminals.

TODO:
- Rename `open` -> `edit`
- Remove auto
- Make this its own package so its easy for console to use it too
- Use this to open editors: https://github.com/Binbiubiubiu/launch_editor/

### DIFF
--- a/cmd/ftl/cmd_open.go
+++ b/cmd/ftl/cmd_open.go
@@ -1,0 +1,121 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"strings"
+
+	"connectrpc.com/connect"
+	"github.com/alecthomas/kong"
+	ftlv1 "github.com/block/ftl/backend/protos/xyz/block/ftl/v1"
+	"github.com/block/ftl/backend/protos/xyz/block/ftl/v1/ftlv1connect"
+	"github.com/block/ftl/common/reflection"
+	"github.com/block/ftl/common/schema"
+	"github.com/block/ftl/internal/exec"
+	"github.com/block/ftl/internal/log"
+	"github.com/block/ftl/internal/projectconfig"
+)
+
+type openCmd struct {
+	Ref    reflection.Ref `arg:"" help:"Language of the module to create." placeholder:"MODULE.ITEM" predictor:"decls"`
+	Editor string         `arg:"" help:"Editor to open the file with." enum:"auto,vscode,intellij" default:"auto"`
+
+	TerminalProgram  string `help:"Terminal program this command is running in which can influence 'auto' editor" env:"TERM_PROGRAM" hidden:""`
+	TerminalEmulator string `help:"Terminal emulator can influence 'auto' editor" env:"TERMINAL_EMULATOR" hidden:""`
+}
+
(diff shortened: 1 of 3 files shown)


cmd/ftl/main.go
```diff
@@ -69,6 +69,7 @@ type SharedCLI struct {
 	Goose     gooseCmd     `cmd:"" help:"Run a goose command."`
 	Mysql     mySQLCmd     `cmd:"" help:"Manage MySQL databases."`
 	Postgres  postgresCmd  `cmd:"" help:"Manage PostgreSQL databases."`
+	Open      openCmd      `cmd:"" help:"Open a file in the editor at the location of a schema declaration."`
```
matt2e, line 72: eh reword this like a human

cmd/ftl/cmd_open.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/block/ftl/internal/log"
+	"github.com/block/ftl/internal/projectconfig"
+)
+
+type openCmd struct {
+	Ref    reflection.Ref `arg:"" help:"Language of the module to create." placeholder:"MODULE.ITEM" predictor:"decls"`
+	Editor string         `arg:"" help:"Editor to open the file with." enum:"auto,vscode,intellij" default:"auto"`
+
+	TerminalProgram  string `help:"Terminal program this command is running in which can influence 'auto' editor" env:"TERM_PROGRAM" hidden:""`
+	TerminalEmulator string `help:"Terminal emulator can influence 'auto' editor" env:"TERMINAL_EMULATOR" hidden:""`
```
alecthomas, line 26: Wut?

matt2e: Yehhh just reading envars to determine if the user is inside VSCode or IntelliJ terminals. Each editor indicates it in different ways. Got a better idea?

alecthomas: Ah I see. There's no good solution to this AFAIK, but I think this would be better if it was in a dedicated "editor" package that could detect the user's editor, or have it configured.

alecthomas:
I also don't know if we should call this "open"... as a user I would expect that to open the console. Maybe "ftl edit" would be better?

cmd/ftl/cmd_open.go
```diff
@@ -0,0 +1,89 @@
...
+
+	if i.Editor == autodetect {
+		if i.TerminalProgram == "vscode" {
+			i.Editor = vscode
+		} else if strings.Contains(i.TerminalEmulator, "JetBrains") {
+			i.Editor = intellij
+		} else {
+			return errors.Errorf("could not auto choose default editor, expected one of %s", strings.Join([]string{vscode, intellij}, ", "))
+		}
+	}
```
alecthomas, line 60: Yeah can we pull this out into a dedicated package and make it extensible instead?

alecthomas: Or maybe just move this into the user's config?

It's very difficult and non-cross-platform to open an editor automatically, so maybe we just punt on it...

For example, this won't work for me at all as I don't use vscode or JetBrains.

matt2e: Yeh puntings easiest

matt2e:
Wes ended up finishing this off in other branches 🏆

PR #4712, opened 2025-02-27
fix: cron scheduled events

### LINKED ISSUE #4708: CronScheduled event is being fired at the wrong time
This event is fired every time the 'next job' is calculated, this means:
- It is only fired for the very next job, other jobs get no events
- It is fired on every schema event even when nothing has changed
- It is fired from every node in the stateful set

We need to figure out what this event should mean and when it should be fired.

wesbillman:
Fixes #4708

These changes should ensure that the CronScheduled event is only fired when:

1. A new job is added to the system
2. A job's schedule changes
3. A job executes and gets rescheduled

### DIFF
--- a/backend/cron/service.go
+++ b/backend/cron/service.go
@@ -97,7 +97,7 @@ func Start(ctx context.Context, config Config, eventSource *schemaeventsource.Ev
 
 	g.Go(func() error {
 		for {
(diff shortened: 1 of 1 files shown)


stuartwdouglas (approved):
This makes things significantly better, we will still get 3 events when raft is in use though, as every node fires the event. We also probably need to only fire them from the leader.

PR #4720, opened 2025-02-27
chore: simplify getting started steps

matt2e:
With interactive `ftl dev` we don’t need to tell people to open a second terminal window anymore.


wesbillman (approved):
Nice! We can also simplify this if we want to:

```sh
mkdir myproject
cd myproject
ftl init myproject .
ftl module new go alice
```

to be:
```
ftl init myproject
```

Or we could remove most of this and link to our docs getting started?

matt2e:
It's still a bit weird. When you make a project it tells you to `cd` into it and run`ftl dev` but our docs here say to first create the module and then run `ftl dev --wait-for alice`
Changing this would break the tests for this documentation though...

wesbillman:
Yeah that is weird. Maybe our docs should just say to do something like this:

```
ftl init myproject
```

Then follow the prompts to cd into `myproject` and run `ftl dev`
Then wait for `ftl dev` to show everything is ok then call `call alice.hello '{name: "Bob"}'` from the ftl dev terminal?

But I'm not sure how the docs testing will do with that... 🤔

PR #4765, opened 2025-03-03
chore: remove DeploymentCreatedEvent

stuartwdouglas:


wesbillman:
Nice! Should we nuke:
https://github.com/block/ftl/blob/main/backend/protos/xyz/block/ftl/timeline/v1/event.proto#L143-L148
and
https://github.com/block/ftl/blob/main/backend/protos/xyz/block/ftl/timeline/v1/event.proto#L172
as well?

I can do that in a follow on PR also.

PR #4824, opened 2025-03-11
fix: schema service can let in invalid schema

stuartwdouglas:


matt2e (approved):
Approved with the same question about best way to do this with raft...

PR #4828, opened 2025-03-11
feat: SQL verbs in JVM

### LINKED ISSUE #3969: JVM DB handle API
query verbs are not themselves associated with any DB, but can be used against a DB

figure out best way to express which DB a query verb is accessing as part of its injection pattern

worstell:
includes unified JVM support + clients in Java; Kotlin clients to follow 
fixes #3969


examples/java/mysql/src/main/ftl-generated-schema/mysql.pb
stuartwdouglas, line 1: I don't think this file is needed?

worstell: ah yep thanks for catching, was generating this file in a prior iteration

stuartwdouglas (approved):

PR #4829, opened 2025-03-11
chore: remove async event timeline event

matt2e:
These are no longer used


frontend/console/src/features/timeline/TimelineIcon.tsx
```diff
@@ -2,7 +2,6 @@ import {
   Call02Icon,
   CallIncoming04Icon,
   ComputerIcon,
-  CustomerServiceIcon,
```
matt2e, line 5: @wesbillman Does anything else need to be removed now that we don't used this icon anywhere else?

PR #4854, opened 2025-03-13
fix: use AllowOrigins as strings to support wildcards

wesbillman:


cmd/ftl/cmd_dev.go
```diff
@@ -99,7 +99,7 @@ func (d *devCmd) Run(
 	}
 
 	// Default to allowing all origins and headers for console requests in local dev mode.
-	d.ServeCmd.Ingress.AllowOrigins = []*url.URL{{Scheme: "*", Host: "*"}}
+	d.ServeCmd.Ingress.AllowOrigins = []string{"*"}
```
wesbillman, line 102: This was previously resulting in a `AllowOrigin` of `*://*` which isn't valid afaik.

matt2e (approved):

jvmakine (approved):

PR #4870, opened 2025-03-14
fix: add positions to more go build errors

matt2e:
This module code:
```go
//ftl:verb
func Verb(ctx context.Context, req struct{}) (TimeResponse, error) {
	return TimeResponse{Time: time.Now()}, nil
}
```
Would previously give this build error which lacks any information of where to find the bad code:
`error:time:build: Build failed: unsupported type <nil>`

We now try and attach the best position available to errors that occur in various places of `build/compile.go`, so now we get this error:
`error:time:build: Build failed: /Users/mtoohey/Code/ftl/examples/go/time/time.go:30:1: could not parse request type: unsupported type <nil>`

At this point in the code the verb's request type does not have any position information so it is given the position of the verb instead (ie column 1).

### DIFF
--- a/common/builderrors/builderrors.go
+++ b/common/builderrors/builderrors.go
@@ -10,7 +10,6 @@ import (
 
 type Position struct {
 	Filename    string
-	Offset      int
 	StartColumn int
 	EndColumn   int
 	Line        int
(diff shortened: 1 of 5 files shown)


common/builderrors/builderrors.go
```diff
@@ -10,7 +10,6 @@ import (
 
 type Position struct {
 	Filename    string
-	Offset      int
```
matt2e, line 13: Also removing this unused field

PR #4952, opened 2025-03-24
fix: add position to sql verbs and data types

matt2e:
- sqlc does not provide enough information for us to get a position of a verb in the source sql file.
- Instead we read in the sql files and try to match `-- name: <name>` lines with the names of the verbs.

### DIFF
--- a/internal/sql/sql.go
+++ b/internal/sql/sql.go
@@ -8,16 +8,21 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/alecthomas/types/optional"
 
 	"github.com/block/ftl/common/schema"
+	"github.com/block/ftl/common/slices"
+	"github.com/block/ftl/common/strcase"
 	"github.com/block/ftl/internal"
 	"github.com/block/ftl/internal/exec"
 	"github.com/block/ftl/internal/log"
 	"github.com/block/ftl/internal/moduleconfig"
 )
+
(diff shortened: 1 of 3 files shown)


internal/sql/sql.go
```diff
@@ -8,17 +8,22 @@ import (
...
 	"github.com/block/ftl/common/schema"
+	"github.com/block/ftl/common/slices"
+	"github.com/block/ftl/common/strcase"
 	"github.com/block/ftl/internal"
 	"github.com/block/ftl/internal/exec"
 	"github.com/block/ftl/internal/log"
 	"github.com/block/ftl/internal/moduleconfig"
 )
 
+var queryNameRegex = regexp.MustCompile(`^-- name: ([^ ]+)`)
```
matt2e, line 25: Confirmed that sqlc does not allow extra spaces before / after `name:`

internal/sql/sql.go
```diff
@@ -8,17 +8,22 @@ import (
...
 	"github.com/block/ftl/common/schema"
+	"github.com/block/ftl/common/slices"
+	"github.com/block/ftl/common/strcase"
 	"github.com/block/ftl/internal"
 	"github.com/block/ftl/internal/exec"
 	"github.com/block/ftl/internal/log"
 	"github.com/block/ftl/internal/moduleconfig"
 )
 
+var queryNameRegex = regexp.MustCompile(`^-- name: ([A-Za-z0-9_]+)`)
```
matt2e, line 25: Confirmed that sqlc does not allow extra spaces between each part

worstell (approved):

PR #4955, opened 2025-03-24
fix: messages to goose do not get parsed as CLI args

matt2e:
Users type messages to goose conversationally like this:
`goose there seem to be errors. I'd like to know why`
Or use quotation marks which should not be lost
`goose please make the verb log out "some example text"`

We now take everything after goose as a single argument we pass to goose.


matt2e:
LMK if we don't want this... I don't think this is 'correct' but when talking to goose it feels very odd when messages to goose fail because of this.

PR #5008, opened 2025-03-26
feat: add topic/subscription info commands

matt2e:
`ftl pubsub topic info <ref> ...`
```
homeloans.submittedApplications
  Partition 0
    Head:
      Offset:      933
      Request Key: req-cron-generator-generateApplication-4zaahctgk8k2meap
      Published:   2025-03-26T05:04:30Z
...
```
`ftl pubsub subscription info <ref> ...`
```
investigator.runCreditChecks
  Partition 0: Trailing head by 7 events     <- Indicates how far behind head. Or idle if it has caught up
    Last Event Consumed:
      Offset:      131
      Request Key: req-ingress-grpc-574bs0bj25lzmn52
      Published:   2025-03-26T05:06:13Z
    Consuming from:
      Offset:      132
      Request Key: req-ingress-grpc-574bs0bj25lzmn52
      Published:   2025-03-26T05:06:14Z
...
```

This is also exposed to FTL MCP


matt2e:
We should clean up the pubsub commands a bit like we did for mysql/postgres ones. We can probable get rid of `pubsub` and just have `ftl topic <subcommand>` and `ftl subscription <subcommand>`

PR #5010, opened 2025-03-26
fix: prevent injecting external topics

matt2e:
Previously we looked for `topic.Publish(...)` calls to detect if a module tried to publish to an external topic.
Now that we use injection for resources, we can instead check if a topic resource is from a different module.

When injecting an external topic I was getting a seemingly unrelated error that was not obvious that it was because it was not allowed (something about a partition mapper couldn't be found).


matt2e:
hmm tests are failing. Will look into the fix tomorrow

PR #5012, opened 2025-03-26
fix: dev provisioner was skipping partition count

matt2e:


stuartwdouglas:
Is this actually what we want? In local dev mode we only ever run a single instance of the runner, so won't think mean that if there is >1 partition then some will never be consumed?

matt2e:
@stuartwdouglas consumer groups spread around the partitions among the consumers. So in dev 1 runner will consume all partitions in parallel.

stuartwdouglas:
Oh yea, duh, sorry, I am fairly tired today...

stuartwdouglas (approved):

jvmakine (approved):

PR #5022, opened 2025-03-28
fix: trace graph shows long durations correctly

matt2e:
fixes https://github.com/block/ftl/issues/4836
Duration text is fixed and graph shows the correct width for each segment.


frontend/console/src/features/traces/TraceDetails.tsx
```diff
@@ -44,7 +45,7 @@ export const TraceDetails: React.FC<TraceDetailsProps> = ({ events, selectedEven
             .sort((a, b) => requestStartTime([a]) - requestStartTime([b]))
             .map((event, index) => {
               const traceEvent = event.entry.value as TraceEvent
-              const eventDurationMs = (traceEvent.duration?.nanos ?? 0) / 1000000
+              const eventDurationMs = traceEvent.duration ? durationToMillis(traceEvent.duration) : 0
```
wesbillman, line 48: Nice! Thank you!

PR #5038, opened 2025-04-01
feat: support transactions in JVM

### LINKED ISSUE #3967: Support transactions
transactions will only be supported within a single module

### LINKED ISSUE #3970: JVM transaction API

### LINKED ISSUE #4970: Support SQL Transactions

worstell:
fixes #3970
fixes #3967
fixes #4970

### DIFF
--- a/jvm-runtime/ftl-runtime/common/deployment/src/main/java/xyz/block/ftl/deployment/VerbProcessor.java
+++ b/jvm-runtime/ftl-runtime/common/deployment/src/main/java/xyz/block/ftl/deployment/VerbProcessor.java
@@ -336,6 +336,7 @@ public class VerbProcessor {
         Collection<AnnotationInstance> verbAnnotations = index.getIndex().getAnnotations(FTLDotNames.VERB);
         log.debugf("Processing %d verb annotations into decls", verbAnnotations.size());
         var beans = AdditionalBeanBuildItem.builder().setUnremovable();
+
         for (var verb : verbAnnotations) {
             boolean exported = verb.target().hasAnnotation(FTLDotNames.EXPORT);
             var method = verb.target().asMethod();
@@ -346,7 +347,23 @@ public class VerbProcessor {
             String className = method.declaringClass().name().toString();
             beans.addBeanClass(className);
             schemaContributorBuildItemBuildProducer.produce(new SchemaContributorBuildItem(moduleBuilder -> moduleBuilder
-                    .registerVerbMethod(method, className, exported, ModuleBuilder.BodyType.ALLOWED)));
+                    .registerVerbMethod(method, className, exported, false, ModuleBuilder.BodyType.ALLOWED)));
+        }
+
+        Collection<AnnotationInstance> transactionAnnotations = index.getIndex().getAnnotations(FTLDotNames.TRANSACTIONAL);
+        for (var txn : transactionAnnotations) {
+            boolean exported = txn.target().hasAnnotation(FTLDotNames.EXPORT);
+            var method = txn.target().asMethod();
+            if (method.hasAnnotation(FTLDotNames.VERB) || method.hasAnnotation(FTLDotNames.CRON)
+                    || method.hasAnnotation(FTLDotNames.SUBSCRIPTION)
+                    || method.hasAnnotation(FTLDotNames.FIXTURE)) {
+                throw new RuntimeException(
+                        "Method " + method + " cannot have both @Transaction and @Verb, @Cron, @Fixture or @Subscription");
+            }
+            String className = method.declaringClass().name().toString();
+            beans.addBeanClass(className);
+            schemaContributorBuildItemBuildProducer.produce(new SchemaContributorBuildItem(moduleBuilder -> moduleBuilder
+                    .registerVerbMethod(method, className, exported, true, ModuleBuilder.BodyType.ALLOWED)));
         }
 
         Collection<AnnotationInstance> cronAnnotations = index.getIndex().getAnnotations(FTLDotNames.CRON);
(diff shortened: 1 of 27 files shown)


jvm-runtime/ftl-runtime/common/deployment/src/main/java/xyz/block/ftl/deployment/VerbProcessor.java
```diff
@@ -346,7 +350,26 @@ public void verbsAndCron(CombinedIndexBuildItem index,
...
             schemaContributorBuildItemBuildProducer.produce(new SchemaContributorBuildItem(moduleBuilder -> moduleBuilder
-                    .registerVerbMethod(method, className, exported, ModuleBuilder.BodyType.ALLOWED)));
+                    .registerVerbMethod(method, className, exported, false, ModuleBuilder.BodyType.ALLOWED)));
+        }
+
+        Collection<AnnotationInstance> transactionAnnotations = index.getIndex().getAnnotations(FTLDotNames.TRANSACTIONAL);
+        for (var txn : transactionAnnotations) {
+            boolean exported = txn.target().hasAnnotation(FTLDotNames.EXPORT);
+            var method = txn.target().asMethod();
+            if (method.hasAnnotation(FTLDotNames.EXPORT)) {
```
stuartwdouglas, line 360: Why does this restriction exist? IMHO it does not make sense, it is very common to have e.g. a HTTP ingress method that does all it's logic inside a single transaction.  

For SQL verbs you end up leaking DB details to callers, which is not great, but there are not concerns like that here.

jvm-runtime/ftl-runtime/common/runtime/pom.xml
```diff
@@ -28,6 +28,10 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-grpc</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-smallrye-context-propagation</artifactId>
```
stuartwdouglas, line 33: Why is this included? This is for propagating context between multiple threads in a single request, which I don't know if we are really concerned with.

jvm-runtime/ftl-runtime/common/runtime/src/main/java/xyz/block/ftl/runtime/FTLRunnerConnectionImpl.java
```diff
@@ -249,8 +267,106 @@ public void close() {
...
+            throw new RuntimeException(e);
+        }
+    }
+
+    @Override
+    public void commitTransaction(String transactionId) {
+        CompletableFuture<String> cf = new CompletableFuture<>();
+        queryService.commitTransaction(CommitTransactionRequest.newBuilder()
+                .setTransactionId(transactionId)
+                .setDatabaseName("testdb")
```
stuartwdouglas, line 302: Same here and rollback.

jvm-runtime/ftl-runtime/common/runtime/src/main/java/xyz/block/ftl/runtime/FTLRunnerConnectionImpl.java
```diff
@@ -249,8 +267,106 @@ public void close() {
     }
 
     @Override
-    public String executeQueryOne(String dbName, String sql, String paramsJson, String[] colToFieldName) {
-        var request = createQueryRequest(dbName, sql, paramsJson, colToFieldName)
+    public String beginTransaction() {
+        CompletableFuture<String> cf = new CompletableFuture<>();
+        queryService.beginTransaction(BeginTransactionRequest.newBuilder().setDatabaseName("testdb").build(),
```
stuartwdouglas, line 272: "testdb" is hard coded.

jvm-runtime/ftl-runtime/common/runtime/src/main/java/xyz/block/ftl/runtime/CallRequestContext.java
```diff
@@ -0,0 +1,56 @@
...
+import jakarta.enterprise.context.RequestScoped;
+import jakarta.inject.Inject;
+
+import io.quarkus.arc.Arc;
+import io.smallrye.context.SmallRyeThreadContext;
+import io.smallrye.context.api.ThreadContextConfig;
+import xyz.block.ftl.v1.Metadata;
+
+@RequestScoped
+public class CallRequestContext {
```
stuartwdouglas, line 12: @RequestScoped is already an implicit call request context. I would just create a `CurrentTransaction` request scoped class to track the transaction id.

jvm-runtime/ftl-runtime/common/runtime/src/main/java/xyz/block/ftl/runtime/CallRequestContext.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @Inject
+    @ThreadContextConfig
+    SmallRyeThreadContext threadContext;
```
stuartwdouglas, line 16: It doesn't look like this is used.

jvm-runtime/ftl-runtime/common/runtime/src/main/java/xyz/block/ftl/runtime/FTLRunnerConnectionImpl.java
worstell: oof, thanks. forgot i'd done that to test along the way

jvm-runtime/ftl-runtime/common/deployment/src/main/java/xyz/block/ftl/deployment/VerbProcessor.java
worstell: i'd lumped these in conceptually as internal DB operations, but your point makes sense-  removed. the remaining restrictions are:
1. txn verb cannot make external calls
2. txn verbs cannot inject other txn verbs (nested transactions)
3. txn verb must access a datasource
4. txn verb can only access a single datasource

stuartwdouglas (approved):

PR #5064, opened 2025-04-02
chore: sql verb polish

worstell:
- if only one column in param or response, unwrap and return type directly (e.g. string rather than QueryResult { string })
- if whole table schema is used, create and use data type that is the singularized name of the table (e.g. `User` from `users`)


backend/controller/sql/database_integration_test.go
```diff
@@ -43,40 +39,6 @@ func TestMySQL(t *testing.T) {
 		in.Call[in.Obj, in.Obj]("mysql", "query", map[string]any{}, func(t testing.TB, response in.Obj) {
 			assert.Equal(t, "hello", response["data"])
 		}),
-		in.VerifySchemaVerb("mysql", "createRequest", func(ctx context.Context, t testing.TB, sch *schemapb.Schema, verb *schemapb.Verb) {
```
worstell, line 46: prefer testing this in the schema tests rather than this integration test

PR #5065, opened 2025-04-02
fix: show trace url as a link in terminal

matt2e:
Uses escape codes to encode trace URL as a hyperlink (based on: https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda)
A better fix would be to detect if the terminal emulator supports this escape sequence if out the hyperlink on the request key line itself.

In iTerm2:
Screenshot 2025-04-03 at 9 37 12 am

In Mac Terminal (fallbacks to just text):
Screenshot 2025-04-03 at 9 37 29 am


matt2e:
This package would make it easy if we want to do it better: https://github.com/savioxavier/termlink

wesbillman (approved):
Nice!!

PR #5068, opened 2025-04-03
fix: JVM docs

stuartwdouglas:


docs/docs/reference/verbs.md
```diff
@@ -151,8 +151,8 @@ Verb clients are generated by FTL. If the callee verb belongs to the same module
 own client:
 
 ```kotlin
-@VerbClient(name="time")
 interface TimeClient {
+    @VerbClient(name="time")
```
matt2e, line 155: This also fails. name is not allowed I think

PR #5124, opened 2025-04-12
chore: improve formatting in goose panel

worstell:
- filter out some more unwanted lines
- support markdown formatting
- improve formatting for code blocks

https://github.com/user-attachments/assets/f532f748-3f5b-4faa-8719-8dd95cd23091

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,0 +1,5 @@
+{
+	"dependencies": {
+		"react-markdown": "^10.1.0"
+	}
+}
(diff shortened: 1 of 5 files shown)


pnpm-lock.yaml
```diff
@@ -6,6 +6,12 @@ settings:
```
worstell, line 6: @stuartwdouglas how did you fix this last time? 😬

stuartwdouglas: `sed -i '' "s/, tarball:.*}/}/" pnpm-lock.yaml` :-)

I don't know what the proper solution is.

package.json
```diff
@@ -0,0 +1,5 @@
+{
```
stuartwdouglas, line 1: Is this supposed to be a new file?

PR #5169, opened 2025-04-17
fix: reset pubsub retries to allow for runner rebalances

matt2e:
fixes https://github.com/block/ftl/issues/5165

Reset subscription would fail if any partitions were not currently claimed by a runner.
This happens whenever a rebalance is needed (as well as after a reset command is sent).
The original issue was filed because the two reset commands are fired in quick succession when tests are run with release binaries, exposing the race condition.

This change does the following:
- Reset pubsub command retries 10 times (500ms delay between each)
- Each attempt only resets the remaining partitions rather than all partitions


alecthomas (approved):
Nice find!

PR #5194, opened 2025-04-22
fix: module name errors in toml file

### LINKED ISSUE #5039: `ftl dev` hangs if module name in go.mod does not match the name in ftl.toml
If this happens the module just hangs with no output, debug logging will give some clues about a failure to read the go.mod however it is not obvious what has cased the issue.

wesbillman:
Fixes #5039

Special handling for if initial builds do not succeed so we can force retry them when modules/config change. Long term we might have to refactor how the engine works to remove lots of these special cases for the engine

Screenshot 2025-04-21 at 11 49 07 AM


wesbillman:
Found a better approach to this which requires less scary changes :)

PR #5202, opened 2025-04-23
feat: new engine dev flow

wesbillman:


internal/buildengine/engine.go
```diff
@@ -377,7 +377,502 @@ func (e *Engine) Modules() []string {
...
+		return errors.WithStack(err)
+	}
+
+	// Store schema after successful build
+	if sch, ok := e.GetModuleSchema(moduleName); ok {
+		builtModules.Store(moduleName, sch)
+	}
+
+	// Deploy the module
+	if err := e.deployModule(ctx, moduleName); err != nil {
```
stuartwdouglas, line 484: It's more complex but technically we can start building other modules as soon as their deps are built, which would potentially provide a small speedup.

wesbillman: yeah true! I'll work on that tomorrow to see if I can eek out some more speed improvements.

I did notice that with this, we  already see more consistent build times and they are a bit faster than the current setup.

wesbillman:
We might want to wait on some of these changes since they're fairly risky and we might have higher priority issues to work through.

wesbillman:
Closing PR to avoid noise but I'll keep the branch around for a bit.

PR #5215, opened 2025-04-24
fix: show error if module-dir does not exist

### LINKED ISSUE #5182: `ftl dev` does not exit on error due to missing module dir
Steps:
- Rename `examples/go` dir to something else
- Run `ftl dev`

Expected:
- FTL should exit with an error saying that `examples/go` could not be found

What actually happens:
- `ftl dev` seems to hang without exiting or printing the error
- It works correctly if `terminal.LaunchEmbeddedConsole` is skipped in `cmd_dev.go`

wesbillman:
Fixes #5182

I'm not sure if this will have unwanted side effects as it could have weird behaviors with new ftl projects where the folder is empty since it's a newly created project.

### DIFF
--- a/internal/buildengine/engine.go
+++ b/internal/buildengine/engine.go
@@ -389,17 +389,11 @@ func (e *Engine) watchForModuleChanges(ctx context.Context, period time.Duration
 	logger := log.FromContext(ctx)
 
 	watchEvents := make(chan watch.WatchEvent, 128)
-	ctx, cancel := context.WithCancelCause(ctx)
 	topic, err := e.watcher.Watch(ctx, period, e.moduleDirs)
 	if err != nil {
-		cancel(errors.Wrap(errors.Join(err, context.Canceled), "watch failed"))
-		return errors.WithStack(err)
+		return errors.Wrap(err, "failed to start watcher")
 	}
 	topic.Subscribe(watchEvents)
-	defer func() {
-		// Cancel will close the topic and channel
-		cancel(errors.Wrap(context.Canceled, "watch stopped"))
-	}()
 
 	// Build and deploy all modules first.
 	if err := e.BuildAndDeploy(ctx, optional.None[int32](), true, false); err != nil {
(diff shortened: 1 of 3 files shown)


worstell (approved):

cmd/ftl/cmd_dev.go
```diff
@@ -79,6 +79,22 @@ func (d *devCmd) Run(
 	}
 	os.Setenv("FTL_DEV_DIRS", strings.Join(absDirs, ","))
 
+	// Validate module directories before launching console or starting build engine.
```
matt2e, line 82: I don't think this is needed. These cases are already be handled in `main`, though it was broken as of last week (fixed here: https://github.com/block/ftl/pull/5180):
```
ftl: error: [<dirs> ...]: stat /Users/mtoohey/Code/ftl/examples/javafake: no such file or directory
🐚 mtoohey ~/Code/ftl [main] $ ftl dev fakedir          
Usage: ftl dev [<dirs> ...] [flags]
Run "ftl dev --help" for more information.
```
```
ftl: error: [<dirs> ...]: stat /Users/mtoohey/Code/ftl/fakedir: no such file or directory
🐚 mtoohey ~/Code/ftl [main] $ ftl dev Dockerfile
Usage: ftl dev [<dirs> ...] [flags]
Run "ftl dev --help" for more information.

ftl: error: [<dirs> ...]: "/Users/mtoohey/Code/ftl/Dockerfile" exists but is not a directory
```

matt2e (approved):
Looks good Wes :+1: I was surprised that you also didn't need to change the language plugins (which also use the watcher), but it looks like they never closed the channel before... So that's nice to clean this up!

cmd/ftl/cmd_dev.go
matt2e: Back when I create the issue about this was when I fixed it made that PR, but the issue remained because the interactive console code was eating the error for some reason. But that appears to be fixed now... 🤷‍♂️

PR #5231, opened 2025-04-28
fix: sync schema before deploying

### LINKED ISSUE #5229: `ftl deploy` does not read schema from server
If you run `ftl deploy` echo against the sandbox it fails because the build engine can't find time.

wesbillman:
@stuartwdouglas does this fix #5229? Or is there more to it?


stuartwdouglas:
Seems to fix it

PR #5314, opened 2025-05-02
fix: validate schema for refs with invalid visibility

matt2e:
fixes https://github.com/block/ftl/issues/5313

### DIFF
--- a/backend/ingress/handler_test.go
+++ b/backend/ingress/handler_test.go
@@ -27,15 +27,15 @@ func TestIngress(t *testing.T) {
 	sch, err := schema.ParseString("", `
 	realm test {
 		module test {
-			data AliasRequest {
+			export data AliasRequest {
 				aliased String +alias json "alias"
 			}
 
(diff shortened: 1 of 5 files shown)


jvmakine (approved):

stuartwdouglas:
I think this failure is due to an existing problem in the JVM runtime with it not exporting decls used in other decls.

stuartwdouglas:
Hopefully https://github.com/block/ftl/pull/5246 will fix the issue

matt2e:
@stuartwdouglas yep great to fix the JVM issue :+1:
I think schema validation should check it regardless, rather than rely on each language always producing valid schema. But also yeh theres the `export realm` case which can't be fixed by transitive exports as they can cross module boundaries.

matt2e:
Blocked by #5315

stuartwdouglas:
That other PR has issues unrelated to the visbillity, so I have broken out the changes into https://github.com/block/ftl/pull/5316

PR #5333, opened 2025-05-04
feat: Fixed realm name for kube deployments

stuartwdouglas:


common/schema/schema.go
```diff
@@ -350,6 +350,7 @@ type SchemaState struct {
 	ChangesetEvents  []*DeploymentRuntimeEvent `protobuf:"3"`
 	DeploymentEvents []*DeploymentRuntimeEvent `protobuf:"4"`
 	Realms           []*RealmState             `protobuf:"5"`
+	InternalRealm    string                    `protobuf:"6"`
```
jvmakine, line 353: I would prefer inferring this from existing `Realms`. I feel that would make this more flexible, if we ever want to support multiple internal realms in the future.

jvmakine (approved):

PR #5351, opened 2025-05-05
fix: make schema position filenames relative to project root

worstell:


go-runtime/schema/common/common.go
```diff
@@ -37,20 +39,26 @@ var (
 	FtlSinglePartitionMapPath = "github.com/block/ftl/go-runtime/ftl.SinglePartitionMap[E any]"
 
 	extractorRegistery = xsync.NewMapOf[reflect.Type, ExtractDeclFunc[schema.Decl, ast.Node]]()
+
+	ProjectRootFlagName = "project_root"
```
wesbillman, line 43: This we still need this to be public? or did the other change fix the issue?

wesbillman (approved):
Awesome! Thank you!

go-runtime/schema/common/common.go
worstell: this gets used in the schema package in `extract.go`

PR #5374, opened 2025-05-06
fix: duplicated, unformatted engine errors

wesbillman:
It's still very verbose with `jvm` errors, but the duplicate errors should be gone now. Module build errors are handled via our `rawEngineUpdates` channel, so we can ignore them in our error logging of `BuildAndDeploy`


wesbillman:
I think we should wait on merging this as future FTL buildengine changes will likely solve this in a more robust way. Also, this can lead to new issues given the errors aren't actually returned in other flows that rely on them.

PR #5379, opened 2025-05-07
feat: change build engine to handle the watches

stuartwdouglas:


go-runtime/goplugin/service.go
```diff
@@ -173,181 +161,47 @@ func (s *Service) Build(ctx context.Context, req *connect.Request[langpb.BuildRe
...
 // buildAndSend builds the module and sends the build event to the stream.
 //
 // Build errors are sent over the stream as a BuildFailure event.
 // This function only returns an error if events could not be send over the stream.
-func buildAndSend(ctx context.Context, stream *connect.ServerStream[langpb.BuildResponse], projectConfig projectconfig.Config, stubsRoot string, buildCtx buildContext,
-	isAutomaticRebuild bool, transaction watch.ModifyFilesTransaction, ongoingState *compile.OngoingState, devMode bool) error {
-	buildEvent, err := build(ctx, projectConfig, stubsRoot, buildCtx, isAutomaticRebuild, transaction, ongoingState, devMode)
+func buildAndSend(ctx context.Context, projectConfig projectconfig.Config, stubsRoot string, buildCtx buildContext,
+	ongoingState *compile.OngoingState, devMode bool) (*connect.Response[langpb.BuildResponse], error) {
+	captureTx := &captureTransaction{}
```
matt2e, line 180: I think the transaction stuff has to be communicated back to the engine somehow, maybe as part of the build result

go-runtime/goplugin/service.go
```diff
@@ -375,38 +229,41 @@ func build(ctx context.Context, projectConfig projectconfig.Config, stubsRoot st
 	return &langpb.BuildResponse{
 		Event: &langpb.BuildResponse_BuildSuccess{
 			BuildSuccess: &langpb.BuildSuccess{
-				ContextId:          buildCtx.ID,
-				IsAutomaticRebuild: isAutomaticRebuild,
-				Errors:             langpb.ErrorsToProto(buildErrs),
-				Module:             moduleProto,
-				Deploy:             deploy,
+				ModifiedFiles: captureTx.files,
```
matt2e, line 232: I dont think a list of files is enough here. We should capture the hashes at the time of modification by the build. And then feed those hashes back into the engine's watcher. Otherwise we miss out on file changes in this case:
- Build starts
- Build modifies a file
- User modifies the same file
- Build completes
A new build should start at this point

internal/buildengine/engine.go
```diff
@@ -1352,6 +1353,7 @@ func (e *Engine) newModuleMeta(ctx context.Context, config moduleconfig.Unvalida
 		plugin:         plugin,
 		events:         events,
 		configDefaults: customDefaults,
+		watch:          watch.NewWatcher(optional.Zero(validConfig.BuildLock), validConfig.Watch...),
```
matt2e, line 1356: Is there a race condition using the build lock here?
Using the build lock here could also cause race conditions. You want the watcher to receive the modified file hashes before it starts polling for changes again. But the build lock is unlocked when the plugin sends the build finished message... but that doesnt mean the engine has handled the build result yet.

matt2e (approved):
Looks great! Love all that code deleted

Added some questions around file modification detection

go-runtime/goplugin/service.go
stuartwdouglas: I don't think it is worth worrying about this, as if a user is modifying files modified by the build while the build is running there are already races we can't guard against. In partular:
- If the user modifies the file slightly earlier the build just wipes out their changes
- Currently there is still a race between us modifying the contents and reading the hash, so there is still a window for a race to occur

In general modifying the files that the build modified while the build is in progress is going to give you a bad time.

wesbillman (approved):
Nice! Thanks @stuartwdouglas !

backend/protos/xyz/block/ftl/language/v1/service.proto
```diff
@@ -240,26 +225,10 @@ service LanguageService {
...
-  //
-  // The request can include the option to "rebuild_automatically". In this case the plugin should watch for
-  // file changes and automatically rebuild as needed as long as this build request is alive. Each automactic
-  // rebuild must include the latest build context id provided by the request or subsequent BuildContextUpdated
-  // calls.
-  rpc Build(BuildRequest) returns (stream BuildResponse);
-
-  // While a Build call with "rebuild_automatically" set is active, BuildContextUpdated is called whenever the
-  // build context is updated.
+  // Build the module and recive the result
```
matt2e, line 228:
```suggestion
  // Build the module and receive the result
```

PR #5421, opened 2025-05-09
fix: support slice params in sql verbs

worstell:


wesbillman (approved):
Nice! Thanks for fixing!

I'm not expert here, but this looks good to me :)

PR #5423, opened 2025-05-10
fix: support slice params in sql verbs

worstell:


jvm-runtime/ftl-runtime/java/deployment/src/main/java/xyz/block/ftl/javalang/deployment/JavaCodeGenerator.java
```diff
@@ -404,7 +404,7 @@ private TypeName toJavaTypeName(Type type, Map<DeclRef, Type> typeAliasMap, Map<
             boolean boxPrimitives) {
         if (type.hasArray()) {
             return ParameterizedTypeName.get(ClassName.get(List.class),
-                    toJavaTypeName(type.getArray().getElement(), typeAliasMap, nativeTypeAliasMap, false));
+                    toJavaTypeName(type.getArray().getElement(), typeAliasMap, nativeTypeAliasMap, true));
```
worstell, line 407: cc @stuartwdouglas

PR #5536, opened 2025-05-20
fix: suffix sql full table types with Row

worstell:


wesbillman (approved):
Nice! This will help a bunch with conflicts we've seen!

PR #5677, opened 2025-06-04
fix: handle quarkus launch failure

matt2e:
- Previously if quarkus failed to launch the build would be stuck for 100 mins waiting to connect
- Now it returns with the actual build errors

### DIFF
--- a/jvm-runtime/plugin/common/service.go
+++ b/jvm-runtime/plugin/common/service.go
@@ -103,7 +103,6 @@ func (s *Service) GenerateStubs(ctx context.Context, req *connect.Request[langpb
 }
 
 func (s *Service) SyncStubReferences(ctx context.Context, req *connect.Request[langpb.SyncStubReferencesRequest]) (*connect.Response[langpb.SyncStubReferencesResponse], error) {
-
 	if req.Msg.Schema == nil {
 		return connect.NewResponse(&langpb.SyncStubReferencesResponse{}), nil
 	}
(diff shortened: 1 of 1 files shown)


jvm-runtime/plugin/common/service.go
```diff
@@ -315,33 +318,33 @@ func (s *Service) handleState(ctx context.Context, state *hotreloadpb.SchemaStat
...
-		err := command.Run()
-		if err != nil {
-			stdout.FinalizeCapture(true)
-			logger.Errorf(err, "Dev mode process exited with error")
-			cancel(errors.Wrap(errors.Join(err, context.Canceled), "dev mode process exited with error"))
-		} else {
-			logger.Infof("Dev mode process exited")
-			cancel(errors.Wrap(context.Canceled, "dev mode process exited"))
-		}
+func (s *Service) launchQuarkusProcess(ctx context.Context, devModeBuild string, projectConfig projectconfig.Config, buildCtx buildContext, stdout *errorDetector, errChan chan error) {
```
matt2e, line 321: Passing in the errChan here so that failure to launch is passed back

jvm-runtime/plugin/common/service.go
```diff
@@ -233,7 +231,12 @@ func (s *Service) runQuarkusDev(ctx context.Context, projectConfig projectconfig
 				Errors: ers,
 			}}}), nil
 	case err := <-errorChan:
```
matt2e, line 233: errChan now gets either the failure to launch error, or the error from the above `go func() {}` (eg connect timeout), whichever fails first

jvm-runtime/plugin/common/service.go
```diff
@@ -119,24 +118,23 @@ func (s *Service) SyncStubReferences(ctx context.Context, req *connect.Request[l
...
 	_ = os.Setenv("QUARKUS_ANALYTICS_DISABLED", "true") //nolint:errcheck
 	logger = logger.Module(req.Msg.BuildContext.ModuleConfig.Name)
 	ctx = log.ContextWithLogger(ctx, logger)
+
 	buildCtx, err := buildContextFromProto(req.Msg.BuildContext)
 	if err != nil {
 		return nil, errors.WithStack(err)
 	}
+
+	release, err := flock.Acquire(ctx, buildCtx.Config.BuildLock, BuildLockTimeout)
```
matt2e, line 132: build lock is now locked on each build, not just the first build in `runQuarkusDev`

jvm-runtime/plugin/common/service.go
```diff
@@ -315,33 +318,33 @@ func (s *Service) handleState(ctx context.Context, state *hotreloadpb.SchemaStat
 	}), nil
 }
 
-func (s *Service) launchQuarkusProcessAsync(ctx context.Context, devModeBuild string, projectConfig projectconfig.Config, buildCtx buildContext, stdout *errorDetector) {
-	go func() {
```
matt2e, line 319: moved the `go` keyword to the actual function call which I think is more idiomatic

jvm-runtime/plugin/common/service.go
stuartwdouglas: Do we even need the build lock now?

stuartwdouglas (approved):

